PR #14, opened 2019-09-10
Table.to_csv Zip Compression

jburchard:
- Add compression type `.zip` to `Table.to_csv()` method.
- Add stand alone `Table.to_csv_zip()` method that allows for the specification of the file name and the ability to append to an existing archive.

### DIFF
--- a/parsons/utilities/zip_archive.py
+++ b/parsons/utilities/zip_archive.py
@@ -4,10 +4,12 @@ import zipfile
 def create_archive(archive_path, file_path, file_name=None, if_exists='replace'):
     """
     `Args:`
-        archive_name: str
+        archive_path: str
             The file name of zip archive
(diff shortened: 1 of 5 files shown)


parsons/utilities/zip_archive.py
```diff
@@ -0,0 +1,46 @@
...
+    `Returns:`
+        Zip archive path
+    """
+
+    if if_exists == 'append':
+        write_type = 'a'
+    else:
+        write_type = 'w'
+
+    if not file_name:
```
ydamit: Of course I defer to your greater experience, but I wanted to raise that I've had difficulty using `if not` to rule out Nonetypes. I've often had to switch to `if variable is not None:`

jburchard:
@ydamit - One more pass please?

PR #29, opened 2019-09-20
Hustle Connector v1

jburchard:
Initial pass at a Hustle connector.

Includes the following methods:

`get_organization()`
`get_organizations()`
`get_group()`
`get_groups()`
`get_lead()`
`get_leads()`
`create_lead()`
`create_leads()`
`update_lead()`

Also created a utility to allow for smart column mapping, so different column names can be passed in and it smartly matches them. That is, you can pass in `fn`, `first` or `first_name` and it will all treat the same.


parsons/utilities/json_format.py
```diff
@@ -11,3 +11,32 @@ def arg_format(arg):
...
+    """
+    Remove empty keys from a dictionary. This method is useful when passing jsons
+    in which a null field will update the value to null and you don't want that.
+    """
+
+    clean_dict = {}
+
+    for k, v in dirty_dict.items():
+        if v:
+            if str and len == 0:
```
eliotst: This line seems like a typo? It will always evaluate to `False` since `len == 0` is always `False` (`len` evaluates as a built-in function).

If you meant `isinstance(v, str) and len(v) == 0`, this will still always evaluate to `False`, since you filter out all empty strings on line 36 (since in Python, empty strings are falsy).

So, I think lines 37 + 38 are unnecessary.

parsons/utilities/json_format.py
```diff
@@ -11,3 +11,32 @@ def arg_format(arg):
     arg_list[0] = arg_list[0].lower()
 
     return ''.join(arg_list)
+
+
+def list_to_string(list_arg):
```
eliotst: This function doesn't seem to be used anywhere in the code.

jburchard:
@eliotst - Thanks for the feedback. Ready for final review / approval.

PR #33, opened 2019-09-21
Revert "VAN: Add ID to find_person() and upsert_person()"

jburchard:
I'm having some failing errors with this with the validation and having some issues sorting them out.
Additionally, and this is on me, I think that we are trying to do _too_ many things in the _find_person()_ method.

My suggestion:
- We just direct users to the `get_person()` method if they have an id. That is more intuitive and in line with typical APIs.
- We move the `match_map` to a separate method called, `find_person_dict()` or something like that. I don't think that it needs to have the search validation, as it is an advanced method.

However, in the midst of this, you were fixing a bug too, right?


jburchard:
I was running into some issues with `_search_validate()`. It was erroring on combinations that should have been okay. I think that the issue was with the flattening of the json.

PR #46, opened 2019-10-01
Table: Fix zip compression bugs.

jburchard:
The `Table.to_s3_csv()` method was not actually compressing the files correctly.

- Fixed compression in `Table.to_s3_csv()`
- Added parameter to `Table.to_csv()` that allows you to specify the underlying file in the archive. I'm hesistant to add more parameters to the `.to_csv()` but I didn't see a better path absent to majorly hacky code. I think that it should be fine.

### DIFF
--- a/parsons/utilities/files.py
+++ b/parsons/utilities/files.py
@@ -162,6 +162,12 @@ def zip_check(file_path, compression_type):
 def extract_file_name(file_path=None, include_suffix=True):
     """
     Extract the file name with the file path string.
+
+    file_path: str
+        The file path
+    include_suffix: boolean
+        If True, includes full file name with suffix. If False returns the
+        file name without the suffix (e.g. "myfile.zip" vs. "myfile").
     """
 
     if not file_path:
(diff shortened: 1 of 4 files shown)


parsons/utilities/files.py
```diff
@@ -152,12 +159,21 @@ def zip_check(file_path, compression_type):
...
+        file name without the suffix (e.g. "myfile.zip" vs. "myfile").
     """
 
     if not file_path:
         return None
 
+    if include_suffix:
+        return file_path.split('/')[-1]
+
     return file_path.split('/')[-1].split('.')[0]
```
eliotst: I know this isn't technically in the scope of the PR, but this line won't properly handle filenames with multiple '.'s (eg `one.two.csv`). instead of just grabbing the first element in the list when splitting on '.', this should join all elements of the list except the last one.

jburchard: Good catch. How would you suggest that I address?

PR #47, opened 2019-10-02
Table: File Size Attribute

jburchard:
Starting to do a little bit of work to prepare for VAN Bulk Import endpoints. They have some limitations around file size, so I thought that this might be useful for that and other things.

`Table.csv_size` - An attribute to tables that essentially ball parks the eventual CSV size by pulling the first 1K rows of the file and creating a CSV. It then extrapolates the entire file size. This should be definitely considered an estimation.


eliotst:
It looks like the checks are failing doing to a lint error.

jburchard:
Fixed lint issue.

In the cold light of day, not sure that this is great, as the calculation might be a little bit off. Interested in thoughts.

eliotst:
I think what you are doing makes sense here. A sample size of 1,000 seems like it should net a pretty reasonable estimation (though I'm no data scientist). One other way to address the use case could be to have a function that breaks up the table into multiple CSV's, either by number of rows (more straightforward to implement) or number of bytes (a lot harder to implement, but more tuned to your use case)..?

PR #56, opened 2019-10-14
VAN: Changed Entity Methods

jburchard:
A couple of the VAN changed entity GET methods.

### DIFF
--- a/parsons/ngpvan/activist_codes.py
+++ b/parsons/ngpvan/activist_codes.py
@@ -1,16 +1,13 @@
 """NGPVAN Activist Code Endpoints"""
 
 from parsons.etl.table import Table
+from parsons.ngpvan.utilities import action_parse
 import logging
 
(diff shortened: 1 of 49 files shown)


parsons/ngpvan/activist_codes.py
```diff
@@ -44,3 +41,60 @@ def get_activist_code(self, activist_code_id):
         r = self.connection.get_request(f'activistCodes/{activist_code_id}')
         logger.info(f'Found activist code {activist_code_id}.')
         return r
+
+    def toggle_activist_code(self, id, activist_code_id, action, id_type='vanid'):
```
elyse-weiss: @jburchard just to be clear - will toggle still work? I know Yotam has it in code.

jburchard: Yes, it does, though is no longer surfaced in the docs. It is a deprecated method.

PR #58, opened 2019-10-16
temporary file management

eliotst:
This commit updates how Parsons handles temporary files. The
changes are meant to minimize the need for users to manage temp
files themselves, specifically when using a Parsons `Table`.

This commit introduces a new class
`parsons.utilities.resources.ResourceManager` that tracks
resources that are used by one or more `Table` so that the
resources (i.e. temp files) are cleaned up when they are no
longer being used by the table(s).

---

This PR does not deal with issues around temporary files returned by `S3.get_file` or `STP.get_file`, etc. I think figuring out how to manage those temporary files might involve changes to the API's, which I was hoping to avoid.


eliotst:
My first Parsons PR!

eliotst:
So, I'm not totally sold on this solution. I'd love to talk through it. I think at the end of the day it's fairly harmless, since it just fixes up some of the behind-the-scenes resource management. 

.. but it doesn't address what seems to be the most common cause of the too many open files issues: `S3.get_file`

PR #67, opened 2019-10-24
fix up exception message from APIConnector

eliotst:
This commit fixes the message passed to the `HTTPError` exception
being raised when the 
`parsons.utilities.api_connector.APIConnector` validates an HTTP
response. The new exception message includes the status code and
reason (if provided), along with the body of the response if it
is JSON data.


jburchard:
LGTM!

PR #69, opened 2019-10-28
Sample Code

elyse-weiss:


sample_code/apply_activist_code.py
```diff
@@ -0,0 +1,57 @@
...
+logger.addHandler(_handler)
+logger.setLevel('INFO')
+
+# Create dictionary of VAN states and API keys from multiline Civis credential
+
+myv_states = {x.split(",")[0]: x.split(",")[1] for x in os.environ['VAN_PASSWORD'].split("\r\n")}
+myv_keys = {k: VAN(api_key=v, db='MyVoters') for k,v in myv_states.items()}
+
+# Create simple set of states for insertion into SQL
+states = "','".join([s for s in myv_keys])
```
ydamit: Not sure we need to keep this in light of the simplified SQL

sample_code/apply_activist_code.py
```diff
@@ -0,0 +1,57 @@
...
+
+# SQL to pull those needing Activist Code
+sql = f"""
```
ydamit: Doesn't seem like this needs to be an f-string

sample_code/s3_to_redshift.py
```diff
@@ -0,0 +1,48 @@
...
+schema = os.environ['SCHEMA']
+
+keys = s3.list_keys(bucket)
+files = keys.keys()
+
+if len(keys) == 0:
+    logger.info("No files to sync today!")
+else:
+    logger.info(f"Pulling {str(len(files))} files down from s3...")
+    for x in files:
```
ydamit: Should we also include the file closer utility, in case it's a lot of files? Or is that too complicated for this sample code?

sample_code/s3_to_s3.py
```diff
@@ -0,0 +1,53 @@
...
+_handler = logging.StreamHandler()
+_formatter = logging.Formatter('%(levelname)s %(message)s')
+_handler.setFormatter(_formatter)
+logger.addHandler(_handler)
+logger.setLevel('INFO')
+
+# Let's write some code!
+
+# Get Vendor Bucket Information - this assumes it lives in Redshift
+bucket_guide = rs.query(f"select * from {table}")
```
ydamit: This seems like a pretty big "if", plus we don't have `table` defined anywhere before this. I'm wondering if it might not be better to use some more `S3` methods, like `list_buckets()` for this rather than assuming it'll be in Redshift?

sample_code/s3_to_s3.py
```diff
@@ -0,0 +1,53 @@
...
+logger.info(f"We will be getting data from {bucket_guide.num_rows} buckets...")
+
+# Define Destination Bucket
+dest_bucket = 'tmc-data'
+
+# Moving Files from Vendor s3 Bucket to Destination s3 Bucket
+for x in bucket_guide:
+
+    logger.info(f"Working on files for {x['s3_bucket']}...")
+    keys = s3_hustle.list_keys(x['s3_bucket'])
```
ydamit: supposed to be `s3_vendor` right?

elyse-weiss:
@ydamit I made the changes, can you do final approval?

Issue #55, opened 2019-10-11
Redshift: One Value Query


jburchard:
Often times, when we run a query, we don't need a Parsons table, we just need a single value. This would be a convenience method that allows for this.

PR #77, opened 2019-11-07
Table: Attribute to return first value in table.

jburchard:
This is a convenience method to return the first value in a Parsons Table. 

This is intended to be used with database queries in which a single value is expected to be returned. Ex: `tbl = rs.query('select count(*) from my_schema.my_table')`.

While this required accessing the index previously, now you just run `tbl.first` to access the first value in the table.

Not sure that `value` is the best name for this attribute and open to suggestions.

Intended to address Issue #55.

### DIFF
--- a/docs/table.rst
+++ b/docs/table.rst
@@ -112,6 +112,8 @@ Parsons Table Attributes
       - A list of column names in the table
     * - ``.data``
       - The actual data (rows) of the table, as a list of tuples (without field names)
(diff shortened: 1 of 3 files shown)


eliotst:
How about `first`?

jburchard:
Far better. I'll make the change.

jburchard:
@eliotst - Ready for final review.

PR #80, opened 2019-11-08
rework temp file handling

eliotst:
This commit changes how we manage temporary files in Parsons. We
were using `tempfile.NamedTemporaryFile` from the Python standard
library, but this was causing errors with some Parsons methods
when run on Windows.

The problem is that `NamedTemporaryFile` opens a file handle and
keeps it open until it is cleaned up. However, the
`parsons.utilities.files.create_temp_file` function that used the
`NamedTemporaryFile` was returning the path to the temp file
created, and not to the file handle itself. As such, users of
these temp files had to open them to be able to read data from and
write data to the temp file. On Windows opening the same file
twice can result in an error.

This commit creates a new class `TempFilePath` modelled after
`NamedTemporaryFile` that creates a temp file on disk, but does
not hold onto an open file handle. Instead it just returns the
path to the newly created temp file, which can then be accessed
by others. When the `TempFilePath` instance is cleaned up (e.g.,
when the current Python process ends), it removes the file from
disk. This should give us the best of both worlds.

### DIFF
--- a/parsons/utilities/files.py
+++ b/parsons/utilities/files.py
@@ -293,6 +293,9 @@ class TempFile:
         """
         # Only try to unlink if we have a valid file path and we haven't yet called close.
         if self.name and not self.remove_called:
-            unlink(self.name)
+            try:
+                unlink(self.name)
+            except FileNotFoundError:
+                pass  # if the file isn't found, our work is done
 
         self.remove_called = True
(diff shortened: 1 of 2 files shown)


eliotst:
Not yet ready for merge. Still need to fully test on Windows.

jburchard:
Really cool. Thank you. +1

PR #81, opened 2019-11-09
VAN: Upload Scores Method

jburchard:
This is meant to replace the old load scores methods. The goal is to make, what is a multistep process a little bit easier.

The old workflow, `VAN.upload_create_load_multi()` required that convert the Parsons table to a compressed zip. It also required that you post the file to the internet, for VAN to grab as well. You also had to calculate averages on your own for the auto approve. Those were a lot of steps.

`VAN.upload_scores()` now allows you to pass in a Parsons table and then does all of the rest of the work for you.

From the example documentation:

```
from parsons import VAN
van = VAN(db='MyVoters') # API key stored as an environmental variable

# If you don't know the id, you can run van.get_scores() to list the
# slots that available and their associated score ids.
score_id = 9999

# Load the Parsons table with the scores. The first column of the table
# must be the person id (e.g. VANID). You could create this from Redshift or
# another source.
tbl = Table.from_csv('winning_scores.csv')

# Specify the score id slot and the column name for each score.
config = [{'id': score_id, 'column': 'winning_model'}]

# If you have multiple models in the same file, you can load them all at the same time.
# In fact, VAN recommends that you do so to reduce their server loads.
config = [{'id': 5555, 'column': 'score1'}, {'id': 5556, 'column': 'score2'}]

# The score file must posted to the internet. This configuration uses S3 to do so. In this
# example, your S3 keys are stored as environmental variables. If not, you can pass them
# as arguments.
job_id = van.upload_scores(tbl, config, url_type='S3', email='info@tmc.org', bucket='tmc-fake')
```

This introduces a new utility `utilities.cloudstorage.post_file()` which is a generalizable method to post files to a public url on the internet. Currently, you need S3 in order to use the `upload_scores()` method, but in the future, we could easily add the ability to do so with Google Cloud Storage or an SFTP.


jburchard:
@elyse-weiss - Would love your thoughts on this, as you've been the biggest user of the `canales` score loading scripts.

@eliotst - Interested in thoughts about how we generalize the cloud storage piece and if this makes sense. Also, just general review of course.

jburchard:
Yes, it does.

It actually automatically calculates the average of the score column. If it is within the tolerance, then it automatically approves.

PR #82, opened 2019-11-11
Salesforce class

ydamit:
@jburchard , here's the very basic start to the Salesforce class. I haven't created any unit tests yet, but I have done linting. Do we want the methods to return Parsons Tables, or is our preference to return lists of dicts and let folks make them into Parsons Tables as needed?


parsons/salesforce/salesforce.py
```diff
@@ -0,0 +1,151 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Salesorce:
+    """
+    `Args:`
+        username: str
+            The Salesforce username (usually an email address).
+            Not required if ``SALESFORCE_USERNAME`` env variable is passed.
```
jburchard: Silly thing, but we are not doing line breaks between these sentences for other classes.

parsons/salesforce/salesforce.py
```diff
@@ -0,0 +1,151 @@
...
+        password: str
+            The Salesforce password.
+            Not required if ``SALESFORCE_PASSWORD`` env variable is passed.
+        security_token: str
+            The Salesforce security token that can be acquired or reset in
+            Settings > My Personal Information > Reset My Security Token.
+            Not required if ``SALESFORCE_SECURITY_TOKEN`` env variable is passed.
+        domain: str
+            If the Saleforce instance is a sandbox, set to `'test'`. Otherwise, not required.
+            Can also be passed through ``SALESFORCE_DOMAIN`` env variable.
```
jburchard: Might it make more sense to have the argument be `test_environment` boolean or something like that if that is the only instance in which it might be set?

parsons/salesforce/salesforce.py
```diff
@@ -0,0 +1,151 @@
...
+            username=self.username,
+            password=self.password,
+            security_token=self.security_token,
+            domain=self.domain
+        )
+
+    def query(self, soql):
+        """
+        `Args:`
+            soql: str
```
jburchard: Typo: `sql`

parsons/salesforce/salesforce.py
```diff
@@ -0,0 +1,151 @@
...
+                The desired query in Salesforce SOQL language (SQL with additional limitations).
+                For reference, see `Salesforce SOQL documentation<https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql.htm>`_.
+        `Returns:`
+            list of dicts with Salesforce data
+        """ # noqa: E501,E261
+
+        return self.client.query_all(soql)
```
jburchard: This should be returning a `Table` object, I think. Also, is `soql` not a typo?

parsons/salesforce/salesforce.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def insert(self, object, data_table):
+        """
+        Insert new records of the desired object into Salesforce
+
+        `Args:`
+            object: str
+                The API name of the type of records to insert.
+                Note that custom object names end in `__c`
```
jburchard: No line break needed.

parsons/salesforce/salesforce.py
```diff
@@ -0,0 +1,151 @@
...
+            data_table: Parsons Table
+                Data for records inserted. Column names must match object field API names, though
+                case need not match. Note that custom field names end in `__c`.
```
jburchard: For other connectors, I've been doing it as:

`tbl: obj`
       A Parsons table. 

Also, does the order need to match. If so, specify. If not, then specify that too.

parsons/salesforce/salesforce.py
```diff
@@ -0,0 +1,151 @@
...
+        `Returns:`
+            list of dicts that have the following data:
+            * success: boolean
+            * created: boolean (if new record is created)
+            * id: str (id of record created, if successful)
+            * errors: list of dicts (with error details)
+        """
+
+        dicts = [row for row in data_table]
+        return getattr(self.client.bulk, object).insert(dicts)
```
jburchard: **This should apply to all of the methods**

You should include some logging in here. Perhaps consider the following pattern instead. This is assuming that you rename `data_table` to `tbl`.

```
r = self.client.bulk.insert(tbl.to_dicts())
logger.info('Inserted or updated {tbl.num_rows} to {object})
return getattr(r)
```

parsons/salesforce/salesforce.py
```diff
@@ -0,0 +1,151 @@
...
+            `Returns:`
+                list of dicts that have the following data:
+                * success: boolean
+                * created: boolean (if new record is created)
+                * id: str (id of record deleted, if successful)
+                * errors: list of dicts (with error details)
+        """
+
+        dicts = [row for row in id_table]
+        if hard_delete is True:
```
jburchard: `True` is implied and doesn't need to be specified.

`if hard_delete:`

jburchard:
The failing test issue is my fault. I'll work on that later today, so you can merge in master, but no need to do anything about it now.

parsons/salesforce/salesforce.py
ydamit: But in my case, this will exceed 100 characters. Should I add a `noqa E501`?

jburchard: I was just thinking like a normal line break prior to hitting 100 characters.

```
The Salesforce username (usually an email address). Not required if ``SALESFORCE_USERNAME`` env variable is passed.
```

parsons/salesforce/salesforce.py
```diff
@@ -0,0 +1,151 @@
+from simple_salesforce import Salesforce as _Salesforce
+from parsons.utilities import check_env
+# from parsons import Table
```
jburchard: Assuming you are going to uncomment this if you are returning a Table.

parsons/salesforce/salesforce.py
ydamit: It's not a typo - Salesforce actually refers to it as SOQL: Salesforce Object Query Language. But if you think it's confusing, I can definitely change it. I just thought it would be a good idea to specify that it's *not* SQL because it can be more limited than SQL.

parsons/salesforce/salesforce.py
ydamit: As above, this would put the character count just over 100.

parsons/salesforce/salesforce.py
ydamit: I decided to improve this a bit so that the logging is sensitive to how many records are processed successfully.

ydamit:
@jburchard , I must have screwed this up when I rebased the branch to reflect the linting you did. It's not showing anything when I run `flake8` locally...

parsons/salesforce/salesforce.py
jburchard: Oh, got it. Well, then leave it as is.

parsons/salesforce/salesforce.py
jburchard: See above suggestion.

parsons/salesforce/salesforce.py
jburchard: Sweet.

parsons/salesforce/salesforce.py
jburchard: Please add the `test_environment` arg to the doc strings.

parsons/salesforce/salesforce.py
```diff
@@ -0,0 +1,191 @@
...
+        s = [x for x in r if x.get('success') is True]
+        logger.info(
+            f'Successfully deleted {len(s)} out of {id_table.num_rows} records from {object}'
+        )
+        return r
+
+    @property
+    def client(self):
+        """
+        Get the Salesforce client to use for making all calls.
```
jburchard: You might want to include a link to the simple-sales-force documentation so folks can see other advanced methods that they can leverage.

parsons/salesforce/salesforce.py
```diff
@@ -0,0 +1,191 @@
...
+                For reference, see `Salesforce SOQL documentation<https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql.htm>`_.
+        `Returns:`
+            list of dicts with Salesforce data
+        """ # noqa: E501,E261
+
+        q = Table(self.client.query_all(soql))
+        logger.info(f'Found {q.num_rows} results')
+        return q
+
+    def insert(self, object, data_table):
```
jburchard: I wonder if in keeping with our overly explanatory naming conventions, should the method be called `insert_record`. If so, I think that the same would and should apply to the other methods.

PR #94, opened 2019-11-15
update version of joblib

eliotst:
This commit updates the version of joblib from 0.11 to 0.13.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ httplib2==0.12.0
 validate-email==1.3
 paramiko==2.4.2
 xmltodict==0.11.0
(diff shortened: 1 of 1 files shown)


jburchard:
Looks like the failure is now for `civis-python`. However, their [latest release](https://github.com/civisanalytics/civis-python/releases), 1.11 allows you to use `joblib 0.13`

eliotst:
@jburchard 

Builds and tests passed. I did a quick test locally, and I was able to run a simple query using the Civis client with 1.11.0 . Not sure if we want to do any more smoke testing before we move forward with the Civis version upgrade?

PR #101, opened 2019-11-19
Return additional data with get_columns

dannyboy15:
[Decimal types in Redshift](https://docs.aws.amazon.com/redshift/latest/dg/r_Numeric_types201.html#r_Numeric_types201-decimal-or-numeric-type) are defined as `decimal(precision, scale)`. With this change, both the precision and the scale are returned.


eliotst:
The lint fixes are in master, so I am going to merge and hope that the fixes aren't overwritten.

PR #104, opened 2019-11-21
Documentation Cleanup

jburchard:
Clean up some formatting issues.

### DIFF
--- a/parsons/mobilize_america/ma.py
+++ b/parsons/mobilize_america/ma.py
@@ -332,4 +332,3 @@ class MobilizeAmerica(object):
         return Table(self.request_paginate(url,
                                            args={'updated_since': iso_to_unix(updated_since)},
                                            auth=True))
-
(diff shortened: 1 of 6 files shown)


parsons/mobilize_america/ma.py
```diff
@@ -104,14 +102,14 @@ def get_organizations(self, updated_since=None):
         return Table(self.request_paginate(self.uri + 'organizations',
                                            args={'updated_since': iso_to_unix(updated_since)}))
 
-    def get_organizations_promoted(self, organization_id, updated_since=None):
+    # NOT IMPLEMENTED
+    # def get_organizations_promoted(self, organization_id, updated_since=None):
```
eliotst: This is more of a personal preference thing, but generally I prefer to delete commented out code than leave it commented out when merging into master.. unless I know that someone will be coming back fairly soon to uncomment and/or handle it. I feel this way especially when I know that the commented out code is in the git commit history somewhere and can be brought back to life when necessary.

jburchard: Ah, makes sense. I'm down with that.

jburchard:
Ready for one more pass...

PR #105, opened 2019-11-22
Gsheet user_entered_value option

ydamit:
Need this in order to be able to enter formulas and not have Google Sheets convert them to text. Added a test function but wasn't sure if it was ok for me to run a live test or not, but I tested the functionality separately.

### DIFF
--- a/parsons/google/google_sheets.py
+++ b/parsons/google/google_sheets.py
@@ -177,6 +177,8 @@ class GoogleSheets(object):
                 Parsons table
             sheet_index: int (optional)
                 The index of the desired worksheet
+            user_entered_value: bool (optional)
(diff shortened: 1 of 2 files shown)


parsons/google/google_sheets.py
```diff
@@ -177,6 +177,8 @@ def append_to_sheet(self, spreadsheet_id, table, sheet_index=0):
                 Parsons table
             sheet_index: int (optional)
                 The index of the desired worksheet
+            user_entered_value: bool (optional)
+                If True, will submit cell values as entered (required for entering formulas)
```
jburchard: Could you be a bit more descriptive as to what things look like when this is a set to False? Like what is the output in the GoogleSheet?

PR #106, opened 2019-11-22
VAN: Upload Saved Lists

jburchard:
The current VAN API (v4) does not have and endpoint in order to upload saved lists. However, the old SOAP (v3) based API does. This method leverages the v3 API in order to push lists to VAN. 

Methods of note:

`VAN.upload_saved_list()` - This leverages the url_type in order to store a file in clould storage, similar to scores.

`VAN.connection._soap_client()` - A SOAP client that is generated when this method is called. It is not created until the call.


parsons/ngpvan/saved_lists.py
```diff
@@ -64,6 +67,82 @@ def download_saved_list(self, saved_list_id):
...
+
+        # VAN errors for this method are not particularly useful or helpful. For that reason, we
+        # will check that the folder exists and if the list already exists.
+        logger.info('Validating folder id and list name.')
+        if folder_id not in [x['folderId'] for x in self.get_folders()]:
+            raise ValueError("Folder does not exist or is not shared with API user.")
+
+        if not replace:
+            if list_name in [x['name'] for x in self.get_saved_lists(folder_id)]:
+                raise ValueError("Saved list already exists. Set to replace or change name."
```
dannyboy15: I would recommend something like: 'Set to replace to `True` or provide a different list name.

parsons/ngpvan/saved_lists.py
```diff
@@ -64,6 +67,82 @@ def download_saved_list(self, saved_list_id):
...
+                                 " name.")
```
dannyboy15: Typo.

parsons/ngpvan/saved_lists.py
jburchard: Fixed.

parsons/ngpvan/saved_lists.py
jburchard: Fixed.

elyse-weiss:
@jburchard am I reviewing or are we trying to test first?

parsons/ngpvan/saved_lists.py
```diff
@@ -64,6 +67,82 @@ def download_saved_list(self, saved_list_id):
...
+        `Returns:`
+            dict
+                Upload results information included the number of matched and saved
+                records in your list.
+        """
+
+        # Move to cloud storage
+        file_name = str(uuid.uuid1()) + '.zip'
+        public_url = cloud_storage.post_file(tbl, url_type, file_path=file_name, **url_args)
+        csv_name = files.extract_file_name(file_name, include_suffix=False) + '.csv'
```
eliotst: Minor thing, but if you just have a temp variable set to `uuid.uuid1()`, you could use it to build the zip and csv filename and wouldn't need to extract here.

Issue #10, opened 2019-09-06
VAN: validation should be optional for upsert_person


kbenker:
When you pass a record to `upsert_person`, it validates that "the minimum combination of fields were passed." The VAN API doesn't actually reject records without this info; it will just always create new records if there's insufficient info to match on. 

I understand that some users may want this validation to avoid adding duplicates, but others may still want to add new records for people they have sparser information on.

jburchard:
@ydamit - FYI

jburchard:
@kbenker So is there any minimum number of fields?

PR #107, opened 2019-11-23
VAN: Fix Bugs

jburchard:
- ~~Remove validation requirement of `van_upsert_person()` per issue #10.~~ - Saving for a future PR. 

- Fix bug so that `van.find_person()` works properly. 

- Remove `scores` from default `van.get_person()` returned value. This is not a standard field which is permissioned to all users, so it was throwing an error with the default setting. (@dannyboy15 for viz)

### DIFF
--- a/parsons/ngpvan/people.py
+++ b/parsons/ngpvan/people.py
@@ -67,6 +67,14 @@ class People(object):
         Create or update a person record.
 
         .. note::
+            Person find must include the following minimum combinations.
+            - first_name, last_name, email
+            - first_name, last_name, phone
+            - first_name, last_name, zip5, date_of_birth
+            - first_name, last_name, street_number, street_name, zip5
+            - email_address
+
+        .. note::
             The arguments that can be passed are a selection of the possible values that
             can be used in a search. A full list of possible values can be found
             `here <https://developers.ngpvan.com/van-api#match-candidates>`_. To use these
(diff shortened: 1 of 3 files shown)


parsons/ngpvan/people.py
```diff
@@ -109,16 +100,18 @@ def upsert_person(self, first_name=None, last_name=None, date_of_birth=None, ema
...
 
         return self._people_search(first_name, last_name, date_of_birth, email, phone,
-                                   street_number, street_name, zip, match_map, create=True)
+                                   street_number, street_name, zip, match_map, create=True,
+                                   validate=False)
 
     def _people_search(self, first_name=None, last_name=None, date_of_birth=None, email=None,
                        phone=None, street_number=None, street_name=None, zip=None, match_map=None,
-                       create=False):
+                       create=False, validate=True):
```
ydamit: Based on what I'm seeing here, do we strictly _need_ `validate` if it's always going to be the opposite of `create`? Or is there a plan to introduce new functionality that would break this complementarity?

parsons/ngpvan/people.py
```diff
@@ -109,16 +100,18 @@ def upsert_person(self, first_name=None, last_name=None, date_of_birth=None, ema
...
         # Internal method to hit the people find/create endpoints
 
         # Ensure that the minimum combination of fields were passed
-        self._valid_search(first_name, last_name, email, phone, date_of_birth, street_number,
-                           street_name, zip, match_map)
+        if validate:
+            self._valid_search(first_name, last_name, email, phone, date_of_birth, street_number,
+                               street_name, zip, match_map)
```
ydamit: I recognize this is was an attempt to quickly resolve one small feature request and one small bug, but there are larger issues with `_valid_search()` and it's possible it's worth resolving some of them now.
1. `street_number` and `street_name` are not processed anywhere in `_people_search()`, and I suspect they are relics of an older API version, because the Address object in the VAN documentation isn't broken out that way. I wonder if we shouldn't just get rid of them entirely until we can get an updated answer from VAN about how the validation logic works.
2. `_valid_search()` currently ignores `match_map` entirely. This might not be as urgent a bug to fix now that `upsert_person()` bypasses validation, but it still unnecessarily limits `find_person()`

jburchard:
That's a really good point. I might punt on that change for the moment.

How would you feel about just dealing with the first two issues, as I would argue that they are pretty urgent?

ydamit:
Happy to try if member work allows! Can I ask for help in two ways:
1. Can you specify which specific issues I should and should not try to address
2. Should I just contribute to this same branch?

jburchard:
Oh, no I meant that I could handle it. Just wanted to get sign off before I move forward on this.

PR #110, opened 2019-11-26
URI option in van_connector and match map addition in _valid_search

kcym-3c:


parsons/ngpvan/people.py
```diff
@@ -146,23 +146,58 @@ def _people_search(self, first_name=None, last_name=None, date_of_birth=None, em
...
-                             Person find must include the following minimum
-                             combinations to conduct a search.
-                                - first_name, last_name, email
-                                - first_name, last_name, phone
-                                - first_name, last_name, zip, dob
-                                - first_name, last_name, street_number, street_name, zip
-                                - email
-                            """)
-
+        if not match_map:
```
ydamit: I _think_ the best way to accomplish this would actually be to hook onto the logic already present in `_people_search()`. My instinct is that we should call `_valid_search()` towards the end of `_people_search()`, and just pass the variable `json`, which will either be a collection of the individual field arguments, or, if provided, `match_map`. Then `_valid_search()` would just need to check the `json` variable. The main down-side to this approach is that we'd need to flatten `json` to get at all the values, but I think it would still save many lines of code and would be overall much more readable (though I could certainly be convinced otherwise).

parsons/ngpvan/van_connector.py
```diff
@@ -5,13 +5,13 @@
...
 
-URI = 'https://api.securevan.com/v4/'
 SOAP_URI = 'https://api.securevan.com/Services/V3/ListService.asmx?WSDL'
 
 
 class VANConnector(object):
 
-    def __init__(self, api_key=None, auth_name='default', db=None):
+    def __init__(self, api_key=None, auth_name='default',
+                 db=None, uri='https://api.securevan.com/v4/'):
```
ydamit: I propose a compromise:
```
URI = 'https://api.securevan.com/v4/'
def __init__(self, api_key=None, auth_name='default', db=None, uri=URI):
```

parsons/ngpvan/van_connector.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, api_key=None, auth_name='default', db=None):
             raise KeyError('Invalid database type specified. Pick one of:'
                            ' MyVoters, MyCampaign, MyMembers, EveryAction.')
 
-        self.uri = URI
+        self.uri = uri
```
ydamit: See proposed compromise above

test/test_van/test_people.py
```diff
@@ -60,6 +60,53 @@ def test_valid_search(self):
...
+                               None, None, match_map_3)
+
+        # Successful with match_map FN/LN/STREETNAME/STREETNUMBER/ZIP
+        match_map_4 = {
+            'firstName': 'Barack',
+            'lastName': 'Obama',
+            'addresses':
+            [{
+                'zipOrPostalCode': 20009,
+                'addressLine1': 'glenwood drive'
```
ydamit: This will not be helpful as part of these testing functions until we replace all instances of `street_number` and `street_name` arguments in the people method with `addressLine1`. There is currently nothing in our own validation process that recognizes street name or number from `addressLine1`.

parsons/ngpvan/people.py
kc-yang2020: I like that approach, should we even give the option for match map? The user would have to know exactly how people objects are structured in VAN, if we do receive a match map maybe we can have _people_search check the structure and throw an error if it's not in the right format. For example if zipOrPostalCode is not in the addresses value.

ydamit: That sounds like a bunch of extra work, but as someone who's had to use `match_map` I know I'd be grateful for the extra checking! It does seem like the checking ideally _should_ happen in `_valid_search()`, but that's mostly for thematic consistency. If it's much simpler to put that into `_people_search()` I certainly wouldn't complain.
I just made sure that the `vanid_endpoints` branch is still available here on Github. There were problems with the way I tried to handle this issue (which may be tangled because that PR tried to fix too many things at once), but on the off chance my code can be of any help to you: https://github.com/move-coop/parsons/blob/vanid_endpoints/parsons/ngpvan/people.py

PR #114, opened 2019-12-03
Postgres Connector

jburchard:
The new Postgres connector creates the following new methods:

`Postgres.query()` - Query the database.
`Postgres.copy()` - Copy a parsons table into the database.

Some notes:

- Under the hood, I've created a new `PostgresCore` class that the `Postgres` class utilitizes. In a future PR, the `Redshift` class will be refactored so that it also relies on the `PostgresCore` class.

- There are some advanced copy arguments that can/should be added to the `Postgres.copy()` method in furture PRs.


parsons/databases/postgres/postgres_core.py
```diff
@@ -0,0 +1,242 @@
...
+            bool
+                True if the table needs to be created, False otherwise.
+        """
+
+        if if_exists not in ['fail', 'truncate', 'append', 'drop']:
+            raise ValueError("Invalid value for `if_exists` argument")
+
+        exists = self.table_exists_with_connection(table_name, connection)
+
+        if exists and if_exists in ['fail', 'truncate', 'append']:
```
eliotst: Double check me on this, but I think you rewrite as:

```
if exists:
    if exists == 'fail':
        ...
    if exists == 'truncate':
        ...
else:
    if exists == 'drop':
        ...
    return True

return False
```

It simplifies the branching a little bit.

parsons/databases/postgres/postgres_create_table.py
```diff
@@ -0,0 +1,242 @@
...
+                  'OFFSET', 'OLD', 'ON', 'ONLY', 'OPEN', 'OR', 'ORDER', 'OUTER', 'OVERLAPS',
+                  'PARALLEL', 'PARTITION', 'PERCENT', 'PERMISSIONS', 'PLACING', 'PRIMARY', 'RAW',
+                  'READRATIO', 'RECOVER', 'REFERENCES', 'RESPECT', 'REJECTLOG', 'RESORT', 'RESTORE',
+                  'RIGHT', 'SELECT', 'SESSION_USER', 'SIMILAR', 'SOME', 'SYSDATE', 'SYSTEM',
+                  'TABLE', 'TAG', 'TDES', 'TEXT255', 'TEXT32K', 'THEN', 'TIMESTAMP', 'TO', 'TOP',
+                  'TRAILING', 'TRUE', 'TRUNCATECOLUMNS', 'UNION', 'UNIQUE', 'USER', 'USING',
+                  'VERBOSE', 'WALLET', 'WHEN', 'WHERE', 'WITH', 'WITHOUT']
+
+
+class PostgresCreateTable(object):
```
dannyboy15: Could this class be used for generating create statements for other databases? If so, I would suggest generalizing it, maybe `DatabaseCreateTable`? It could be initialized with a database type if slight differences are expected.

```
    def __init__(self, db_type):
        self.db_type = db_type
```

parsons/databases/postgres/postgres_core.py
```diff
@@ -0,0 +1,242 @@
...
+                Whether to commit the transaction immediately. If ``False`` the transaction will
+                be committed when the connection goes out of scope and is closed (or you can
+                commit manually with ``connection.commit()``).
+
+        `Returns:`
+            Parsons Table
+                See :ref:`parsons-table` for output options.
+        """
+
+        # To Do: Have it return an ordered dict to return the
```
eliotst: I think `OrderedDict` is about returning dictionaries where the columns are in the right order. The query should return the rows in the right order because the list's order should be stable.

parsons/databases/postgres/postgres_create_table.py
dannyboy15: Also realizing this doesn't actually create the table so maybe should be renamed accordingly.

parsons/databases/postgres/postgres_create_table.py
```diff
@@ -0,0 +1,242 @@
...
+    def create_statement(self, tbl, table_name, padding=None, distkey=None, sortkey=None,
+                         varchar_max=None, varchar_truncate=True, columntypes=None):
+        # Generate a table create statement. Distkeys and sortkeys are only used by
+        # Redshift and should not be passed when generating a create statement for
+        # Postgres.
+
+        # Validate and rename column names if needed
+        tbl.table = petl.setheader(tbl.table, self.column_name_validate(tbl.columns))
+
+        if tbl.num_rows == 0:
```
eliotst: Minor thing, but you could move this up to before the previous line because you don't need to rename the columns if you aren't going to create the table.

parsons/databases/postgres/postgres_create_table.py
```diff
@@ -0,0 +1,242 @@
...
+        pass
+
+    def create_statement(self, tbl, table_name, padding=None, distkey=None, sortkey=None,
+                         varchar_max=None, varchar_truncate=True, columntypes=None):
+        # Generate a table create statement. Distkeys and sortkeys are only used by
+        # Redshift and should not be passed when generating a create statement for
+        # Postgres.
+
+        # Validate and rename column names if needed
+        tbl.table = petl.setheader(tbl.table, self.column_name_validate(tbl.columns))
```
eliotst: It seems an odd side effect of calling `to_redshift` that the table itself is mutated by having its columns renamed. It'd be nice to create a copy of the table before renaming things, though I'm not sure it's worth it if there's not a way to do it without copying the WHOLE `petl` table.

parsons/utilities/check_env.py
```diff
@@ -1,9 +1,13 @@
...
-def check(env, field=None):
-    """Check if an envrionment variable is set for __init__"""
-    if not field:
+def check(env, field, *args):
+    """
+    Check if an environment variable has been set. If it has not been set
+    and the passed field or arguments have not been passed, then raise an
+    error.
+    """
+    if not field and not args:
```
eliotst: Above, you are passing in `pgpass` and I think you are trying to skip the `os.environ` check if `pgpass` is `True`. But`not args` will always be `True` if you pass in any additional args, even if the value is `False`.

test/test_databases/test_postgres.py
```diff
@@ -0,0 +1,179 @@
...
+        exp_sql = """create table tmc.test (\n  "id" smallint,\n  "name" varchar(5)) \ndistkey(ID) ;"""  # noqa: E501
+        self.assertEqual(sql, exp_sql)
+
+        # Assert that an error is raised by an empty table
+        empty_table = Table([['Col_1', 'Col_2']])
+        self.assertRaises(ValueError, self.pg.create_statement, empty_table, 'tmc.test')
+
+# These tests interact directly with the Postgres database
+
+@unittest.skipIf(not os.environ.get('LIVE_TEST'), 'Skipping because not running live test')
```
dannyboy15: Might be a good idea to add db testing to the CI workflow - https://discuss.circleci.com/t/how-can-i-start-testing-database-connection-in-circleci/21512

test/test_databases/test_postgres.py
```diff
@@ -0,0 +1,179 @@
...
+import datetime
+from test.utils import validate_list
+
+# The name of the schema and will be temporarily created for the tests
+TEMP_SCHEMA = 'parsons_test'
+
+# These tests do not interact with the Postgres Database directly, and don't need real credentials
+
+
+class TestPostgres(unittest.TestCase):
```
dannyboy15: I would suggest renaming it for more clarity. `TestPostgresCreateTable`

parsons/databases/postgres/postgres_create_table.py
jburchard: Complete.

parsons/databases/postgres/postgres_create_table.py
jburchard: I totally agree on this. I'd like to kick it to another PR, because I think that what I really need is a `Copy` class that can store things like this and then revert once complete/fails.

parsons/databases/postgres/postgres_core.py
jburchard: Removed.

parsons/databases/postgres/postgres_core.py
jburchard: Cleaned it up a bit, but there is still a bunch going on unfortunately.

parsons/utilities/check_env.py
jburchard: It was intentional, though admittedly a little hacky.

parsons/databases/postgres/postgres_core.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        with self.connection() as connection:
+            return self.table_exists_with_connection(table_name, connection, view)
+
+    def table_exists_with_connection(self, table_name, connection, view=True):
+        table_name = table_name.lower().split('.')
+
+        # Check in pg tables for the table
+        sql = """select count(*) from pg_tables where schemaname='{}' and
+                 tablename='{}';""".format(table_name[0], table_name[1])
```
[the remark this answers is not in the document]
jburchard: Great idea. Thank you for the code suggestion.

parsons/databases/postgres/postgres_core.py
```diff
@@ -0,0 +1,242 @@
...
+
+        `Returns:`
+            Psycopg2 `connection` object
+        """
+
+        # Create a psycopg2 connection and cursor
+        conn = psycopg2.connect(user=self.username, password=self.password,
+                                host=self.host, dbname=self.db, port=self.port,
+                                connect_timeout=self.timeout)
+        yield conn
```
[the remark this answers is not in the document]
eliotst: I think this should work:
```
try:
    yield conn
catch:
    conn.rollback()
    raise
else:
    conn.commit()
finally:
    conn.close()
```

That will commit and close if there's no error, rollback and close if there is an error (and re-raise the error so it bubbles up to the caller).

parsons/databases/postgres/postgres_core.py
```diff
@@ -0,0 +1,242 @@
...
+
+        conn.commit()
+        conn.close()
+
+    @contextmanager
+    def cursor(self, connection):
+        cur = connection.cursor(cursor_factory=psycopg2.extras.DictCursor)
+        yield cur
```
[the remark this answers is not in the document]
eliotst: Here, I think you just need:

```
try:
    yield cur
finally:
    cur.close()
```

Just to make sure that the cursor is closed even if there's an error.

parsons/databases/postgres/postgres_core.py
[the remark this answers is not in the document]
jburchard: Fixed.

parsons/databases/postgres/postgres_core.py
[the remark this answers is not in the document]
jburchard: Fixed.

parsons/databases/postgres/postgres_core.py
```diff
@@ -0,0 +1,242 @@
...
+        # Check in pg tables for the table
+        sql = """select count(*) from pg_tables where schemaname='{}' and
+                 tablename='{}';""".format(table_name[0], table_name[1])
+
+        # TODO maybe convert these queries to use self.query_with_connection
+
+        with self.cursor(connection) as cursor:
+
+            cursor.execute(sql)
+            result = cursor.fetchone()[0]
```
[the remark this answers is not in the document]
jburchard: I think that I understand this and fixed.

test/test_databases/test_postgres.py
jburchard: Yes. That could be a good future PR.

test/test_databases/test_postgres.py
jburchard: Fixed.

parsons/utilities/check_env.py
eliotst: So, I think that even if the pgpass file doesn't exist, `not args` will be False, and so you will never go to the environment. Because if you pass in any additional arguments to `check`, `args` is truthy. Could you just add a new optional positional argument that is `skip` or something that means "just give me `field` back"?

parsons/databases/postgres/postgres_create_statement.py
```diff
@@ -0,0 +1,242 @@
...
+        # Populate empty values for the columns
+        for col in table.columns:
+            longest.append(0)
+            type_list.append('')
+
+        for row in cont:
+            for i in range(len(row)):
+                # NA is the csv null value
+                if type_list[i] == 'varchar' or row[i] == 'NA':
+                    pass
```
eliotst: If we `pass` here, does that mean we aren't going to count the characters in this string to increase the size of our `longest`? You could make the `else` below an `elif` and check whether it's `varchar` there so.

parsons/databases/postgres/postgres_create_statement.py
```diff
@@ -0,0 +1,242 @@
...
+                else:
+                    var_type = self.data_type(row[i], type_list[i])
+                    type_list[i] = var_type
+                # Calculate width
+                if len(str(row[i]).encode('utf-8')) > longest[i]:
+                    longest[i] = len(str(row[i]).encode('utf-8'))
```
eliotst: It's probably not a big deal.. the cost of `encode('utf-8')` is probably not that high, and you are only doing this when you have a longer string than before, but you could store the `len(str(row[i]).encode('utf-8'))` value in a variable and not have to calculate twice.

jburchard: Done.

parsons/databases/postgres/postgres_create_statement.py
eliotst: You're right. Sorry. I read the `pass` as `continue`.

parsons/databases/postgres/postgres.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class Postgres(PostgresCore):
+    """
+    A Postgres class to connect to database. Credentials can be passed from a ``.pgpass`` file
+    stored in your home directory, environmental variables or passed arguments.
+
+    Args:
+        username: str
+            Required if env variable ``PGUSERNAME`` not populated
```
tonywhittaker: This should be `PGUSER` instead of `PGUSERNAME` per the PostgreSQL documentation

> [PGUSER behaves the same as the user connection parameter.](https://www.postgresql.org/docs/12/libpq-envars.html)

parsons/databases/postgres/postgres.py
```diff
@@ -0,0 +1,74 @@
...
+        password: str
+            Required if env variable ``PGPASSWORD`` not populated
+        host: str
+            Required if env variable ``PGHOST`` not populated
+        db: str
+            Required if env variable ``PGDB`` not populated
```
tonywhittaker: Similar to above this should be `PGDATABASE` instead of `PGDB`

>[PGDATABASE behaves the same as the dbname connection parameter.](https://www.postgresql.org/docs/12/libpq-envars.html)

jburchard:
@tonywhittaker - Final changes made. Approve?

PR #127, opened 2019-12-18
bigquery: add copy functionality

eliotst:
This commit adds a `copy` method to the
`parsons.google.google_bigquery.GoogleBigQuery` connector. The
`copy` method can be used to load a Parsons table into a BigQuery
table by uploading the file to Google Cloud Storage.

### DIFF
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -72,7 +72,7 @@ class GoogleBigQuery:
                 The GoogleCloudStorage Connector to use for loading data into Google Cloud Storage.
             job_config: object
                 A LoadJobConfig object to provide to the underlying call to load_table_from_uri
-                on the BigQuery client.
+                on the BigQuery client. The function will create its own if not provided.
             **load_kwargs: kwargs
                 Arguments to pass to the underlying load_table_from_uri call on the BigQuery
                 client.
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -3,8 +3,8 @@ from google.cloud import exceptions
 from parsons import Table
 from parsons.google.utitities import setup_google_application_credentials
 from parsons.google.google_cloud_storage import GoogleCloudStorage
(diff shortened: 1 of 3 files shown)


parsons/google/google_bigquery.py
```diff
@@ -44,6 +50,85 @@ def __init__(self, app_creds=None, project=None, location=None):
...
+            if_exists: str
+                If the table already exists, either ``fail``, ``append``, ``drop``
+                or ``truncate`` the table.
+            temp_gcs_bucket: str
+                The name of the Google Cloud Storage bucket to use to stage the data to load
+                into BigQuery. Required if `GCS_TEMP_BUCKET` is not specified.
+            gcs_client: object
+                The GoogleCloudStorage Connector to use for loading data into Google Cloud Storage.
+        """
+        tmp_gcs_bucket = tmp_gcs_bucket or os.environ.get('GCS_TEMP_BUCKET')
```
jburchard: Out of curiosity, is there a reason why you didn't use the `check_env` [util](https://github.com/move-coop/parsons/blob/master/parsons/utilities/check_env.py)?

jburchard:
One question, actually, are there any advanced configurations for copying that the Google Client can take that we could pass as `**kwargs`. Similar to all of the arguments with Redshift, but a little bit simpler to read?

eliotst:
> One question, actually, are there any advanced configurations for copying that the Google Client can take that we could pass as `**kwargs`. Similar to all of the arguments with Redshift, but a little bit simpler to read?

Yeah. There is a `location` and `project` parameter, which can be used to override the defaults on the client. There's also a `retry` parameter which defines how to retry if there's a failure in the call. I can pass all of that through.

The other thing I will do is add a `job_config` parameter that will allow passing in custom `LoadJobConfig` options.

parsons/google/google_bigquery.py
eliotst: Just forgot about it. I'll swap that out for this.

PR #130, opened 2020-01-11
Table.column_data()

jburchard:
This method returns the rows in a column as a list.

```
tbl = Table([['a', 'b'],['1', 2], ['2', 1]])
lst = tbl.column_data('a')
print (lst)
>>> ['1', '2']
```

Potentially, the most elegant solution would be to be able to use the Pandas pattern for column selection, such as (`tbl['a']`). I'm interested in feedback on this one.


ydamit:
Both seem fine!

PR #137, opened 2020-01-27
Bill com

dekedor:
Adds functionality for accessing the Bill.com API. This class will allow the use to

-list users and customers
-query customer and invoice information
-create customer and invoices
-send invoices


parsons/bill_com/bill_com.py
```diff
@@ -171,80 +202,110 @@ def checkCustomer(self, customer1, customer2):
...
                 The customer's email address
-            messageSubject:
+            message_subject:
                 The subject of the email to send to the customer
-            messageBody:
-                The boyd of the email to send to the customer
+            message_body:
+                The body of the email to send to the customer
+            **kwargs:
+                Any other details for sending the invoice
```
ydamit: I love that you've added these to all the methods, but I don't actually see where these are being incorporated in the code. It's possible we'd never need to, so maybe we can just take this out for now until the need arises.

test/test_bill_com/test_bill_com.py
```diff
@@ -0,0 +1,265 @@
...
+                                                                           'chartOfAccountId': '0',
+                                                                           'departmentId': '0',
+                                                                           'locationId': '0',
+                                                                           'actgClassId': '0',
+                                                                           'jobId': '0',
+                                                                           'description': None,
+                                                                           'taxable': False,
+                                                                           'taxCode': 'Non'}]}]}
+
+        self.fake_invoice_read_json = self.fake_invoice_read_json = \
```
ydamit: This looks like a copy/paste gone wrong (probably my most frequent type of error). I can just fix it in branch.

PR #147, opened 2020-01-31
stop the global package dependencies.

### LINKED ISSUE #146: Too many dependencies
The current pattern in `parsons` is to load every submodule into `parsons/__init__.py`.

This is neither sustainable nor feasible for systems that require smaller upload sizes/memory footprints (e.g. AWS Lambda).  The top of that file makes the goal clear:
"Eg. This allows for: `from parsons import VAN`"
However, it means every install and deploy of parsons requires an ever-increasing number of dependencies.

Proposal:
* Remove the pre-amble -- or at least stop adding new ones.  Admittedly this is backwards incompatible
  * If we must keep backwards incompatibility, then let's create a second lib called `parsons_core` where we move the sub-directories, and then import them in `parsons/` from parsons_core.
* Start documenting the best practice of importing `from parsons_core.nvpvan import VAN`
* Document for low environments that they can install/import with `pip install --no-deps parsons` -- maintain a requirements-core.txt for cross-source dependencies that can/should stay in core.

schuyler1d:
fixes #146 

This is a pretty nutso/invasive patch, however it manages to be backwards compatible but should still work if a subset of dependencies is removed/uninstalled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ We have a Parsons Docker container hosted on [DockerHub](https://cloud.docker.co
 ```
 # VAN - Download activist codes to a CSV
 
-from parsons import VAN
+from parsons.ngpvan import VAN
 van = VAN(db='MyVoters')
 ac = van.get_activist_codes()
 ac.to_csv('my_activist_codes.csv')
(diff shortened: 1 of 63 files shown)


eliotst:
Thanks for this! We will work to review your changes, and we'll get back with comments soon.

schuyler1d:
closing to prefer https://github.com/move-coop/parsons/pull/152

Issue #122, opened 2019-12-11
s3.transfer_bucket - document that original file is not removed from original bucket


elyse-weiss:
Documentation should be clear here.

jburchard:
Should we rename the method to be `s3.copy_bucket()` instead? That seems to be more descriptive of what's going on.

elyse-weiss:
My only objection is I know both Martha and I have used `transfer_bucket()` in a number of scripts, and I think to avoid breaking things we should avoid changing names of functions.

We could add a parameter to allow folks to delete from original bucket if they want?

elyse-weiss:
other proposal: add parameter as to whether to drop from original bucket

eliotst:
One thing to pay attention to here is permissions around being able to delete objects.

PR #148, opened 2020-02-03
Adds an option to delete original files when transferring between buckets

ChrisC:
Adds an additional arg `remove_original` to the `transfer_bucket` method. Default is set to False.

See #122

### DIFF
--- a/parsons/aws/s3.py
+++ b/parsons/aws/s3.py
@@ -351,7 +351,7 @@ class S3(object):
             if remove_original:
                 try:
                     self.remove_file(origin_bucket, origin_key)
-                except:
+                except Exception:
                     pass
 
             if public_read:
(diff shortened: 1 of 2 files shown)


parsons/aws/s3.py
```diff
@@ -346,6 +348,11 @@ def transfer_bucket(self, origin_bucket, origin_key, destination_bucket,
 
             copy_source = {'Bucket': origin_bucket, 'Key': key}
             self.client.copy(copy_source, destination_bucket, dest_key)
+            if remove_original:
+                try:
+                    self.remove_file(origin_bucket, origin_key)
+                except Exception:
+                    pass
```
jburchard: Perhaps we should include some logging to indicate that it failed, or lift up the boto3 error that was raised in the logs.

PR #151, opened 2020-02-05
s3: support extra args in more methods

eliotst:
This commit adds support for passing extra args to more of the
methods in the `parsons.aws.s3.S3` class. The new methods are:

 * copy
 * get_file
 * list_keys


jburchard:
Could you include links to S3 documentation listing potential kwargs?

PR #152, opened 2020-02-05
Avoid hard requirement for all vendor libraries to import parsons

### LINKED ISSUE #146: Too many dependencies

schuyler1d:
Fixes #146 

This is take-2 from #147 after @eliotst  gave feedback.

### DIFF
--- a/parsons/__init__.py
+++ b/parsons/__init__.py
@@ -1,32 +1,143 @@
 # Provide shortcuts to importing Parsons submodules
-# Eg. This allows for: `from parsons import VAN`
-from parsons.ngpvan.van import VAN
-from parsons.targetsmart.targetsmart_api import TargetSmartAPI
-from parsons.targetsmart.targetsmart_automation import TargetSmartAutomation
-from parsons.databases.redshift.redshift import Redshift
-from parsons.aws.s3 import S3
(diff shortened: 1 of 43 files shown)


eliotst:
Looks good to me. It does look like there's a merge conflict with master. Could you pull in the latest from our parsons repo? I'll approve after that.

PR #153, opened 2020-02-06
Freshdesk Connector

jburchard:
This is a connector for [Freshdesk](url) which is an omnichannel customer support platform. This initial PR focuses on core tables to extract from Freshdesk, though could be improved in the future to include methods that create tickets and other objects.

This PR also adds a new method `Table.sort()` since the Freshdesk API sorting is a little bit weird.

### DIFF
--- a/parsons/freshdesk/freshdesk.py
+++ b/parsons/freshdesk/freshdesk.py
@@ -9,7 +9,7 @@ logger = logging.getLogger(__name__)
 PAGE_SIZE = 100
 
 
-class FreshDesk():
+class Freshdesk():
     """
     Instantiate FreshDesk Class
 
--- a/parsons/freshdesk/freshdesk.py
+++ b/parsons/freshdesk/freshdesk.py
@@ -1,0 +1,178 @@
+from parsons.utilities import check_env
+from parsons.utilities.api_connector import APIConnector
+import re
+from parsons import Table
+import logging
+
+logger = logging.getLogger(__name__)
+
+PAGE_SIZE = 100
+
+
+class FreshDesk():
+    """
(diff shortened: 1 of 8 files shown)


parsons/freshdesk/freshdesk.py
```diff
@@ -0,0 +1,178 @@
...
+            data.extend(r.json())
+
+        return data
+
+    def transform_table(self, tbl, expand_custom_fields=None):
+        # Internal method to transform a table prior to returning
+
+        tbl.move_column('id', 0)
+        tbl.sort()
+        if expand_custom_fields:
```
eliotst: I like that this is optional.

parsons/freshdesk/freshdesk.py
```diff
@@ -0,0 +1,178 @@
...
+            tbl.unpack_dict('custom_fields', prepend=False)
```
SorenSpicknall: Is there already an intended use case for unpacking in this flat manner? I don't think there's any reason to hold up the PR over design considerations here, but Eliot had mentioned in a call that perhaps there might be unintended consequences to breaking each dictionary field into a new column when custom fields might change over time. Should we consider doing some checking here or somewhere else to make sure schemas are aligned when the custom columns change?

jburchard: I think that is should be checked, but I would suggest that it be checked as part of a synchronization script that leverages these methods. The use case is for users that don't anticipate their custom fields changing over time and want to 1) avoid additional side tables and joins in a database 2) parsing of nested jsons in a DB.

I set the default to not unpack, so that the data model remains consistent.

PR #158, opened 2020-02-17
Database Sync: Initial Methods

jburchard:
This class allows for for tables to be synchronized between multiple databases. The pattern is the following:

- `DbSync.table_sync_full()` copies _all_ data in a source table to a destination table.
- `DbSync.table_sync_incremental()` copies only new records based on a specified primary key. This only works for tables that are updated incrementally.

**Example Usage**

```
   # Create source and destination database objects
   source_rs = Postgres()
   destination_rs = Postgres()

   # Create db sync object and run sync.
   db_sync = DBSync(source_rs, destination_rs) # Create DBSync Object
   db_sync.table_sync_full('parsons.source_data', 'parsons.destination_data')

   # Create db sync object and run sync.
   db_sync = DBSync(source_pg, destination_pg) # Create DBSync Object
   db_sync.table_sync_incremental('parsons.source_data', 'parsons.destination_data', 'myid')
```

**Notes**

- **Currently, this class only allows for the synchronization of Postgres tables. Future PRs will provide support for Redshift and BigQuery.** 
- The `DbSync.table_sync_incremental()` can be likely be used for in conjunction with a `date_created` column, however this use case has not yet been tested extensively.
- This PR introduces a new `Table` class, which creates a table object. There are many potential future uses for this class outside of the DBSync model.


parsons/databases/db_sync.py
```diff
@@ -0,0 +1,182 @@
...
+            new_row_count = source_tbl.get_new_rows_count(primary_key, dest_max_pk)
+            logger.info(f'Found {new_row_count} new rows in source table.')
+
+            # Copy rows in chunks.
+            copied_rows = 0
+            while copied_rows < new_row_count:
+
+                # Get a chunk
+                rows = source_tbl.get_new_rows(primary_key=primary_key,
+                                               max_value=dest_max_pk,
```
eliotst: I think `max_value` is a bit of a misnomer. Maybe `start_value` would work as a name, since it's the value you are starting from as you paginate through the table?

parsons/databases/db_sync.py
```diff
@@ -0,0 +1,182 @@
...
+            # Question: Should I check that the row counts match too?
+
+        else:
+            # Get count of rows to be copied.
+            new_row_count = source_tbl.get_new_rows_count(primary_key, dest_max_pk)
+            logger.info(f'Found {new_row_count} new rows in source table.')
+
+            # Copy rows in chunks.
+            copied_rows = 0
+            while copied_rows < new_row_count:
```
eliotst: Just a suggestion: I think you could also just loop until `rows.num_rows == 0` from your `get_new_rows` call below. That would make it so you don't need the `get_new_rows_count` call above.

parsons/databases/table.py
```diff
@@ -0,0 +1,137 @@
...
+        Check if the passed primary key column is distinct.
+        """
+
+        sql = f"""
+               SELECT
+               COUNT(DISTINCT {primary_key}) - COUNT(*)
+               FROM {self.table};
+               """
+
+        if self.db.query(sql).first > 0:
```
eliotst: I may be mis-reading, but I think this is supposed to be `< 0`.

parsons/databases/db_sync.py
```diff
@@ -0,0 +1,182 @@
...
+    Sync tables between databases. Works with ``Postgres`` databases.
+
+    `Args:`
+        source_db: Database connection object
+            A database object.
+        destination_db: Database connection object
+            A database object.
+        chunk_size: int
+            The number of rows per transaction copy when syncing a table. The
+            default value is 500,000 rows.
```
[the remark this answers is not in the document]
jburchard: For the time being, I'm going to lower this to 100K. I just want to lower the cost of each individual copy statement, even though it might take longer. It is hard to estimate the optimal size based on rows, since this doesn't take into account the number of columns.

Perhaps in the future, we could come up with some fancy estimations of total cells in a chunk, but that would be for a future PR, I think.

parsons/databases/db_sync.py
jburchard: I'd like to save this for my next PR to think about some optimizations.

PR #159, opened 2020-02-19
Implement redash support

schuyler1d:
For https://redash.io/ for loading data from redash queries.

MoveOn uses Redash internally to manage and save/run SQL queries from several inhouse databases/sources.

### DIFF
--- a/parsons/redash/redash.py
+++ b/parsons/redash/redash.py
@@ -1,0 +1,168 @@
+import json
+import logging
+import requests
+import time
+
+from parsons.etl.table import Table
+
+logger = logging.getLogger(__name__)
+
+
+class Redash(object):
+    """
+    Instantiate Redash Class
+
+    `Args:`
+        base_url: str
+            The base url for your redash instance (excluding the final /)
+        user_api_key: str
+            The user API key found in the User's profile screen
+        pause_time int
+            Specify time between polling for refreshed queries (Defaults to 3 seconds)
+        verify: bool
+            For https requests, should the certificate be verified (Defaults to True)
+    `Returns:`
+        Redash Class
+    """
+    args = {'base_url': None,
+            'user_api_key': None,
+            'pause_time': 3,
+            'verify': True,
+            'query_id': None,
+            'params': None}
+
+    def __init__(self,
+                 base_url="http://localhost",
+                 user_api_key=None,
+                 pause_time=3,
+                 verify=True):
+        self.base_url = base_url
+        self.user_api_key = user_api_key
+        self.pause = pause_time   # seconds to delay
+        self.verify = verify  # for https requests
+        self.session = requests.Session()
+        if user_api_key:
+            self.session.headers.update({'Authorization': f'Key {user_api_key}'})
+
+    def _poll_job(self, session, job, query_id):
+        while job['status'] not in (3, 4):
+            poll_url = '{}/api/jobs/{}'.format(self.base_url, job['id'])
+            response = session.get(poll_url, verify=self.verify)
+            response_json = response.json()
+            job = response_json.get(
+                'job',
+                {'status': 'Error NO JOB IN RESPONSE: {}'.format(json.dumps(response_json))})
+            logger.debug('poll', poll_url, query_id, job['status'], job.get('error'))
+            time.sleep(self.pause)
+
+        if job['status'] == 3:  # 3 = completed
+            return job['query_result_id']
+        elif job['status'] == 4:  # 3 = ERROR
+            raise Exception('Redash Query {} failed: {}'.format(query_id, job['error']))
+
+    def get_fresh_query_results(self, query_id, params=None):
+        """
+        Make a fresh query result and get back the CSV http response object back
+        with the CSV string in result.content
+
+        `Args:`
+            query_id: str or int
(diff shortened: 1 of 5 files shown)


parsons/redash/redash.py
```diff
@@ -0,0 +1,168 @@
...
+            'query_id': None,
+            'params': None}
+
+    def __init__(self,
+                 base_url="http://localhost",
+                 user_api_key=None,
+                 pause_time=3,
+                 verify=True):
+        self.base_url = base_url
+        self.user_api_key = user_api_key
```
eliotst: You should look in the environment for the API key, too. So, `self.user_api_key = check('REDASH_API_KEY', user_api_key)`

parsons/redash/redash.py
```diff
@@ -0,0 +1,168 @@
...
+        self.pause = pause_time   # seconds to delay
+        self.verify = verify  # for https requests
+        self.session = requests.Session()
+        if user_api_key:
+            self.session.headers.update({'Authorization': f'Key {user_api_key}'})
+
+    def _poll_job(self, session, job, query_id):
+        while job['status'] not in (3, 4):
```
eliotst: It might be good to add a timeout arg to the class to allow users to specify the max time they're willing to wait.. and exit the loop accordingly.

parsons/redash/redash.py
```diff
@@ -0,0 +1,168 @@
...
+        `Args:`
+            query_id: str or int
+                The query id of the query
+            params: dict
+                If there are parameters in the query (e.g. "{{datelimit}}" in the query),
+                then this is a dict that will pass the parameters in the POST.
+                We add the "p_" prefix for parameters, so if your query had ?p_datelimit=....
+                in the url, you should just set 'datelimit' in params here.
+        `Returns:`
+            Response Class
```
eliotst: Generally, we prefer Parsons Connectors to return the response data as `Table`s (unless the return value is a scalar). Do you see the user needing something else from the `Response` other than the data in the response body?

parsons/redash/redash.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def get_fresh_query_results(self, query_id, params=None):
+        """
+        Make a fresh query result and get back the CSV http response object back
+        with the CSV string in result.content
+
+        `Args:`
+            query_id: str or int
+                The query id of the query
+            params: dict
```
eliotst: I think the more "Parson-ic" (see what I did there?) way of doing this would be to leverage `**kwargs`. If possible, we also like to drop a link in the docstring to the corresponding API docs on possible args.. though a quick scan of the Redash docs makes me think such a link might not exist.

eliotst: Also, if there are very common parameters that get passed in (not sure, since I'm not familiar with Redash), it might be a good idea to name them explicitly as optional function args.

parsons/redash/redash.py
```diff
@@ -0,0 +1,168 @@
...
+            params['api_key'] = query_api_key
+        response = self.session.get(f'{self.base_url}/api/queries/{query_id}/results.csv',
+                                    params=params,
+                                    verify=self.verify)
+        if response.status_code != 200:
+            raise Exception(f'Failed getting results for query {query_id}. {response.text}')
+        return response
+
+    @classmethod
+    def load_to_table(cls, refresh=True, **kwargs):
```
eliotst: I really like this pattern. We might have to look at incorporating this into some of our other Connectors.

parsons/redash/redash.py
```diff
@@ -0,0 +1,168 @@
...
+        with the CSV string in result.content
+
+        `Args:`
+            query_id: str or int
+                The query id of the query
+            query_api_key: str
+                If you did not supply a user_api_key on the Redash object, then you can
+                supply a query_api_key to get cached results back anonymously.
+        `Returns:`
+            Response Class
```
eliotst: Same question as above re: `Response` vs. `Table`.

schuyler1d:
Thanks for the review!  Ok:
* added `check()` support; I needed to add a optional=False arg to that utility method for it to work with optional Nones/0's.
* All methods return Table()s now

Re `params`, they are specific to the query, and it would be difficult with load_to_table. I would like to keep these separate so they don't cause trouble if their names overlap with current (or future) method arguments.  Just as an example, I've personally had redash queries with "base_url" as a parameter.

parsons/redash/redash.py
```diff
@@ -0,0 +1,186 @@
...
+                then this is a dict that will pass the parameters in the POST.
+                We add the "p_" prefix for parameters, so if your query had ?p_datelimit=....
+                in the url, you should just set 'datelimit' in params here.
+                If you set this with REDASH_QUERY_PARAMS environment variable instead of passing
+                the values, then you must include the "p_" prefixes and it should be a single
+                url-encoded string as you would see it in the URL bar.
+        `Returns:`
+            Table Class
+        """
+        query_id = check('REDASH_QUERY_ID', query_id, optional=True)
```
eliotst: This makes sense to do here, because this is sometimes called by the `load_to_table` classmethod below.. but for future reference, usually the environment variable pattern is more about `__init__` parameters than necessarily specific function parameters.

PR #161, opened 2020-02-20
Fixed odd Freshdesk contacts parameter

SorenSpicknall:
I noticed while testing Freshdesk-related code that the `updated_since` parameter is actually supposed to be `_updated_since` when being used with calls to their contacts endpoint, and the current parameter isn't recognized when passed to Freshdesk. I made a few quick substitutions to align the Parsons parameter with what's expected by the Freshdesk API for contacts.

### DIFF
--- a/parsons/freshdesk/freshdesk.py
+++ b/parsons/freshdesk/freshdesk.py
@@ -104,7 +104,7 @@ class Freshdesk():
         return self.transform_table(tbl, expand_custom_fields)
 
     def get_contacts(self, email=None, mobile=None, phone=None, company_id=None,
-                     state=None, _updated_since=None, expand_custom_fields=None):
+                     state=None, updated_since=None, expand_custom_fields=None):
         """
         Get contacts.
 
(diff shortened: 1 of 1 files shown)


parsons/freshdesk/freshdesk.py
```diff
@@ -104,7 +104,7 @@ def get_tickets(self, ticket_type=None, requester_id=None, requester_email=None,
         return self.transform_table(tbl, expand_custom_fields)
 
     def get_contacts(self, email=None, mobile=None, phone=None, company_id=None,
-                     state=None, updated_since=None, expand_custom_fields=None):
+                     state=None, _updated_since=None, expand_custom_fields=None):
```
eliotst: I would say keep the Parsons parameter as `updated_since`. The underscore is a peculiarity of the FreshDesk API that we can hide from the Parsons user in our parameter naming.

SorenSpicknall:
Agreed on that change. I've masked the parameter for cleanliness.

PR #162, opened 2020-02-21
Bill.com Parsons Class

ydamit:
All credit goes to my friend Kevin (@dekedor). All blame (including potentially bungling the rebase) belongs with me.

### DIFF
--- a/parsons/bill_com/bill_com.py
+++ b/parsons/bill_com/bill_com.py
@@ -1,0 +1,316 @@
+import requests
+import json
+
+
+class BillCom(object):
+    """
+    `Args:`
+        user_name: str
+            The Bill.com username
+        password: str
+            The Bill.com password
+        org_id: str
+            The Bill.com organization id
+        dev_key: str
+            The Bill.com dev key
+        api_url:
+            The Bill.com end point url
+    """
+
+    def __init__(self, user_name, password, org_id, dev_key, api_url):
+        self.headers = {
+            "Content-Type": "application/x-www-form-urlencoded"
+        }
+        params = {
+            "userName": user_name,
+            "password": password,
+            "orgId": org_id,
+            "devKey": dev_key
+        }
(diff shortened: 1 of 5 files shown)


parsons/bill_com/bill_com.py
```diff
@@ -0,0 +1,316 @@
...
+            **kwargs:
+                Any other fields to pass
+
+        `Returns:`
+            A list of dictionaries of user information for every user from start_user to max_user.
+        """
+        data = {
+           "start": start_user,
+           "max": max_user,
+           **locals()['kwargs']
```
eliotst: You should just be able to write `**kwargs`

parsons/bill_com/bill_com.py
```diff
@@ -0,0 +1,316 @@
...
+        `Args:`
+            start_user: int
+                The index of first user to return. Starts from 0 (not 1).
+            max_user: str
+                The index of the max user to return
+            **kwargs:
+                Any other fields to pass
+
+        `Returns:`
+            A list of dictionaries of user information for every user from start_user to max_user.
```
eliotst: The Parsons convention is to wrap the response in a `Table` object when returning multiple records.

eliotst: This also applies to the other `list_` methods below.

parsons/bill_com/bill_com.py
```diff
@@ -0,0 +1,316 @@
...
+        """
+        if "id" in customer1.keys():
+            if customer1["id"] == customer2["id"]:
+                return True
+        if "id" not in customer1.keys() and customer2['email']:
+            if customer1['email'].lower() == customer2['email'].lower():
+                return True
+        return False
+
+    def create_customer(self, customer_name, customer_email, **kwargs):
```
eliotst: I think it might make sense to call this `get_or_create_customer`.

PR #167, opened 2020-02-26
upgrade to facebook business 6.0.0

bxjw:
The Graph API was using v4.0, which does not work for new Facebook apps


eliotst:
Thanks for this! I just want to do a little manual testing on my end with the new Facebook Business library version to make sure things are backwards compatible with how Parsons uses the library. I will approve the PR if things look good in my testing.

bxjw:
Makes sense! Thanks for being on the ball.

PR #173, opened 2020-03-08
VAN: Add Custom Fields Endpoint

jburchard:
Adds the following methods:

`VAN.get_custom_field()` - Retrieve a single custom field object as a json.
`VAN.get_custom_fields()` - Retrieve all custom fields as a `Table` object.
`VAN.get_custom_fields_values()` - Retrieve a table of all custom fields values with their associated custom field id.

Also:

- Cleaned up some documentation for VAN.
- Fixed a bug for `VAN.get_code()`

### DIFF
--- a/docs/ngpvan.rst
+++ b/docs/ngpvan.rst
@@ -242,7 +242,7 @@ Codes
 =============
 Custom Fields
 =============
-.. autoclass:: parsons.ngpvan.van.Codes
+.. autoclass:: parsons.ngpvan.van.CustomFields
    :inherited-members:
 
 ======
--- a/docs/ngpvan.rst
+++ b/docs/ngpvan.rst
@@ -215,40 +215,16 @@ People
 .. autoclass:: parsons.ngpvan.van.People
    :inherited-members:
 
-===========
-Bulk Import
-===========
-.. autoclass:: parsons.ngpvan.van.BulkImport
-   :inherited-members:
-
 ==============
 Activist Codes
 ==============
 .. autoclass:: parsons.ngpvan.van.ActivistCodes
    :inherited-members:
 
-================
-Survey Questions
-================
-.. autoclass:: parsons.ngpvan.van.SurveyQuestions
-   :inherited-members:
-
-======
-Events
-======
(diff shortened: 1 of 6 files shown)


docs/ngpvan.rst
```diff
@@ -215,40 +215,16 @@ People
 .. autoclass:: parsons.ngpvan.van.People
    :inherited-members:
 
-===========
-Bulk Import
-===========
```
ydamit: :(
Eagerly awaiting its glorious return!

docs/ngpvan.rst
```diff
@@ -263,26 +239,28 @@ Codes
...
    :inherited-members:
 
-=================
-Canvass Responses
-=================
-.. autoclass:: parsons.ngpvan.van.CanvassResponses
+=============
+Custom Fields
+=============
+.. autoclass:: parsons.ngpvan.van.Codes
```
ydamit: Is it right that this is Codes?

parsons/ngpvan/custom_fields.py
```diff
@@ -0,0 +1,73 @@
...
+        self.connection = van_connection
+
+    def get_custom_fields(self, field_type='contacts'):
+        """
+        Get custom fields.
+
+        `Args:`
+            field_type : str
+                Filter by custom field group type. Must be one of ``contacts`` or
+                ``contributions``.
```
ydamit: In light of this binary distinction, and the unlikeliness that it will change any time soon, should we consider making this be a boolean along the line of `get_custom_fields(self, contrib_fields=False)` and then add a line of code to parse that above line 26 below?

docs/ngpvan.rst
jburchard: Good catch. Fixed.

parsons/ngpvan/custom_fields.py
jburchard: I considered that and I think that it is more straightforward to the user to keep it this way. Additionally, I included a default argument of `contacts` so presumably the user won't even have to enter in a value most of the time. And, as you mention, this will possibly future proof for additional custom field types that are included.

PR #174, opened 2020-03-08
Van people

ydamit:
This PR is an attempt to allow for updating VAN records using VANIDs. This also necessitated cleaning up the `_person_search()` and `_valid_search()` methods, and resolving the confusion of how street name/number map to actual address data.

This was done in a way that hopefully remains backwards-compatible. Leaving the prospect of VAN ID as a named arg in the `upsert_person()` for a subsequent PR (`find_person()` may not require this since there's `get_person()`).

Successfully tested in the following scenarios:
- find_person() using match_map with `firstName`, `lastName`, `addressLine1`, `zipOrPostalCode`
- find_person() using the same match_map but with a different record's email in the `email` arg (to prove that match_map overrides)
- find_person() with `first_name`, `last_name`, `street_number`, `street_name`, and `zip` args (to prove that using args without match_map still works)
- find_person() using match_map that only has `vanId`
- upsert_person() using match_map that only has `vanId` and `middleName`
- upsert_person() using the same match_map but with a different record's email in the `email` arg (to prove that match_map overrides)
- upsert_person() with `first_name`, `last_name`, `street_number`, `street_name`, `zip`, and `phone` args (to update phone)

### DIFF
--- a/parsons/ngpvan/people.py
+++ b/parsons/ngpvan/people.py
@@ -11,8 +11,7 @@ class People(object):
         self.connection = van_connection
 
     def find_person(self, first_name=None, last_name=None, date_of_birth=None, email=None,
-                    phone=None, phone_type=None, street_number=None, street_name=None, zip=None,
-                    match_map=None):
+                    phone=None, phone_type=None, street_number=None, street_name=None, zip=None):
         """
         Find a person record.
 
@@ -25,12 +24,6 @@ class People(object):
             - first_name, last_name, zip5, date_of_birth
             - first_name, last_name, street_number, street_name, zip5
             - email_address
-
-        .. note::
-            The arguments that can be passed are a selection of the possible values that
-            can be used in a search. A full list of possible values can be found
-            `here <https://developers.ngpvan.com/van-api#match-candidates>`_. To use these
-            values, pass in a dictionary using the match_map argument.
 
         `Args:`
             first_name: str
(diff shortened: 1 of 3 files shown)


jburchard:
This is really great and before I dig in, I'm wondering if we might want to consider a different pattern for this as the method has become really involved (my fault on that one). Perhaps we could consider breaking this into different methods:

`VAN.find_person()` - Only allows for you to search by the standard args

`VAN.find_person_advanced()` - Only allows you to pass in a `match_map` json. We should probably just give it a different name and include a link to the VAN documentation of the standard Person object.

`VAN.upsert_person()` - Only allows for you to search by the standard args

`VAN.upsert_person_advanced()` - Only allows you to pass in a `match_map` json. We should probably just give it a different name and include a link to the VAN documentation of the standard Person object.

For upsert with VANID, might that be better served by a method called `VAN.update_person()` which requires VANID? Perhaps also following the basic/advanced pattern.

I know that there is a lot in here, so eager for your feedback. It seems like all of the pieces are in the code, but it would just require a little bit of work to break them out in to component methods.

ydamit:
@jburchard , I'm not as sure about the method names I came up with, but I think this does everything discussed.

parsons/ngpvan/people.py
```diff
@@ -11,11 +11,59 @@ def __init__(self, van_connection):
...
+            phone: str
+                Phone number of any type (Work, Cell, Home)
+            street_number: str
+                Street Number
+            street_name: str
+                Street Name
+            zip: str
+                5 digit zip code
+        `Returns:`
+            A person dict
```
jburchard: Perhaps a `A person dict object`.

parsons/ngpvan/people.py
```diff
@@ -27,12 +75,36 @@ def find_person(self, first_name=None, last_name=None, date_of_birth=None, email
...
-            values, pass in a dictionary using the match_map argument.
+            A full list of possible values for the json, and its structure can be found
+            `here <https://developers.ngpvan.com/van-api#match-candidates>`_.
+
+        `Args:`
+            match_json: dict
+                A dictionary of values to match against.
+            fields: The fields to return. Leave as default for all available fields
+        `Returns:`
+            A person dict
```
jburchard: A person dict object.

parsons/ngpvan/people.py
```diff
@@ -27,12 +75,36 @@ def find_person(self, first_name=None, last_name=None, date_of_birth=None, email
...
+
+        logger.info(f'Finding a match for json details.')
+
+        return self._people_search(match_json=match_json)
+
+    def update_person(self, id=None, id_type='vanid', first_name=None, last_name=None,
+                      date_of_birth=None, email=None, phone=None, phone_type=None,
+                      street_number=None, street_name=None, zip=None):
+        """
+        Update a person record based on a provided ID.
```
jburchard: I would make it clear here that the arguments provided beyond `id` are going to updated the record. That's correct, right?

PR #176, opened 2020-03-09
MySQL Connector: Query

jburchard:
Initial MySQL connector with the following two methods:

`MySQL.query()`
`MySQL.query_with_connection()`


parsons/databases/mysql/mysql.py
```diff
@@ -0,0 +1,188 @@
...
+        port: int
+            Can be set by env variable ``MYSQL_PORT`` or argument.
+    """
+
+    def __init__(self, host=None, username=None, password=None, db=None, port=3306):
+
+        self.username = check_env.check('MYSQL_USERNAME', username)
+        self.password = check_env.check('MYSQL_PASSWORD', password)
+        self.host = check_env.check('MYSQL_HOST', host)
+        # self.db = None
```
eliotst: This looks like it can be cleaned up.

parsons/databases/mysql/mysql.py
```diff
@@ -0,0 +1,188 @@
...
+
+        `Returns:`
+            Parsons Table
+                See :ref:`parsons-table` for output options.
+        """
+        with self.cursor(connection) as cursor:
+
+            # The python connector can only execute a single sql statement, so we will
+            # break up each statement and execute them separately.
+            for s in sql.strip().split(';'):
```
eliotst: It looks like there's a `multi` keyword which can be passed into the `execute` call. [From what I can tell](https://overiq.com/mysql-connector-python-101/executing-queries-using-connector-python/#executing-multiple-queries-with-execute), it looks like the return values is a list of result sets.. not sure how that works with the `fetchmany` method below.

I definitely think we should lean on the connector for parsing out the different statements.. if I have a query that contains a `;` in a string (e.g, `select * from tmc_data.table where response = 'a;b;c'`), then `split` above will create 3 strings, none of which is the complete statement. Not a super likely scenario, but given enough users, someone will probably hit it.

jburchard:
I'd like to spend some with the multi and push an update in a separate PR.

PR #179, opened 2020-03-12
Adding Pagination to List Methods

ydamit:
By default, the `max` for list calls was set to 999, but it turns out there can be more than 999 of a thing. So I introduced a method to paginate results for list methods, and a test method for it as well.

### DIFF
--- a/parsons/bill_com/bill_com.py
+++ b/parsons/bill_com/bill_com.py
@@ -119,12 +119,12 @@ class BillCom(object):
         # Flag to simulate do-while
         first_run = True
 
-        while first_run or len(r) == max_ct:
+        while first_run or len(response) == max_ct:
             first_run = False
             data['start'] += max_ct
             data['max'] += max_ct
-            r = self.get_request_response(data, "List", object_name)
-            r_table.concat(Table(r))
+            response = self.get_request_response(data, "List", object_name)
+            r_table.concat(Table(response))
 
         return r_table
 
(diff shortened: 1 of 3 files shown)


parsons/bill_com/bill_com.py
```diff
@@ -101,6 +101,33 @@ def get_request_response(self, data, action, object_name, field="response_data")
         """
         return self.post_request(data, action, object_name)[field]
 
+    def paginate_list(self, response, data, object_name):
```
jburchard: I believe if you prepend the method with a `_` (e.g. `_paginate_list`) then it won't show up in the documentation. Otherwise, this internal method will. Alternatively, you can use `#` instead of `"""` and it won't show up in the documentation.

Looks like there are other methods in this class that have the same issue.

parsons/bill_com/bill_com.py
```diff
@@ -140,9 +170,12 @@ def get_customer_list(self, start_customer=1, max_customer=999, **kwargs):
            "max": max_customer,
            **kwargs
         }
-        return Table(self.get_request_response(data, "List", "Customer"))
 
-    def get_invoice_list(self, start_invoice=1, max_invoice=999, **kwargs):
+        r = self.get_request_response(data, "List", "Customer")
+
+        return self.paginate_list(r, data, "Customer")
```
jburchard: Not a huge deal, but curious why you aren't calling this within the `get_request_response` call instead of each individual method?

PR #180, opened 2020-03-13
Redshift: Alter table width on Redshift.copy()

jburchard:
This PR does the following:

- Improves the performance of `Redshift.copy()`. Previously, it was leveraging the `Redshift.copy_s3()` method, which - when creating a new table - was causing it to download the file from S3 an additional time to scan the data. Now it scans the data prior to uploading to S3.

- Add an `alter_table` argument to Redshift.copy() which will alter the width of a target table columns if they are too narrow for the incoming data. This should significantly reduce failures on copy. Unfortunately, if the table has an associated dependent view, this will fail. Hence, for now, the default will be `False`, but we will consider shifting that in the future.

### DIFF
--- a/parsons/databases/redshift/redshift.py
+++ b/parsons/databases/redshift/redshift.py
@@ -410,7 +410,7 @@ class Redshift(RedshiftCreateTable, RedshiftCopyTable, RedshiftTableUtilities, R
                 target table. This will also fail if the target table has an `IDENTITY`
                 column and that column name is among the source table's columns.
             alter_table: boolean
-                Will check if the target table varchar widths are wide enough to copy in the 
+                Will check if the target table varchar widths are wide enough to copy in the
                 table data. If not, will attempt to alter the table to make it wide enough. This
                 will not work with tables that have dependent views.
             aws_access_key_id:
--- a/parsons/databases/redshift/redshift.py
+++ b/parsons/databases/redshift/redshift.py
@@ -430,33 +430,28 @@ class Redshift(RedshiftCreateTable, RedshiftCopyTable, RedshiftTableUtilities, R
         else:
             cols = None
 
-        key = None
-
         with self.connection() as connection:
 
+            # Check to see if the table exists. If it does not or if_exists = drop, then
+            # create the new table.
+            if self._create_table_precheck(connection, table_name, if_exists):
+                sql = self.create_statement(tbl, table_name, padding=padding,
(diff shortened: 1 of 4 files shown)


parsons/databases/redshift/redshift.py
```diff
@@ -420,32 +424,66 @@ def copy(self, table_obj, table_name, if_exists='fail', max_errors=0, distkey=No
...
+                    logger.info(f'{table_name} created.')
+
+                # If alter_table is True, then alter table if the table column widths
+                # are wider than the existing table.
+                elif alter_table:
+                    self.alter_varchar_column_widths(tbl, table_name)
 
-            self.temp_s3_delete(key)
+                else:
+                    pass
```
eliotst: You don't need this `else: pass` here.. I feel like the code reads a little better without it.

parsons/databases/redshift/redshift.py
```diff
@@ -420,32 +424,66 @@ def copy(self, table_obj, table_name, if_exists='fail', max_errors=0, distkey=No
...
-                         acceptanydate=acceptanydate, dateformat=dateformat, timeformat=timeformat,
-                         blanksasnull=blanksasnull, nullas=nullas, emptyasnull=emptyasnull,
-                         acceptinvchars=acceptinvchars, truncatecolumns=truncatecolumns,
-                         columntypes=columntypes, specifycols=cols,
-                         aws_access_key_id=aws_access_key_id,
-                         aws_secret_access_key=aws_secret_access_key, compression='gzip')
+        with self.connection() as connection:
 
-        finally:
+            try:
```
eliotst: Minor thing, but the `try:` here is only really needed starting at line 460 (after the `temp_s3_copy` call), since the `finally:` is just about cleaning up the S3 object.. which won't exist until after the `temp_s3_copy` call.

parsons/databases/redshift/redshift.py
```diff
@@ -420,32 +424,66 @@ def copy(self, table_obj, table_name, if_exists='fail', max_errors=0, distkey=No
...
-        cols = None
+        # Specify the columns for a copy statement.
         if specifycols:
-            cols = table_obj.columns
+            cols = tbl.columns
+        else:
+            cols = None
 
-        try:
+        key = None
```
eliotst: Just noticed.. I don't think you would need this if you moved the `try:` down to line 460 (see comment below).

PR #181, opened 2020-03-13
Actionkit bulk upload

schuyler1d:
Support uploading batch uploads from a Table or csv to ActionKit (as documented at https://roboticdogs.actionkit.com/docs/manual/api/rest/uploads.html)


parsons/action_kit/action_kit.py
```diff
@@ -463,3 +465,83 @@ def create_generic_action(self, page, email=None, ak_id=None, **kwargs):
...
+                progress_url: an API URL to get progress on upload processing
+                res: requests http response object
+        """
+        import_page = check_env.check('ACTION_KIT_IMPORTPAGE', import_page)
+        csv_file = StringIO()
+        outcsv = csv.writer(csv_file)
+        for row in table.table:
+            outcsv.writerow(row)
+        return self.bulk_upload_csv(import_page, StringIO(csv_file.getvalue()),
+                                    autocreate_user_fields=autocreate_user_fields)
```
jburchard: This should work utilizing existing `Table` methods so you don't have to use the `csv` module.

```
return self.bulk_upload_csv(import_page, table.to_csv(), autocreate_user_fields=autocreate_user_fields)
```

parsons/action_kit/action_kit.py
```diff
@@ -463,3 +465,83 @@ def create_generic_action(self, page, email=None, ak_id=None, **kwargs):
...
+                  <https://roboticdogs.actionkit.com/docs/manual/api/rest/uploads.html#create-a-multipart-post-request>`
+        `Returns`:
+            dict
+                success: whether upload was successful
+                progress_url: an API URL to get progress on upload processing
+                res: requests http response object
+        """
+        import_page = check_env.check('ACTION_KIT_IMPORTPAGE', import_page)
+        if csv_file is None:
+            csv_file = check_env.check('ACTION_KIT_CSVFILE', csv_file)
```
jburchard: Is there a specific reason to include the argument as a environmental variable? That's just not something that we've done with other methods (outside of instantiating classes), so I'm curious as to why. Additionally, it's not documented.

parsons/action_kit/action_kit.py
```diff
@@ -463,3 +465,83 @@ def create_generic_action(self, page, email=None, ak_id=None, **kwargs):
...
+    def bulk_upload_csv(self, import_page=None, csv_file=None, autocreate_user_fields=0):
+        """
+        Bulk upload a csv file of new users or user updates.
+        If you are uploading a table object, use bulk_upload_table instead.
+        See `ActionKit User Upload Documentation
+             <https://roboticdogs.actionkit.com/docs/manual/api/rest/uploads.html>`
+        Be careful that blank values in columns will overwrite existing data.
+
+        If you get a 500 error, try sending a much smaller file (say, one row),
+        which is more likely to return the proper 400 with a useful error message
```
jburchard: I'd put this as a call out box for the documentation.
```
.. note::
        If you get a 500 error, try sending a much smaller file (say, one row),
        which is more likely to return the proper 400 with a useful error message
```

parsons/action_kit/action_kit.py
```diff
@@ -463,3 +465,83 @@ def create_generic_action(self, page, email=None, ak_id=None, **kwargs):
...
+            self._base_endpoint('upload'),
+            files={'upload': csv_file},
+            data={'page': import_page,
+                  'autocreate_user_fields': int(autocreate_user_fields)})
+        rv = {'res': res,
+              'success': res.status_code == 201,
+              'progress_url': res.headers.get('Location')}
+        return rv
+
+    def bulk_upload_table(self, import_page, table, autocreate_user_fields=0):
```
jburchard: An (undocumented) norm is that we typically include `table` as the first argument in a method that takes a Parsons table.

parsons/action_kit/action_kit.py
schuyler1d: My objection to this is that it uses the file-system, and I'd rather this be able to stay in-memory.

I'm happy to make a `to_csv_memory` in tofrom.py if you prefer that.

parsons/action_kit/action_kit.py
schuyler1d: I guess I over-interpreted feedback from the redash PR which requested env vars:
https://github.com/move-coop/parsons/pull/159#discussion_r382195027
That said, it seemed like it might be useful if someone wanted to have a generic script that just called upload_to_csv() and then provided by-environment variable when they had a different file.  However, I get that whereas the Redash use-case was loading data, this writes it, so maybe better not to.

I'll remove it

schuyler1d:
Responding to changes I:
* flipped the arguments
* removed the import_page/csv_filename env variable support
* made the 500 flag a 'note'

Some things I'd like to stick with:
* keeping the table => csv in-memory
* keep the bulk_upload_csv method -- I agree that we should push people to use Tables to pass data around, but we have a lot of use-cases where the csv already exists and bulk_upload_table depends on this code anyway.  I tried to 'nudge' users toward bulk_upload_table() in the doc comments of bulk_upload_csv.  I suppose we could make it a hidden method, but it seemed like it might be better just to have both in this particular case, since the API is literally uploading a csv.

PR #186, opened 2020-03-16
AWS-based framework to distribute table processing across Lambda invocations

schuyler1d:
The first call by divides up the table, uploading it to s3, and invoking lambdas to process portions of the data requesting specific data ranges in subsequent requests.

This is some code we used internally, and I'm adapting for within parsons.  I haven't tested it yet, but would like some early feedback on the approach:
* Leveraging the (undocumented) norm of putting `table` arguments first, this assumes there will be many methods that take a table as first argument and can process a table into actions/events/records in whatever system we are syncing to.
* I added support for class instantiation across Lambda invocations, since that is also a persistent pattern in Parsons -- however, it adds some complexity to the documented arguments and code.
* While the general pattern of class instantiation is good for services, I'm not quite convinced that the `bucket` argument should necessitate the same model -- especially when I can add check_env() for the bucket, it seems overkill to make a class, when Lambda will have all the other properties
* I was pretty shy about adding [Zappa](https://github.com/Miserlou/Zappa) as a dependency (do unto others...) -- while we may (or may not) make it a dependency ourselves (MoveOn), it seems better to keep it an optional dependency and to move over a few shims -- all the more so to add the class-supporting/instantiating arguments.

While this is pretty specific 'environment' code, I'd like to encourage use for those that are on AWS already.  It makes Lambda a more compelling (and friendly) platform for processing datasets -- even larger ones -- a lot cheaper than running an ec2 instance in most cases.


schuyler1d:
ok, we've deployed this and used it successfully so I think it's ready for review/feedback.

schuyler1d:
ok. feedback welcome on the rst file -- I figured I'd put it in a generic AWS area, where we can also group links to S3 and Redshift, in case someone is looking there instead of directly in the list.

PR #187, opened 2020-03-18
MySQL: Create Table

jburchard:
Adds the `MySQL.copy()` method similar to `Redshift` and `Postgres` in which the table is automatically created in the database if it does not already exist.

Of note, the copy method utilizes the MySQL multi-insert functionality rather than `LOAD DATA`. While this introduces a slight performance hit (less so than other DBs), there are security concerns with `LOAD DATA LOCAL`. Additionally, many configurations of MySQL DBs don't allow `LOAD DATA` by default.


parsons/databases/mysql/create_table.py
```diff
@@ -0,0 +1,159 @@
...
+        # current type is.
+
+        # Stop if the current type is already a varchar
+        if current_type == 'varchar':
+            return current_type
+
+        # Convert to string to reevaluate data type
+        try:
+            t = ast.literal_eval(str(val))
+        except ValueError:
```
SorenSpicknall: Can you outline the conditions for these two errors? Is one covering the case where the incoming value is already a string, and the other covering the case where that string is empty?

parsons/databases/mysql/create_table.py
```diff
@@ -0,0 +1,159 @@
...
+        if type(t) == float:
+            return 'float'
+
+        # If an int and current type is not a float, then determine type of int.
+        if type(t) == int and current_type != 'float':
+            # Make sure that it is a valid integer
+            if not self.is_valid_integer(val):
+                return 'varchar'
+            # Use smallest possible int type above TINYINT
+            if (-32768 < t < 32767) and current_type not in ['int', 'bigint', 'mediumint']:
```
SorenSpicknall: Is there a system function call we could draw these boundary values from, rather than hardcoding them?

parsons/databases/mysql/create_table.py
```diff
@@ -0,0 +1,159 @@
...
+                return 'smallint'
+            elif (-8388608 < t < 8388607) and current_type not in ['int', 'bigint']:
```
SorenSpicknall: Is there a system function call we could draw these boundary values from, rather than hardcoding them?

parsons/databases/mysql/create_table.py
```diff
@@ -0,0 +1,159 @@
...
+                return 'mediumint'
+            elif (-2147483648 < t < 2147483647) and current_type not in ['bigint']:
```
SorenSpicknall: Is there a system function call we could draw these boundary values from, rather than hardcoding them?

parsons/databases/mysql/create_table.py
```diff
@@ -0,0 +1,159 @@
...
+
+            # Calculate width if a varchar
+            if col_type == 'varchar':
+                row_width = len(str(row.encode('utf-8')))
+
+                # Evaluate width vs. current max width
+                if row_width > col_width:
+                    col_width = row_width
+
+        return col_type, col_width
```
SorenSpicknall: Maybe we should default to making the column width wider than the existing maximum for rows already in place?

parsons/databases/mysql/create_table.py
```diff
@@ -0,0 +1,159 @@
...
+        for idx, col in enumerate(columns):
+
+            # Lowercase
+            col = col.lower()
+
+            # Strip whitespace
+            col = col.strip()
+
+            # Replace spaces with underscores
+            col = col.replace(' ', '_')
```
SorenSpicknall: Are there other special characters that need to be stripped out or replaced?

jburchard: These were lifted from the `Redshift.copy()` and it has served as well to date. But, as corner cases arise, we can just slot them in.

jburchard:
@SorenSpicknall thanks for the feedback.

- Added config to the top of `create_table.py` that lifts the integer values out of the underlying code. I also included a link to the MySQL documentation around `INT` column type.

- Added a config at the top to set padding of varchar columns. The current value is `.25` which creates a 25% pad to the width. Does that seem like a good starting point?

Ready for rereview.

PR #188, opened 2020-03-20
actionkit bulk_user_upload with user_fields_only optimization

schuyler1d:
This preserves an optimization which ActionKit has for user_field-only uploads.

From @jburchard 's flag of max-ing out memory, I switch away from using the optimization if the record length exceeds a million.  Happy to use another arbitrary number to balance parson's users -- maybe something worth documenting and/or deciding in-terms of when/what to keep in-memory.

Also, I'm dribbling out changes, as I start to use my own e.g. past PR -- lmk, if it's better to 'fully bake' things.  This is a pretty low-priority change, so no rush on review.

### DIFF
--- a/parsons/action_kit/action_kit.py
+++ b/parsons/action_kit/action_kit.py
@@ -464,7 +464,7 @@ class ActionKit(object):
                                return_full_json=True,
                                **kwargs)
 
-    def bulk_upload_csv(self, csv_file, import_page, autocreate_user_fields=0):
+    def bulk_upload_csv(self, csv_file, import_page, autocreate_user_fields=0, user_fields_only=0):
         """
         Bulk upload a csv file of new users or user updates.
         If you are uploading a table object, use bulk_upload_table instead.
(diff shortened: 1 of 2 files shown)


parsons/action_kit/action_kit.py
```diff
@@ -536,6 +541,16 @@ def bulk_upload_table(self, table, import_page, autocreate_user_fields=0):
                 res: requests http response object
         """
         import_page = check_env.check('ACTION_KIT_IMPORTPAGE', import_page)
-        return self.bulk_upload_csv(table.to_csv(temp_file_compression='gzip'),
+        compress_maybe = dict()
+        user_fields_only = int(not any([
+            h for h in table.table.header()
```
eliotst: Could you add a line in the docstring above indicating that tables with only the user fields will be fast processed? Just to call people's attention to it.

parsons/action_kit/action_kit.py
```diff
@@ -464,7 +464,7 @@ def create_generic_action(self, page, email=None, ak_id=None, **kwargs):
                                return_full_json=True,
                                **kwargs)
 
-    def bulk_upload_csv(self, csv_file, import_page, autocreate_user_fields=0):
+    def bulk_upload_csv(self, csv_file, import_page, autocreate_user_fields=0, user_fields_only=0):
```
jburchard: Could you indulge me and change to `autocreate_user_fields=False` and `user_fields_only=False` to keep with the style of the rest of the library?

PR #189, opened 2020-03-23
Handle nonexistent target table during upsert

SorenSpicknall:
Added functionality to catch a failed upsert due to a missing target table, roll back that query, and transfer the entirety of the staging table to a newly-created target table.


SorenSpicknall:
> Could you add a quick unit test to test this functionality?

Just pushed a change within `test_redshift` to cover the new table creation behavior expected when upsert comes across a missing table.

PR #197, opened 2020-03-31
Add New/Mode integration

mirabuck:
Adding a new integration with New/Mode's API.


eliotst:
Thank you so much for this! I really appreciate the contribution. I will work to do a full review this over the next day or so, but I do want to flag that one of the automated checks failed. It looks like it is due to a missing `,` on line 75 of the `parsons/__init__.py` change.

kporras07:
Thanks! I just fixed it

parsons/newmode/newmode.py
```diff
@@ -0,0 +1,182 @@
...
+        self.api_password = check_env.check('NEWMODE_API_PASSWORD', api_password)
+
+        if (api_version is None):
+            api_version = "v1.0"
+
+        self.api_version = check_env.check('NEWMODE_API_VERSION', api_version)
+
+        self.client = Client(api_user, api_password, api_version)
+
+    def convertToTable(self, data):
```
eliotst, line 37: Python PEP8 convention for naming variables, functions, and methods is to use `lower_case_with_underscores` over `camelCase`. Please rename the methods / functions in the class to match.

parsons/newmode/newmode.py
```diff
@@ -0,0 +1,182 @@
...
+        # Internal method to create a Parsons table from a data element.
+        table = None
+        if (type(data) is list):
+            table = Table(data)
+        else:
+            table = Table([data])
+
+        return table
+
+    def getTools(self, params={}):
```
eliotst, line 47: In the name of ensuring we have complete documentation, we like to make sure all of our Connectors' "public" methods have doc-strings that describe the method API. For an example, you can see the `VAN` connector's `find_person` method: https://github.com/move-coop/parsons/blob/73adc2504a7ca2d3edd91378676ede56e82324a6/parsons/ngpvan/people.py#L15-L47

parsons/newmode/newmode.py
```diff
@@ -0,0 +1,182 @@
...
+            api_version = "v1.0"
+
+        self.api_version = check_env.check('NEWMODE_API_VERSION', api_version)
+
+        self.client = Client(api_user, api_password, api_version)
+
+    def convertToTable(self, data):
+        # Internal method to create a Parsons table from a data element.
+        table = None
+        if (type(data) is list):
```
eliotst, line 40: It's a minor point, but in Python parenthesis are unnecessary in simple conditional checks like this.. you could just write this as `if type(data) is list:`

parsons/newmode/newmode.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def getTool(self, tool_id, params={}):
+        tool = self.client.getTool(tool_id, params=params)
+        if (tool):
+            return self.convertToTable(tool)
+        else:
+            logging.warning("Empty tool returned")
+            return []
+
+    """
```
eliotst, line 63: This docstring needs to immediately follow the `def` statement below to be properly associated in the documentation with the method.

eliotst, line 63: This docstring needs to immediately follow the `def` statement below to be properly associated in the documentation with the method.

kporras07:
@eliotst thanks for your comments. I just added tests and docs. Please let me know if there's any other comment.

parsons/newmode/newmode.py
```diff
@@ -0,0 +1,305 @@
...
+            The id of the tool to return.
+        params:
+            Extra parameters sent to New/Mode library.
+    Returns:
+        Tools information as table.
+    """
+    def get_tool(self, tool_id, params={}):
+        tool = self.client.getTool(tool_id, params=params)
+        if tool:
+            return self.convert_to_table(tool)
```
eliotst, line 75: The convention in Parsons is for single record methods like this to return the dictionary instead of a `Table`. This applies to all single record methods in the Connector (e.g. `get_tool`, `get_campaign`, etc)

parsons/newmode/newmode.py
```diff
@@ -0,0 +1,305 @@
...
+        else:
+            logging.warning("Empty tool returned")
+            return []
```
eliotst, line 78: Likewise, for single record methods, when there is nothing to return, Parsons methods should return `None`.

parsons/newmode/newmode.py
```diff
@@ -0,0 +1,305 @@
...
+        # Internal method to create a Parsons table from a data element.
+        table = None
+        if type(data) is list:
+            table = Table(data)
+        else:
+            table = Table([data])
+
+        return table
+
+    """
```
eliotst, line 46: You will need to move the `"""` docstrings to after the method signature, and indented. Otherwise, the documentation library won't be able to see it to generate the API documentation.

So:
```
    def get_tools(self, params={}):
        """
        Get existing tools.
        Args:
            params:
                Extra parameters sent to New/Mode library.
            Returns:
                Tools information as table.
        """
```

You'll need to do this for all of the public methods in the class.

kporras07:
Thanks @eliotst 
New feedback addressed.

PR #198, opened 2020-03-31
NGPVAN: Fix Create Code

jburchard:
Fixes a bug with VAN.create_code()


parsons/ngpvan/codes.py
```diff
@@ -106,17 +106,19 @@ def create_code(self, name=None, parent_code_id=None, description=None,
                     ]
         """
 
+        json = {"parentCodeId": parent_code_id,
+                "name": name,
+                "codeType": code_type,
+                "supportedEntities": se,
```
SorenSpicknall: Since se isn't defined up here, either, shouldn't this take the form of:

`"supportedEntities": supported_entities`

PR #205, opened 2020-04-06
Improve Redshift upsert staging table column types

SorenSpicknall:
When a destination table already exists, this added code ensures that the staging table(s) created during a Redshfit upsert have columns that mirror those of the destination table. This helps avoid issues with copying data from table to table further on down the line.

### DIFF
--- a/parsons/databases/redshift/redshift.py
+++ b/parsons/databases/redshift/redshift.py
@@ -691,6 +691,10 @@ class Redshift(RedshiftCreateTable, RedshiftCopyTable, RedshiftTableUtilities, R
 
                 # Copy to a staging table
                 logger.info(f'Building staging table: {staging_tbl}')
+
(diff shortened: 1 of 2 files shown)


jburchard:
LGTM.

jburchard:
Might have some conflicts with @schuyler1d PR.

SorenSpicknall:
Yeah, I didn't realize that was merged in. These changes may not be necessary, since the behavior was handled differently in that PR. I'll likely close this PR once I've tested to ensure the case I was solving is solved by @schuyler1d's changes as well.

schuyler1d:
fwiw, your CREATE TABLE command looks more idiomatic/better than the way I wrote it here:
https://github.com/move-coop/parsons/blob/master/parsons/databases/redshift/redshift.py#L470

maybe replace my line with yours (but use `template_table`) and then we will also have more tests.

SorenSpicknall:
Closing this in order to make changes on another branch.

PR #206, opened 2020-04-06
Improve template table schema use, add testing for Redshift upsert

SorenSpicknall:
Working off of discussion in #205, I changed the query used to generate a matching DDL for a specific template table in the Redshift copy function. Then, I added some testing to cover a case where this functionality is used: when an incoming Parsons table has different datatypes than the destination table.


jburchard:
@schuyler1d - Any thoughts on this?

Issue #202, opened 2020-04-03
Hustle - return lead id for create_lead and create_leads


mkwoods927:
Currently the Hustle commands `create_lead` and `create_leads` return `None`. This ticket is a request to return the ids of the leads that are created with either one.

PR #208, opened 2020-04-08
Hustle: Return Lead Ids

jburchard:
Enhancement to `Hustle.create_leads()` method that returns a table of newly created lead_ids when creating new leads in the system.


jburchard:
Addresses #202

PR #209, opened 2020-04-08
action_kit.bulk_upload_table: create an option not to overwrite on empty values

schuyler1d:
This splits up a table into several tables based on empties.

Changes the API slightly -- return value is a dict with success:  but then an array of "results".  The API is sufficiently new, that I think this should be ok, but we can change it conditionally based on the option, if we want to preserve full backwards compatibility.

bikesheds on what to name the parameter welcome :-)

### DIFF
--- a/parsons/action_kit/action_kit.py
+++ b/parsons/action_kit/action_kit.py
@@ -1,6 +1,8 @@
 import json
 import logging
 import requests
+
+from parsons.etl.table import Table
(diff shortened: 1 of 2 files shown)


schuyler1d:
I can put it back -- it just seemed added complexity without enough gain.  LMK

jburchard:
No need. Just curious.

PR #223, opened 2020-04-20
Redshift.upsert: default to compupdate=False because compression isn't useful for temp tables

schuyler1d:
Our Redshift CPU was starting to get quite stressed when we deployed something with parsons' redshift.upsert.  After investigation, it seemed that the default of compupdate=True in the `copy` command was causing very slow analysis (even with several thousand rows) -- but especially with tables with a lot of *columns* (ours had 109).

It seems that especially for temporary tables, like in upsert, leaving compupdate=True is unlikely to be helpful, since the table will be deleted soon.

I'm shy about suggesting that the default in `.copy()` change, but it seems that there are a lot of 'upload a temporary-ish table' that `.copy()` is/might be used for where compupdate=False is the better option/default.  Maybe, if nothing else, it would be worth documenting.

cc: @ydamit (after discussion)


jburchard:
Beautiful.

PR #225, opened 2020-04-24
add the Rock the Vote connector

### LINKED ISSUE #119: Create Rock The Vote Class
We have code in the TMC private repo, and can probably easily create a class.

h/t to Gerard at ACRONYM who I originally stole said code from.

eliotst:
This commit adds the `RockTheVote` connector that wraps around the
Rock the Vote's Rocky API. This commit only implements endpoints
for requesting registrant reports, which is a report to list out
registration records for a given partner.

---
Fixes #119


docs/rockthevote.rst
```diff
@@ -0,0 +1,11 @@
+Rock the Vote
+=============
+
+`Rock the Vote <https://www.rockthevote.org/>`_ is an online registration tool.
+
+The Parsons Connector makes use of Rock the Vote's Rocky API. In order to authenticate with the
+API, users will need to specify the ID and the API key of the RTV partner organization for the
+data.
+
```
jburchard, line 9: Perhaps a little quick start? Generate report + get report?

parsons/rockthevote/rtv.py
```diff
@@ -0,0 +1,211 @@
...
+    'user-agent': 'curl/7.54.0'
+}
+"""Standard request header for sending requests to the API."""
+
+STATUS_URL_PARSE_REGEX = re.compile(r'(\d+)$')
+"""Regex for parsing the report ID from the status URL."""
+
+
+class RTVFailure(Exception):
+    """Exception raised when there is an error with the connector."""
```
jburchard, line 32: Noticed this pattern in Sky's code too. Would it make sense to move towards this as a standard?

parsons/rockthevote/rtv.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def create_registration_report(self, before=None, since=None):
+        """
+        Create a new registration report.
+
+        `Args:`
+            before: datetime.datetime
+                Date before which to return registrations for
+            since: datetime.datetime
+                Date for filtering registrations
```
jburchard, line 60: Mostly just acknowledging that passing in a `datetime` object is a new pattern for us. How do we feel about it?

jburchard: Also, if you pass both parameters is the report time bound within those two dates?

parsons/rockthevote/rtv.py
```diff
@@ -0,0 +1,211 @@
...
+            raise RTVFailure('Timed out waiting for report')
+
+        # Download the report data
+        download_response = self.client.request(download_url, 'get', params=credentials)
+
+        # Check to make sure the call got a valid response
+        if download_response.status_code == requests.codes.ok:
+            report_data = download_response.text
+            with open('/tmp/rtv.csv', 'w') as fh:
+                fh.write(report_data)
```
jburchard, line 167: How is this being used?

parsons/rockthevote/rtv.py
```diff
@@ -0,0 +1,211 @@
...
+            # Pythonic (snake case)
+            normalized_column_names = [
+                re.sub(r'\s', '_', name).lower()
+                for name in table.columns
+            ]
+            normalized_column_names = [
+                re.sub(r'[^A-Za-z\d_]', '', name)
+                for name in normalized_column_names
+            ]
+            table.table = petl.setheader(table.table, normalized_column_names)
```
jburchard, line 182: Would love this as a `Table` class method. I feel like we could use this in many many places.

parsons/rockthevote/rtv.py
eliotst: Yeah. I think it's a good idea to raise custom exceptions. An outstanding question is how we handle inheritance: do we have some common `ParsonsException` or do we have our custom exceptions inherit from corollary standard Python exceptions (e.g. `ValueError`). @schuyler1d do you have thoughts?

parsons/rockthevote/rtv.py
eliotst: I am changing the code to accept a string.. and then I'm going to pull in a `dateutil` library to help [parse](https://dateutil.readthedocs.io/en/stable/parser.html) the strings into a `datetime`.

eliotst: In my testing, it looks like specifying both results in registrations from between the two dates, yes.

parsons/rockthevote/rtv.py
eliotst: Good catch. It was for local testing.

parsons/rockthevote/rtv.py
eliotst: Done.

PR #232, opened 2020-05-04
bigquery: DB sync

eliotst:
This commit updates the Parsons BigQuery connector to support DB
syncs to/from BigQuery. In order to make this change, the commit
also modifies the `query` method in the connector to use the
BigQuery
[DB-API](https://googleapis.dev/python/bigquery/latest/dbapi.html)
interface.

The commit also updates the BigQuery connector API to
require users to qualify their table names with the name of the
dataset, so align the BigQuery API with the more standard SQL
connectors (Redshift, MySQL).

This commit also makes a few minor changes to the DB sync class
to fix some small bugs, namely:
 * Add the `table` method to MySQL to allow using MySQL for with
 `DBSync`
 * Ensure that the connection and cursor in the Redshift connector
 are closed, even if the query raises an exception.
 * Fixes a bug where the DB sync API wouldn't support incremental
 table syncs where the primary key wasn't an integer.
 * Work around a limitation in BigQuery where the `autodetect`
 of a copy job would confuse headers for a data row in the case
 where the table data was all strings.
 * Fix DB Sync `table_sync_incremental` so it works even if the
 destination table is empty.

### DIFF
--- a/parsons/databases/db_sync.py
+++ b/parsons/databases/db_sync.py
@@ -133,35 +133,41 @@ class DBSync:
         dest_max_pk = destination_tbl.max_primary_key(primary_key)
 
         # Check for a mismatch in row counts.
-        if dest_max_pk > source_max_pk:
+        if dest_max_pk and dest_max_pk > source_max_pk:
             raise ValueError('Destination DB primary key greater than source DB primary key.')
 
         # Do not copied if row counts are equal.
         elif dest_max_pk == source_max_pk:
             logger.info('Tables are in sync.')
             return None
-            # Question: Should I check that the row counts match too?
 
         else:
             # Get count of rows to be copied.
-            new_row_count = source_tbl.get_new_rows_count(primary_key, dest_max_pk)
+            if dest_max_pk is not None:
+                new_row_count = source_tbl.get_new_rows_count(primary_key, dest_max_pk)
+            else:
+                new_row_count = source_tbl.num_rows
+
             logger.info(f'Found {new_row_count} new rows in source table.')
 
-            # Copy rows in chunks.
             copied_rows = 0
-            while copied_rows < new_row_count:
-
+            # Copy rows in chunks.
+            while True:
                 # Get a chunk
                 rows = source_tbl.get_new_rows(primary_key=primary_key,
-                                               start_value=dest_max_pk,
+                                               cutoff_value=dest_max_pk,
                                                offset=copied_rows,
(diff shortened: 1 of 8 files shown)


parsons/databases/db_sync.py
```diff
@@ -133,35 +133,41 @@ def table_sync_incremental(self, source_table, destination_table, primary_key,
         dest_max_pk = destination_tbl.max_primary_key(primary_key)
 
         # Check for a mismatch in row counts.
-        if dest_max_pk > source_max_pk:
+        if dest_max_pk and dest_max_pk > source_max_pk:
```
jburchard, line 136: Looks like you are checking `dest_max_pk` twice.

parsons/google/google_bigquery.py
```diff
@@ -1,13 +1,58 @@
...
+        'dataset': None,
+        'table': None,
+    }
+    if len(parts) > 0:
+        parsed['table'] = parts[0]
+    if len(parts) > 1:
+        parsed['dataset'] = parts[1]
+    if len(parts) > 3:
+        parsed['project'] = parts[2]
+    return parsed
```
jburchard, line 55: Oh my. This is a lot. Thank you!

parsons/google/google_bigquery.py
```diff
@@ -50,18 +95,18 @@ def __init__(self, app_creds=None, project=None, location=None):
         # This attribute will be used to hold the client once we have created it.
         self._client = None
 
+        self._dbapi = dbapi
+
         self.dialect = 'bigquery'
 
-    def copy(self, table_obj, dataset_name, table_name, if_exists='fail',
+    def copy(self, table_obj, table_name, if_exists='fail',
```
jburchard, line 102: Any breaking changes concerns?

eliotst: Yes. This will break anyone who has been using BigQuery. I will definitely flag in the release notes. I couldn't think of a great way to solve the problem where the DB Sync expects the copy args to be `table_obj`, `table_name` (in that order).

I am happy to revisit if you think it is going to be too big of an impact.

parsons/databases/db_sync.py
eliotst: So, I was getting an error if the `dest_max_pk` was `None`: `TypeError: '>' not supported between instances of 'NoneType' and 'int'`

I actually think I need to change this though to `if dest_max_pk is not null and dest_max_pk > source_max_pk:`

PR #235, opened 2020-05-05
Action network

dekedor:


parsons/action_network/action_network.py
```diff
@@ -0,0 +1,233 @@
...
+                "email_addresses": email_addresses_field,
+                "given_name": given_name,
+                "family_name": family_name,
+                "languages_spoken": languages_spoken,
+                "postal_addresses": postal_addresses,
+                "custom_fields": {**kwargs}
+              },
+            "add_tags": tags
+        }
+        self._add_entry("people", data, silent)
```
eliotst, line 154: Minor thing, but my preference is to get rid of the `_add_entry`, `_update_entry`, and `_get_entry` methods. I think the code is a little more readable if you can see inline what `_add_entry` is doing, and if you get rid of the `silent` argument for `_add_entry` and `_update_entry` they end up being a single line each, which is easy enough to just run here within this method.

`_get_entry_list` is doing a lot more, so it makes sense to have it be a separate method.

parsons/action_network/action_network.py
```diff
@@ -43,58 +46,33 @@ def _get_entry_list(self, object_name, limit=10, timeout=60):
...
-        # updates fields for a given entry
-        # only the fields to be updated need to be included
-        response = self.api.put_request(url=f"{self.api_url}/{object_name}/{entry_id}",
-                                        json=json.dumps(data), success_codes=[204, 201, 200])
-        if not silent:
-            logger.info(f"{object_name.capitalize()} entry {entry_id} successfully updated")
-        return response
 
-    def get_people_list(self, limit=10, timeout=60):
+    def get_people(self, limit=None, per_page=25, page=None):
```
eliotst, line 52: Could you add documentation in the `Args` below for the two new arguments? Same for `get_tags`.

dekedor: Wow, that was careless on my part. Thanks for catching that.

eliotst:
Thank you again for this!

Issue #238, opened 2020-05-06
GoogleSheets() class ability to share sheet


retacg:
The GoogleSheets() parsons class doesn't currently provide the capability to just share a sheet with an individual nor make the sheet readable, or editable by anyone. Looks like gspread has the capability to do so via `share` -- more info here: https://gspread.readthedocs.io/en/latest/api.html#gspread.models.Spreadsheet.share and it would be great to just have the capacity to do that via parsons too. 

I am most interested in being able to quickly make a sheet viewable by anyone with the link after it's created.

PR #240, opened 2020-05-07
GoogleSheets: Share A Spreadsheet

jburchard:
Adds the following methods:

- ``get_spreadsheet_permissions()``
- ``share_spreadsheet()``

Addresses Issue #238

### DIFF
--- a/parsons/google/google_sheets.py
+++ b/parsons/google/google_sheets.py
@@ -108,6 +108,50 @@ class GoogleSheets:
         index = self.get_sheet_index_with_title(spreadsheet_id, title)
         return self.read_sheet(spreadsheet_id, index)
 
+    def share_spreadsheet(self, spreadsheet_id, sharee, share_type='user', role='reader',
+                    notify=True, notify_message=None, with_link=False):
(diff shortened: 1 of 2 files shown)


parsons/google/google_sheets.py
```diff
@@ -108,6 +108,50 @@ def read_sheet_with_title(self, spreadsheet_id, title):
         index = self.get_sheet_index_with_title(spreadsheet_id, title)
         return self.read_sheet(spreadsheet_id, index)
 
+    def share_spreadsheet(self, spreadsheet_id, sharee, share_type='user', role='reader',
+                          notify=True, notify_message=None, with_link=False):
+        """
+        Share a spreadsheet with a user.
```
retacg, line 114: for clarity, i would edit this to say "Share a spreadsheet with a user and/or the public." I had wished gspread was more clear in their description of "share()" to explicitly say that you can change the overall permissions of the sheet. Open to your thoughts though!

jburchard: Great suggestion!

PR #247, opened 2020-05-18
Add PDI class

dannyboy15:
The PDI class implements a connector to [PDI's API](https://api.bluevote.com/docs/index#). The class implements the following endpoints:
- FlagIDs
- Flags
- Universes
- Questions
- AcquisitionTypes 

~Additionally, it appears that PDI has deprecated a few of them.~ PDI has deprecated some endpoints, none of which are implemented by this class.

The local tests have passed. Live tests were tested about a month ago and passed (I no longer have an api key so I can't test them again).


jburchard:
Would it possible to indicate which ones are deprecated?

dannyboy15:
> Would it possible to indicate which ones are deprecated?

I had worded it in a confusing manner, sorry. I clarified that none of the endpoints implemented in the PDI class are deprecated.

SorenSpicknall:
I filled out some additional content in the __init__.py for the PDI connector, and added some additional details to the RST documentation. I'll give it another once-over before approval, but it generally looks good to me!

PR #249, opened 2020-05-18
Slack: Handle rate limiting and binary files

dannyboy15:
* The connector will now automatically retry if it is rate limited. It uses the `Retry-After` header to determine how long to wait.

* The connector can now upload/post files that are binary (e.g. `pdf`, `jpg`, etc). To do so, pass `is_binary=True`. 

```python
...
slack.upload_file(['some_channel'], 'myfile,pdf', is_binary=True):
```

### DIFF
--- a/parsons/notifications/slack.py
+++ b/parsons/notifications/slack.py
@@ -1,7 +1,8 @@
-from parsons.etl import Table
+from parsons import Table
 from slackclient import SlackClient
 from slackclient.exceptions import SlackClientError
 import os
+import time
 
 
 class Slack(object):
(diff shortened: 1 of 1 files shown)


parsons/notifications/slack.py
```diff
@@ -173,3 +200,12 @@ def _paginate_request(self, endpoint, collection, **kwargs):
                 next_page = False
 
         return Table(items)
+
+    def _check_response(self, resp):
```
eliotst, line 204: This method does not seem to be used. It looks like it could be cleaned up.

PR #255, opened 2020-05-20
Documentation Housekeeping

jburchard:
Did a bit of work trying to consolidate some of the tools (AWS, Google) and just some minor issues that I saw. No live code was touched.


docs/newmode.rst
```diff
@@ -1,7 +1,7 @@
 New/Mode
 ==========
 
-`New/Mode <https://www.newmode.net/>` is the multi-channel advocacy & civic engagement platform
+`New/Mode <https://www.newmode.net/>`_ is the multi-channel advocacy & civic engagement platform
```
SorenSpicknall, line 4: Do you think it'd be worth modifying the copy here to be less like marketing speak? Most of the other connectors have very dry, matter-of-fact descriptions of the products they're connecting to, whereas this one seems to be a copy/pasted summary from the vendor.

jburchard: Fixed.

PR #263, opened 2020-05-24
GoogleSheets Refactor

jburchard:
This PR does some significant refactoring to the GoogleSheets class and reflects new patterns developed since that time. 

- Previously, there were separate methods to read worksheets, depending on whether you passed the worksheet index or the title of the worksheet. The methods have now been combined and it determines whether it is a title of an index based on the value passed.

- The old methods have been removed from the documentation, but still exist and throw a deprecation warning in the log.

- Adds ``**kwargs`` to a few methods to make them backward compatible.

- Renames some methods that referred to worksheets as `sheets`, which could easily be confused with `spreadsheets`. 

- Adds a method `list_worksheets()` that returns a list of worksheets in a given spreadsheet.

### DIFF
--- a/parsons/google/google_sheets.py
+++ b/parsons/google/google_sheets.py
@@ -55,7 +55,10 @@ class GoogleSheets:
 
         elif isinstance(worksheet, str):
             idx = self.list_worksheets(spreadsheet_id).index(worksheet)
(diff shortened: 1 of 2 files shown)


jburchard:
Ready for final review.

PR #265, opened 2020-05-27
phone2action: handle empty advocates results

eliotst:
This commit updates the `Phone2Action` connector to handle an
empty list from the call to the advocates endpoint on the P2A API.
Before the commit, the call to get advocates was failing when
trying to unpack the advocates data.

This commit also adds the `__bool__` magic method to the Parsons
`Table` class to support simple bool checks on the emptiness of
a table (e.g. `if not tbl:` instead of `if tbl.num_rows > 0`)

### DIFF
--- a/test/test_p2a.py
+++ b/test/test_p2a.py
@@ -175,7 +175,7 @@ class TestP2A(unittest.TestCase):
         self.assertTrue(validate_list(fields_exp, self.p2a.get_advocates()['fields']))
 
     @requests_mock.Mocker()
-    def test_get_advocates(self, m):
+    def test_get_advocates__by_page(self, m):
 
         response = copy.deepcopy(adv_json)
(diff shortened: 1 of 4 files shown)


test/test_p2a.py
```diff
@@ -175,7 +175,7 @@ def test_get_advocates(self, m):
         self.assertTrue(validate_list(fields_exp, self.p2a.get_advocates()['fields']))
 
     @requests_mock.Mocker()
-    def test_get_advocates(self, m):
+    def test_get_advocates__by_page(self, m):
```
SorenSpicknall, line 178: Is this an unintentional extra underscore?

eliotst: I'm trying to differentiate between "test `get_advocates_by_page`" (which isn't a method) and "test `get_advocates` by page".

PR #274, opened 2020-06-11
Cover new column expansion case in Redshift testing

SorenSpicknall:
In #273, we implemented automatic resizing of varchar columns when using the Redshift upsert function. This PR adds testing to cover that new case.


eliotst:
Were you able to run these yourself?

SorenSpicknall:
> Were you able to run these yourself?

They run correctly on CircleCI, but I haven't run testing locally. Is there a different level of testing or configuration for testing that I should use, besides the configuration used by CircleCI?

eliotst:
In order to run the live tests, you will need to run `export LIVE_TESTS=1` on the command line. Then, run `pytest test/test_redshift.py`.. that will run the live tests and the offline tests.

Issue #73, opened 2019-10-31
Build Azure Class


elyse-weiss:
[here is some helpful code](https://gist.github.com/elyse-weiss/6bd367da334268b572aa9951446a3877) to get started with!

elyse-weiss:
We talked about this today!

pjsier:
I'm interested in working on this if it's available!

elyse-weiss:
@pjsier go for it! Just updated the first comment with some code we have on hand that never made it into a formal class!

PR #294, opened 2020-07-05
Add Azure Blob Storage connector

pjsier:
Progress on #73 (might close if it was only intended for Azure storage). Adds a connector for Azure Blob Storage primarily based on the Google Cloud Storage connector, but also pulling a bit from the S3 connector.

Here are a few things in particular I'm interested in feedback on:

- I split out `AZURE_ACCOUNT_NAME` and `AZURE_ACCOUNT_DOMAIN` from `AZURE_ACCOUNT_URL` because the account domain is almost always the same thing, and I haven't seen as strong of conventions for environment variables for Azure storage as I've seen around S3
- The example code uses an access token for authentication, and I broadened that here to support any credential accepted by the `BlobServiceClient`
- As a way of enforcing stricter typing the Azure storage SDK has a lot of wrapper types for things like setting headers when uploading blobs. To make this a little easier to use in `put_blob`, I'm pulling out `kwargs` so that people people can supply `content_type=` as a keyword argument instead of importing the `ContentSettings` type, initializing it with `content_type`, and passing it to `put_blob` in the `content_settings` keyword argument. If that's wrapping the API too much though I can remove it

Feedback is welcome! I also wrote up some notes on the process of writing a connector for the first time that I'll share in the Slack

### DIFF
--- a/docs/azure.rst
+++ b/docs/azure.rst
@@ -1,0 +1,74 @@
+Azure
+======
+
+************
+Blob Storage
+************
+
+========
+Overview
+========
+Azure Blob Storage is a cloud file storage system. It uses storage accounts to organize containers (similar to
+"buckets" for other storage providers) in which to store arbitrary files referred to as blobs.
+
+This connector currently only implements block blobs and not page or append blobs.
+
+==========
+Quickstart
+==========
+
+**List containers and blobs**
+
+.. code-block:: python
+
+  from parsons import AzureBlobStorage
+
+  azure_blob = AzureBlobStorage()
+
+  # Get all container names for a storage account
+  container_names = azure_blob.list_containers()
+
+  # Get all blob names for a storage account and container
+  blob_names = azure_blob.list_blobs(container_names[0])
+
+
+**Create a blob from a file or ``Table``**
+
+.. code-block:: python
+
+  from parsons import AzureBlobStorage, Table
+
+  azure_blob = AzureBlobStorage()
+
+  container_name = 'testcontainer'
+
+  # Upload a CSV file from a local file path and set the content type
+  azure_blob.put_blob(container_name, 'test1.csv', './test1.csv', content_type='text/csv')
+
+  # Create a Table and upload it as a JSON blob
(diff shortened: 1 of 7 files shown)


parsons/azure/azure_blob_storage.py
```diff
@@ -0,0 +1,390 @@
...
+        self.account_domain = check_env.check('AZURE_ACCOUNT_DOMAIN', account_domain)
+        self.account_url = os.getenv('AZURE_ACCOUNT_URL', account_url)
+        if not self.account_url:
+            self.account_url = f'https://{self.account_name}.{self.account_domain}/'
+        else:
+            if not self.account_url.startswith('http'):
+                self.account_url = f'https://{self.account_url}'
+            # Update the account name and domain if a URL is supplied
+            parsed_url = urlparse(self.account_url)
+            self.account_name = parsed_url.netloc.split(".")[0]
```
eliotst, line 49: I think if you move this block up before the `check_env` call for `account_name` and `account_domain`, then a user could pass in an `account_url` and not a `account_name` without the `check_env` raising an `Exception`.

pjsier: Ah good point, thanks for catching that! I'll make that change

docs/azure.rst
```diff
@@ -0,0 +1,74 @@
...
+************
+
+========
+Overview
+========
+Azure Blob Storage is a cloud file storage system. It uses storage accounts to organize containers (similar to
+"buckets" for other storage providers) in which to store arbitrary files referred to as blobs.
+
+This connector currently only implements block blobs and not page or append blobs.
+
```
eliotst, line 15: I apologize for not catching this on the first pass. We are trying to add a new section to our documentation for each connector around how to get account information. Could you add a small blurb and / or a link on how to get credentials for Azure?

pjsier: Should be updated now! There are a lot of different ways of generating credentials so I linked out for most of the details

shaunagm:
Thanks @pjsier and congrats on making the first Ragtag PR!  :confetti_ball:

PR #299, opened 2020-07-09
Redshift + S3: Move Some Logging To DEBUG

jburchard:
A first pass a moving a couple of table utility methods from `INFO` to `DEBUG`. This should reduce the size of our logs significantly.

@SorenSpicknall @eliotst - Anything else that we want to add to this PR? Just thread some ideas.

### DIFF
--- a/parsons/aws/s3.py
+++ b/parsons/aws/s3.py
@@ -115,7 +115,7 @@ class S3(object):
         """
 
         keys_dict = dict()
-        logger.info(f'Fetching keys in {bucket} bucket')
+        logger.debug(f'Fetching keys in {bucket} bucket')
 
         continuation_token = None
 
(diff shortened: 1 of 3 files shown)


eliotst:
I feel like one rule of thumb is we don't need to log at `info` information that the function caller already knew when calling or information that the function returns.

`S3.list_keys` logs "Fetching keys in X Bucket" and "Retrieved X keys".. both of which could be logged by the application if it wanted, since it gave the name of the bucket to the function, and gets a list of length X back from the function.

Likewise, `S3.key_exists` logs "Found {key} in {bucket}" vs. "Did not find {key} in {bucket}"

jburchard:
@eliotst - Okay, I added some of the feedback re: S3. I think that this is a good first start and we should just continue to think about specific methods that can be moved to DEBUG. I'll create an issue that we can just update as needed.

PR #304, opened 2020-07-13
New Azure class to handle account authentication methods

SorenSpicknall:
Added a new class, AzureAccountAccess, to eventually handle various methods of authentication for resource access. Initially, added one authentication method, using [TokenCredentials](https://docs.microsoft.com/en-us/python/api/azure-cosmosdb-table/azure.cosmosdb.table.common.tokencredential.tokencredential?view=azure-python).

The intent here is to get one authentication method up and running immediately, then build out this class in more detail down the road with other valid Azure authentication methods.

### DIFF
--- a/parsons/azure/azure_account_access.py
+++ b/parsons/azure/azure_account_access.py
@@ -1,6 +1,15 @@
 import adal
 
 class AzureAccountAccess(object):
+    """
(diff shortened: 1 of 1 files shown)


SorenSpicknall:
Closing because other, different functionality will be put in place to achieve the same goals.

Issue #233, opened 2020-05-04
ActionKit Updates


mkwoods927:
This issue is for building out functions to update events and signups in the ActionKit class. API documentation for these is:

- [Updating an event](https://roboticdogs.actionkit.com/docs/manual/api/rest/examples/updateevent.html)
- [Updating a signup](https://roboticdogs.actionkit.com/docs/manual/api/rest/examples/updatesignup.html)

rgriff23:
Can I work on this?

PR #307, opened 2020-07-14
add actionkit update events and signups

rgriff23:
This PR adds two new methods:

`ActionKit.update_event()`
`ActionKit.update_event_signup()`

I also updated `.gitignore` to prevent tracking of PyCharm project files.

Issue addressed: #233 

All pytest tests passed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,6 +91,9 @@ ENV/
 env.bak/
 venv.bak/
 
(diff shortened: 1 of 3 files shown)


eliotst:
This looks great @rgriff23 ! Thanks for the contribution. We should be able to review it in the next few days.

Issue #286, opened 2020-06-25
Improve contributor onboarding


shaunagm:
Highest Priority:

- [x] move contributing.md file to the top level directory
- [x] update contributing guide to link to useful_resources, not sample_code folder (must have gotten moved)
- [x] contributing.md should have clear walkthrough of steps, including cloning from github, setting up a virtualenv, up through running tests when they've finished installing
- [x] contributing.md (and maybe also readme.md) should include links to communications channels so people know how to talk to us, where to ask for help, how to report bugs, feature requests, etc
- [x] quickstart guide needs to not require creating accounts, setting env variables, etc.  need to pick a connectorwhere we can create a dummy account (or the API itself provides a dummy endpoint) that we can use here.
- [x] add instructions for building the docs (see https://github.com/move-coop/parsons/pull/313#issuecomment-660397875)

Lower Priority:
- [x] add "capitalize Parsons" to best practices in contrib docs
- [ ] contributing.md file should provide extra context/explanations for things that might be new to people, beyond just "run this command"
  - unit testing & what they should see when they run pytest
  - setting up a virtual environment
  - the difference between installing via pip and setting up as a developer (cloning, etc)
- [ ] we should make sure install is testing on Windows and Linux as well as Mac, and it's clear what people on those platforms need to do differently, if anything
- [ ] look into adding install instructions for Anaconda users

ehrenfoss:
I'm new to parsons and I'd like to help with this. I'm on Windows, and I have some Salesforce experience and can look into dummy account / developer account creation from the command line.

shaunagm:
That would be great, @ehrenfoss!  It looks like the authentication guide is the main thing the Salesforce docs need, although if you could update the quickstart guide to show the salesforce connector class being initialized with an environmental variable that'd be great too.  !

shaunagm:
Oh, sorry @ehrenfoss, I think I slightly misunderstood and thought you were responding to this issue (https://github.com/move-coop/parsons/issues/269) and offering to update the Salesforce connector docs.  If Salesforce does allow dummy accounts it would be a great choice for the quickstart guide.

pjsier:
Some additional guidance on how to format the docstrings would be useful. It looks like they might be following [Google's style guide](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings), and if they are stating that explicitly and linking to the document would probably solve that.

[...]

PR #309, opened 2020-07-14
Reorganize contributing.md

shaunagm:
Address some of the items in #286

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,6 +110,8 @@ The following is a list of best practices to consider when writing code for the 
 
 * Inline comments explaining complex codes and methods are appreciated.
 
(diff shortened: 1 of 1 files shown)


shaunagm:
There's a reference tucked away on line 31 (https://github.com/move-coop/parsons/pull/309/files#diff-6a3371457528722a734f3c51d9238c13R31) but we could definitely make it more prominent - maybe even put it in the list of ways to contribute at the very top?

eliotst:
Maybe add a second reference under the "Coding Conventions" header.. since there is some overlap there.. maybe something like "For more information about best practices, see the How to Build a Connector doc" or something..?

shaunagm:
I've updated my branch but weirdly it's not showing up here.  You can see it here: https://github.com/move-coop/parsons/compare/master...shaunagm:update-contribution-docs?expand=1

PR #314, opened 2020-07-17
Add slack classmethod to optionally send a message through a webhook rather than the full-access api_key

schuyler1d:


parsons/notifications/slack.py
```diff
@@ -86,6 +91,34 @@ def users(self, fields=['id', 'name', 'deleted', 'profile_real_name_normalized',
 
         return tbl
 
+    @classmethod
+    def message(cls, channel, text, webhook=None, api_key=None, parent_message_id=None):
```
eliotst, line 95: I think in the name of avoiding having multiple ways to do the same thing, could you remove the `api_key` fallback?

schuyler1d: done!

PR #315, opened 2020-07-17
action_kit: some upload features

schuyler1d:
1. bulk_upload_table set_only_columns argument 
2. collect upload_errors 
3. warn on missing email or user_id column uploads

### DIFF
--- a/parsons/action_kit/action_kit.py
+++ b/parsons/action_kit/action_kit.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import requests
+import time
 
 from parsons.etl.table import Table
 from parsons.utilities import check_env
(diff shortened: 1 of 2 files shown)


parsons/action_kit/action_kit.py
```diff
@@ -616,5 +625,40 @@ def _split_tables_no_empties(self, table):
...
         return results
+
+    def collect_upload_errors(self, result_array):
+        """
+        Collect any upload errors as a list of objects from bulk_upload_table 'results' key value.
+        This waits for uploads to complete, so it may take some time if you uploaded a large file.
+        `Args:`
+            result_array: list
+                After receiving a dict back from bulk_upload_table you may want to see if there
+                were any errors in the uploads.  If you call collect_upload_errors()
```
ydamit, line 645: Seems like this line might be incomplete?

parsons/action_kit/action_kit.py
```diff
@@ -616,5 +625,40 @@ def _split_tables_no_empties(self, table):
...
+        `Returns`:
+            [dict]
```
ydamit, line 647: I'm assuming the `error_data` result includes `id` so that it's joinable back to the original results. Might be important to include that here, along with any other relevant details about what we get back from the `uploaderror` endpoint.

schuyler1d:
dur, annoyingly, id to join against the original data isn't as easy -- but I document the 'upload' key that with some string munging you could join against.  At least for my use-case, it's not that important to do a join, because the split is done 'for me' with different options of bulk_upload_table() -- so errors are about *something* in the data from the single table.

Fwiw, in my use-case, I'm just detecting the existence at all of errors and then printing the messages (in a slack notification -- if we see anything or anything unusual, then we'd need to debug further (rather than handle different errors, though I suppose that could be useful for others, which is why I pass the whole object)

schuyler1d:
You might need to approve this one more time -- there was a merge conflict with Asher's recent `with` PR

Issue #310, opened 2020-07-16
VAN: EA Expand Fields Cause Error on MyVoters


ydamit:
The `get_person()` method includes a default list for the `expand_fields` arg, but one of those: `codes` returns a specific 400 error about that not being available, and two others: `contribution_history` and `organization_roles` return generic 500 errors.

Should be easy to solve by removing those list elements if `self.db_code == 0`

PR #316, opened 2020-07-18
VAN.get_person() - Fix MyVoters Error

jburchard:
The default fields returned from `VAN.get_person()` are only returned with EveryAction, not MyVoters. This PR removes them when the database is set to MyVoters.

Addresses issue #310.

### DIFF
--- a/parsons/ngpvan/people.py
+++ b/parsons/ngpvan/people.py
@@ -363,7 +363,7 @@ class People(object):
         # Removing the fields that are not returned in MyVoters
         NOT_IN_MYVOTERS = ['codes', 'contribution_history', 'organization_roles']
 
-        if self.connection.db == 0:
+        if self.connection.db_code == 0:
             expand_fields = [v for v in expand_fields if v not in NOT_IN_MYVOTERS]
 
         logger.info(f'Getting person with {id_type} of {id} at url {url}')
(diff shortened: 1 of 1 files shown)


parsons/ngpvan/people.py
```diff
@@ -360,6 +360,12 @@ def get_person(self, id, id_type='vanid', expand_fields=[
 
         expand_fields = ','.join([json_format.arg_format(f) for f in expand_fields])
 
+        # Removing the fields that are not returned in MyVoters
+        NOT_IN_MYVOTERS = ['codes', 'contribution_history', 'organization_roles']
+
+        if self.db == 0:
```
eliotst, line 366: I think you want this to be `self.connection.db_code` ... `self.db` the value passed in to the `__init__` by the user, which appears to be either `MyVoters`, `MyMembers`, `MyCampaign`, or `EveryAction`. `self.connection.db_code` is the numeric value.

jburchard:
Ready for re-review.

parsons/ngpvan/people.py
```diff
@@ -360,6 +360,12 @@ def get_person(self, id, id_type='vanid', expand_fields=[
 
         expand_fields = ','.join([json_format.arg_format(f) for f in expand_fields])
 
+        # Removing the fields that are not returned in MyVoters
+        NOT_IN_MYVOTERS = ['codes', 'contribution_history', 'organization_roles']
+
+        if self.connection.db == 0:
```
ydamit, line 366: I think you still want this to be `self.connection.db_code == 0` OR `self.connection.db == 'MyVoters'` (I prefer the former).

jburchard:
Ready for final review. Sorry for all of the back and forth on this.

PR #325, opened 2020-07-28
Bump Python Version

jburchard:
Update the setup to reflect that we support Python 3.8

Follows PR #303.


eliotst:
Good catch. Could you also update the README.md while you're there? Right now it says `Parsons is only compatible with Python 3.6/7` .. Could we update it to say `Parsons works best with Python 3.6, 3.7, and 3.8` or something?

PR #331, opened 2020-08-06
Create infrastructure for sample code

shaunagm:
This is my first stab at creating a new infrastructure for handling sample code.  I expect it to need significant edits before merging.

This PR:

- adds a README describing our sample script process, and including a table listing existing scripts
- creates a template_script.py file that outlines how we'd like the scripts to be structure
- edits the existing four scripts to be in line with the template

Some particular feedback I'd like:

- do the label names I've chosen work?  once I get the thumbs up, I'll create the labels and update the PR so the label links in the README link to the labels.
- does the structure of the template_script look good?
- does how we're now handling config variables look good?  I want to isolate config variables in their own section at the top so they're easy to use.  Currently these scripts may overwrite environmental variables, which may not be ideal.
- have I correctly described what each of the four sample scripts does?

Also if anyone can actually _test_ the refactored sample scripts that'd be swell.

### DIFF
--- a/useful_resources/sample_code/README.md
+++ b/useful_resources/sample_code/README.md
@@ -1,0 +1,20 @@
+# Sample Scripts
+
+The scripts in this folder perform data operations, typically moving data between one connector and another. To perform the same operation, all you should need to do is change the configuration variables in the `Configuration Variables` section of each script. You can also use these scripts as a jumping off point to create new scripts doing similar opeations.
+
+If you can't find the script you want here, you can create an issue in the tracker with the label [sample script request](). Please put in as much detail as possible what you would like the script to do.
+
+If you have a script you'd like to add, you have two options. You can create an issue in the tracker with the contents of your script, and add the label [sample script to add](). Or you can add the script yourself.
+
+If you wish to add the script yourself, please use the `template_script.py` file so that your contribution's structure is consistent with all the other scripts.  Please keep most comments as these are directed at the eventual user of your script, but delete any comments labeled "//To Script Writer//" once you understand their advice.
+
+Please also add your new script to the table below.
+
+# Existing Scripts
+
+| File Name   | Brief Description |  Connectors Used   |
+| ----------- | ----------- |  ----------- |
+| apply_activist_code.py  | Gets activist codes stored in Redshift and applies to users in Van | Redshift, VAN|
+| s3_to_redshift.py  |  Moves files from S3 to Redshift| Redshift, S3|
+| s3_to_s3.py | Get files from vendor s3 bucket and moves to own S3 bucket | S3 |
+| update_user_in_actionkit.py | Adds a voterbase_id (the Targetsmart ID) to users in ActionKit |Redshift, ActionKit |
--- a/useful_resources/sample_code/README.md
+++ b/useful_resources/sample_code/README.md
@@ -2,9 +2,9 @@
 
 The scripts in this folder perform data operations, typically moving data between one connector and another. To perform the same operation, all you should need to do is change the configuration variables in the `Configuration Variables` section of each script. You can also use these scripts as a jumping off point to create new scripts doing similar opeations.
 
-If you can't find the script you want here, you can create an issue in the tracker with the label [sample script request](). Please put in as much detail as possible what you would like the script to do.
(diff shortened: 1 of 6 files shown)


useful_resources/sample_code/apply_activist_code.py
```diff
@@ -1,22 +1,40 @@
...
-set_env_var(os.environ['REDSHIFT_HOST'])
-set_env_var(os.environ['REDSHIFT_CREDENTIAL_USERNAME'])
-set_env_var(os.environ['REDSHIFT_CREDENTIAL_PASSWORD'])
-rs = Redshift()
+# Setup
 
-# AWS setup - this assumes a Civis Platform parameter called "AWS"
+for name, value in config_vars.items():    # sets variables if provided in this script
+    if value.strip() != "":
+        os.environ[name] = value
```
eliotst, line 35: I think as long as `value` is a non-blank value, it makes sense to overwrite the environment variable's value.

useful_resources/sample_code/apply_activist_code.py
```diff
@@ -1,22 +1,40 @@
...
-set_env_var(os.environ['REDSHIFT_PORT'])
-set_env_var(os.environ['REDSHIFT_DB'])
-set_env_var(os.environ['REDSHIFT_HOST'])
-set_env_var(os.environ['REDSHIFT_CREDENTIAL_USERNAME'])
-set_env_var(os.environ['REDSHIFT_CREDENTIAL_PASSWORD'])
-rs = Redshift()
+# Setup
 
-# AWS setup - this assumes a Civis Platform parameter called "AWS"
+for name, value in config_vars.items():    # sets variables if provided in this script
```
eliotst, line 33: This might be worth pulling out into a utility function at some point in Parsons-proper that we could build out over time for configuring the environment.

useful_resources/sample_code/template_script.py
```diff
@@ -0,0 +1,45 @@
...
+
+for name, value in config_vars.items():    # if variables specified above, sets them as environmental variables
+    if value.strip() != "":
+        os.environ[name] = value
+
+# //To Script Writer// : instantiate connectors here, eg: rs = Redshift().
+
+# Logging
+
+logger = logging.getLogger(__name__)
```
eliotst, line 37: We have a Parsons `logger` that is exported out of the `parsons` package. It might make sense to just use that, and encourage users to just use the Parsons logger. We could build out some more user-friendly log configuration functionality too, since I find Python logging config to be pretty complicated sometimes.

useful_resources/sample_code/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+Please also add your new script to the table below.
+
+# Existing Scripts
+
+| File Name   | Brief Description |  Connectors Used   |
+| ----------- | ----------- |  ----------- |
+| apply_activist_code.py  | Gets activist codes stored in Redshift and applies to users in Van | Redshift, VAN|
+| s3_to_redshift.py  |  Moves files from S3 to Redshift| Redshift, S3|
+| s3_to_s3.py | Get files from vendor s3 bucket and moves to own S3 bucket | S3 |
```
eliotst, line 19: I think "vendor" to "own" is maybe what the original script was for, but I think the script can be used for transferring data between any two buckets. I think we can generalize this to just "source" bucket to "target" bucket or something.

useful_resources/sample_code/s3_to_s3.py
```diff
@@ -1,26 +1,41 @@
...
+# Set the configuration variables below or set environmental variables of the same name and leave these
+# with empty strings.  We recommend using environmental variables if possible.
+
+config_vars = {
+        # S3 (source) (note: this assumes a Civis Platform parameter called "AWS")
+        "S3_TEMP_BUCKET": "",
+        "AWS_ACCESS_KEY_ID": "",
+        "AWS_SECRET_ACCESS_KEY": "",
+        # S3 (vendor) (note: this assumes a Civis Platform parameter called "AWS_VENDOR")
+        'AWS_VENDOR_SECRET_ACCESS_KEY': "",
```
eliotst, line 18: If we repurpose as "source" / "destination", we would need to change these environment variable keys.. `AWS_SOURCE_SECRET_ACCESS_KEY`, `AWS_DESTINATION_SECRET_ACCESS_KEY`, etc.

PR #332, opened 2020-08-07
Postgres: Fix generate_data_types

dannyboy15:
Currently the `generate_data_types` method doesn't correctly handle `float` for "NA" values. For example:
This table
| c1   | c2   |   c3 | c4   | c5   | c6   | c7   |
|------|------|------|------|------|------|------|
| a    |      |  1   | NA   | 1.4  | 1    | 2    |
| b    |      |  2   | NA   | 1.4  | 1    | 2    |
| c    |      |  3.4 | NA   |      |      | a    |
| d    |      |  5   | NA   | 1.4  | 1    | 2    |
| e    |      |  6   | NA   | 1.4  | 1    | 2    |
| f    |      |  7.8 | NA   | 1.4  | 1    | 2    |
| g    |      |  9   | NA   | 1.4  | 1    | 2    |

gets interpreted with types
`['varchar', 'varchar', 'smallint', '', 'varchar', 'varchar', 'varchar']` 
instead of
`['varchar', 'varchar', 'decimal', 'varchar', 'decimal', 'smallint', 'varchar']`

### DIFF
--- a/parsons/databases/postgres/postgres_create_statement.py
+++ b/parsons/databases/postgres/postgres_create_statement.py
@@ -82,7 +82,8 @@ class PostgresCreateStatement(object):
             return 'varchar'
 
         if type(t) in [int, float]:
-            if (type(t) in [int] and current_type not in ['float', 'varchar']):
+            if (type(t) in [int] and
+                    current_type not in ['decimal', 'varchar']):
 
                 # Make sure that it is a valid integer
                 if not self.is_valid_integer(val):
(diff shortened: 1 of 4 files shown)


jburchard:
Is this also an issue with Redshift?

jburchard:
Based on our conversation, seems like we should push this and then create a separate PR for the cross dialect class?

dannyboy15:
Yeah, a separate PR makes sense for that. I'll make the update to the Redshift class for now though.

Issue #319, opened 2020-07-21
Load CSV data from an S3 manifest


eliotst:
Redshift uses [manifest files](https://docs.aws.amazon.com/redshift/latest/dg/loading-data-files-using-manifest.html) when loading data from S3 (using `COPY`) and when exporting data into S3 (using `UNLOAD`). It could be useful to have a method on the `S3` class that would be able to load multiple CSV files from S3 into a single `Table`.

Under the hood, the method would just need to:
  1) download the manifest file from S3
  2) use the manifest to figure out which files to grab from S3
  3) load each of the files into a `Table`
  4) concatenate all of the files into a single `Table`

johnarky:
Unless dannyboy15 will be assigned, I'd like to do this.
Is there any problem with the manifest?
Do the files to be concatenated have the same columns?  Is this a concern for Parsons?

shaunagm:
Status of issue: @dannyboy15 is working on this.

dannyboy15:
I think we can close this issue since the PR was merged.

PR #335, opened 2020-08-07
Table: Add support for manifests files to from_s3_csv

dannyboy15:
Addresses #319, @eliotst was this what you had in mind?


parsons/etl/tofrom.py
```diff
@@ -675,10 +678,26 @@ def from_s3_csv(cls, bucket, key, aws_access_key_id=None, aws_secret_access_key=
...
-        return cls(petl.fromcsv(file_obj, **csvargs))
+            s3_files = [x["url"] for x in manifest["entries"]]
+
+        else:
+            s3_files = [file_obj]
+
+        tbls = []
+        for file in s3_files:
+            # TODO handle urls that end with '/', i.e. urls that point to "folders"
+            _, _, bucket_, key_ = files.split("/", 3)
```
eliotst, line 693: I think you want to split `file` here, not `files`.

eliotst: I haven't had a chance to test this out, but it looks like if `from_manifest_file` is `False`, then `file` will be a local file path (return value from `get_file`), where if `from_manifest_file` is `True`, then `file` will be the URL from the manifest file. I think you can move this `split` and `get_file` after you get the URLs from the manifest (line 685). That way `s3_files` will always be a list of local paths.

eliotst:
@dannyboy15 Just wanted to check in here. Do you have any questions on my review?

dannyboy15:
@eliotst Sorry for the delay here. When I wrote it initially, I didn't have a way to test it. I just pushed some changes that should fix the errors. I tested also tested but let me know if you see any other issues.

eliotst (approved):
This looks like it'll work. Thanks!

Issue #301, opened 2020-07-10
Redshift get_table_definition method based on non-standard Redshift views


eliotst:
The `get_table_definition` + `get_table_definitions` methods in `Redshift` are based on views that aren't standard in Redshift, so these functions will not work for most people. We should either rework the SQL in the methods to use only standard Redshift functionality, or else remove them.

dannyboy15:
I ran into this issue as well. I fixed it locally by including the sql that generates the views directly in parsons. I'll open a PR to push my changes.

ChrisC:
Just noting that we merged in these changes but had to walk them back in #589 because we were seeing errors in some of our pipelines that use these methods.

PR #336, opened 2020-08-07
Redshift Table Utilities: fix get definition methods

dannyboy15:
The sql for the underlying queries is now included instead of relying on views to exist in the database.

Addresses #301

### DIFF
--- a/parsons/databases/redshift/queries/asl-license.txt
+++ b/parsons/databases/redshift/queries/asl-license.txt
@@ -1,0 +1,39 @@
+Amazon Software License
+
+1. Definitions
+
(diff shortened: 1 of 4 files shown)


eliotst:
Thanks for this @dannyboy15 ! We are discussing internally whether or not we can find a solution that doesn't require adding another license. I will update this PR in the next couple of days to let you know what we found.

dannyboy15:
> Thanks for this @dannyboy15 ! We are discussing internally whether or not we can find a solution that doesn't require adding another license. I will update this PR in the next couple of days to let you know what we found.

@eliotst are there any updates about this?

Issue #266, opened 2020-05-27
Add Bloomerang class


mkwoods927:
This ticket is for adding a Bloomerang class to Parsons.

[API documentation is here.](https://bloomerang.co/features/integrations/api/rest-api/)

The most important capabilities will be:

- Pull constituents and their emails, addresses, and phone numbers
- Pull donations, recurring donations, recurring donation payments, pledges, and pledge payments
- Pull custom fields and custom field values
- Create & update constituents
- Create donations, recurring donations, and recurring donation payments
- Create interactions

rgriff23:
I'll take this one on if that's alright. I won't have API access, so I can't do a complete testing.

mkwoods927:
Adding more specifics on what methods would be helpful here! Methods to pull lists in addition to individual records would definitely be helpful.

In terms of constituents and contact info, it looks like get/constituent and get/constituents both return the primary e-mail, primary phone, and custom fields and values, so it's not a high priority to create separate methods for those. I don't think address gets returned though, so get/addresses would be useful!

For transactions, both get/transactions and get/transactions/designations are valuable. The first gives some basic info, while the second provides more detail like custom field data, campaign, and fund.

rgriff23:
> In terms of constituents and contact info, it looks like get/constituent and get/constituents both return the primary e-mail, primary phone, and custom fields and values, so it's not a high priority to create separate methods for those. I don't think address gets returned though, so get/addresses would be useful!

It actually looks like PrimaryAddress (with street, city, state, etc) is a field returned for constituents: https://bloomerang.co/features/integrations/api/rest-api#/Constituents/get_constituent__id_

mkwoods927:
Oh great, then no need to make a separate method!

shaunagm:
Closing this as the class was created by https://github.com/move-coop/parsons/pull/337. Any enhancements should be in their own issue. Thanks again, Randi!

PR #337, opened 2020-08-08
Bloomerang class, docs, and tests

rgriff23:
This is a WIP addressing Issue #266. It's my first attempt at a connector, so I'm pausing for feedback. 

Summary of work:

- Added Bloomerang class, including methods for authorization and get/post/put/delete requests for constituents. I used the Hustle connector as my main example since it similarly uses headers for Bearer authorization. 
- Added Bloomerang docs, modeled after ActionKit
- Added a few tests, modeled after ActionKit

It would be very helpful if someone could point me to good examples for writing tests, since I don't have much experience with this, but I want to learn. I tried to set up some simple tests by following ActionKit as an example, but I'm not confident in my work. Currently my 'test_get_constituent' function is failing but I don't understand why, since I tried to set it up exactly the same as the ActionKit 'test_get_user' function. 

I'd also appreciate clarity on which methods should be added. For example, the issue description says "Pull constituents and their emails, addresses, and phone numbers", but it isn't clear to me if this is all covered by 'get_constituent', or if we want separate methods for emails, addresses, and phone numbers, since those are API endpoints as well. I'm also not sure if I should be adding methods for working with lists in addition to individual items (i.e., a 'get_constituents' method in addition to 'get_consituent').

### DIFF
--- a/parsons/bloomerang/bloomerang.py
+++ b/parsons/bloomerang/bloomerang.py
@@ -36,7 +36,8 @@ class Bloomerang(object):
     def __init__(self, api_key=None, client_id=None, client_secret=None):
         self.api_key = check_env.check('BLOOMERANG_API_KEY', api_key, optional=True)
         self.client_id = check_env.check('BLOOMERANG_CLIENT_ID', client_id, optional=True)
-        self.client_secret = check_env.check('BLOOMERANG_CLIENT_SECRET', client_secret, optional=True)
+        self.client_secret = check_env.check('BLOOMERANG_CLIENT_SECRET', client_secret,
+                                             optional=True)
         self.uri = URI
         self.uri_auth = URI_AUTH
         self.conn = self._conn()
--- a/parsons/bloomerang/bloomerang.py
+++ b/parsons/bloomerang/bloomerang.py
@@ -55,7 +55,7 @@ class Bloomerang(object):
             self._generate_access_token()
             headers['Authorization'] = f"Bearer {self.access_token}"
         else:
-            logger.warning('Initializing API Connector without authorization credentials.')
+            raise Exception('Missing authorization credentials.')
         return APIConnector(uri=self.uri, headers=headers)
 
     def _generate_authorization_code(self):
@@ -78,6 +78,10 @@ class Bloomerang(object):
             url = url + f'{entity_id}/'
         return url
 
+    @staticmethod
+    def _base_pagination_params(page_number=1, page_size=50):
+        return {'skip': page_size * (page_number - 1), 'take': min(page_size, 50)}
+
     def _base_create(self, endpoint, entity_id=None, **kwargs):
         return self.conn.post_request(url=self._base_endpoint(endpoint, entity_id),
                                       json=json.dumps({**kwargs}))
@@ -86,8 +90,8 @@ class Bloomerang(object):
         return self.conn.put_request(url=self._base_endpoint(endpoint, entity_id),
                                      json=json.dumps({**kwargs}))
 
-    def _base_get(self, endpoint, entity_id=None):
-        return self.conn.get_request(url=self._base_endpoint(endpoint, entity_id))
+    def _base_get(self, endpoint, entity_id=None, params=None):
+        return self.conn.get_request(url=self._base_endpoint(endpoint, entity_id), params=params)
 
     def _base_delete(self, endpoint, entity_id=None):
         return self.conn.delete_request(url=self._base_endpoint(endpoint, entity_id))
@@ -96,7 +100,9 @@ class Bloomerang(object):
         """
         `Args:`
             **kwargs:`
-                Fields to include, e.g., FirstName = 'Rachel'
+                Fields to include, e.g., FirstName = 'Rachel'.
+
+                See the Bloomerang API docs for a full list of `fields <https://bloomerang.co/features/integrations/api/rest-api#/Constituents/post_constituent>`_. # noqa
         """
         return self._base_create('constituent', **kwargs)
 
@@ -106,7 +112,9 @@ class Bloomerang(object):
             constituent_id:
                 Constituent ID to update
             **kwargs:`
-                Fields to update, e.g., FirstName = 'RJ'
+                Fields to update, e.g., FirstName = 'RJ'.
+
+                See the Bloomerang API docs for a full list of `fields <https://bloomerang.co/features/integrations/api/rest-api#/Constituents/put_constituent__id_>`_. # noqa
         """
         return self._base_update('constituent', entity_id=constituent_id, **kwargs)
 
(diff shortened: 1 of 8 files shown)


mkwoods927:
Thank you @rgriff23 ! I added some more detail in the issue #266  to answer your questions about which methods would be most useful. Unfortunately I do not have experience writing tests but I'm sure folks who do will chime in!

eliotst:
Thanks @rgriff23 . This is great! In terms of unit testing, we have a number of Connectors that use the `requests` library to make calls to HTTP APIs. Generally, we leverage the `requests-mock` library (already in requirements.txt) to mock out the server to make unit testing that much easier.

An example is the Phone2Action Connector's tests: https://github.com/move-coop/parsons/blob/master/test/test_p2a.py

We also have the `parsons.utilities.api_connector.APIConnector` class that we often use to do some basic wrapping around the `requests` library. You would recommend you take a look for managing calls to the HTTP endpoint if it will work for what you need.

test/test_bloomerang/test_bloomerang.py
```diff
@@ -0,0 +1,74 @@
...
+    @requests_mock.Mocker()
+    def test_delete_constituent(self, m):
+        m.delete(f'{self.bloomerang.uri}constituent/{CONSTITUENT_ID}/', json=TEST_CONSTITUENT_DELETE)
+        self.assertEqual(self.bloomerang.delete_constituent(CONSTITUENT_ID), TEST_CONSTITUENT_DELETE)
+
+    @requests_mock.Mocker()
+    def test_get_constituents(self, m):
+        m.get(f'{self.bloomerang.uri}constituents?skip=0&take=50&id=1|2|3/', json=TEST_CONSTITUENTS)
+        #self.assertEqual(self.bloomerang.get_constituents(CONSTITUENT_ID_LIST), TEST_CONSTITUENTS['Results'])
+        self.assertEqual(Table(self.bloomerang.get_constituents(CONSTITUENT_ID_LIST)), Table(TEST_CONSTITUENTS['Results']))
```
rgriff23, line 74: @eliotst @shaunagm I posted a question in the Ragtag Parsons channel, but in case it isn't easy to answer it there, I updated my PR to highlight the code giving me issues. 

Line 74 currently fails, and just above it in line 73 you can see the commented out 'passing' version. As you can see, the lines are identical except that the failing version wraps both sides of the assert with `Table`.

eliotst: For connector methods that return multiple objects like `get_constituents`, we want the method to return a `Table`. Does the test pass if you update `get_constituents` to return a `Table` and you wrap `TEST_CONSTITUENTS` in a `Table`.. like:

```
self.assertEqual(self.bloomerang.get_constituents(CONSTITUENT_ID_LIST), Table(TEST_CONSTITUENTS['Results']))
```

rgriff23: No, the test does not pass if I do that (that is how I originally tried to implement the method and test, and it results in the same error).

eliotst: Okay. I will take a look this afternoon!

rgriff23: Cool, thanks in advance for the help... once this is resolved I should be able to make a final PR for the connector

eliotst: Okay.. I think I got it.. I think that comparison between `Table` objects doesn't work as expected.. so the workaround here is if you a) return a `Table` from `get_constituents`, and b) in the test, pass the result from `get_constituents` (a `Table`) into `list`, you can compare that `list` with `TEST_CONSTITUENTS['Results']`.

So:

```
    @requests_mock.Mocker()
    def test_get_constituents(self, m):
        m.get(f'{self.bloomerang.uri}constituents?skip=0&take=50&id=1|2|3/', json=TEST_CONSTITUENTS)
        #self.assertEqual(self.bloomerang.get_constituents(CONSTITUENT_ID_LIST), TEST_CONSTITUENTS['Results'])
        self.assertEqual(list(self.bloomerang.get_constituents(CONSTITUENT_ID_LIST)), TEST_CONSTITUENTS['Results'])
```

rgriff23: Gotcha, I can implement this workaround. I also just discovered another workaround:

The reason I expected I would be able to compare two `Table` objects is that I saw `ActionNetwork().get_people()` comparing two `Table` objects in `test_get_people`. However, the `ActionNetwork` test uses `assert_matching_tables` from `test.utils` rather than `unittest.TestCase.assertEqual` (which is what I'm using).

I found that using `assert_matching_tables` to test for the equality between the `Table` objects works as well.

rgriff23:
@EliotStone this PR is ready for a final review. 

Summary:

The Bloomerang class includes the following public methods:

- Constituents: create, update, get, delete, get list
- Transactions: create, update, get, delete, get list
- Transaction designations: get, get list
- Interactions: create, update, get, delete, get list

Also included are working tests for all methods, an RST doc, and doc strings for all public methods.

I added Bloomerang's '200' to default success codes for APIConnector request methods.

eliotst:
Awesome. Thanks @rgriff23 ! I will take a look at this tomorrow. Thanks for all your great work!

eliotst (changes requested):
This is great. Thanks for all of your hard work on this. I just had a couple of questions about the list records methods (e.g. `get_transactions`).

parsons/bloomerang/bloomerang.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def delete_constituent(self, constituent_id):
+        """
+        `Args:`
+            constituent_id:
+                Constituent ID to delete
+        """
+        return self._base_delete('constituent', entity_id=constituent_id)
+
+    def get_constituents(self, constituent_ids):
```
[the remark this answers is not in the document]
rgriff23: @eliotst thanks, almost done with revisions. One follow question about your last suggestion: Should the `page_number` parameter be required, or optional? If the latter, what should the behavior be when no `page_number` is provided?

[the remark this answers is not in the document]
eliotst: Good question. What we have done in other places is if no `page_number` is provided, we have the Connector page through all of pages from the API and return a single `Table` with everything (e.g. [Phone2Action](https://github.com/move-coop/parsons/blob/cc449e31a5e76bc3bd451d43cd6e3902769cba4b/parsons/phone2action/p2a.py#L35-L54)).. but that's a non-trivial change, and definitely more nice-to-have. I think for this PR it's okay to just default to the first page (`0` or `1` depending on what the API considers "first").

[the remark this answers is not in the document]
rgriff23: Okay, I'll keep it simple and default to the first page for this PR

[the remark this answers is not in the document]
rgriff23: > it might be worth testing whether or not we can pass a `take` higher than the default 50 items. 50 is a little small for a page, especially if someone is trying to download a lot of constituents.

It actually seems like 50 is the maximum for this API

eliotst (approved):
This is great. Thanks!

parsons/bloomerang/bloomerang.py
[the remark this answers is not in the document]
shaunagm: Quick follow up - given this discussion about pagination, do you think we should add some guidance to the How To Build a Connector docs? It seems like this would be a common issue.

Issue #269, opened 2020-06-02
Improve connector documentation


shaunagm:
The connector documentation to be more consistent.  Each connector needs, in order:

1) an overview, which explains the service that the connector connects to, and summarizes what subsets of endpoints currently are covered.  
2) a detailed guide to getting authentication info.  
3) a "quick start guide" that demonstrates instantiating the connector and using a method.  

See [the action kit connector documentation](https://move-coop.github.io/parsons/html/action_kit.html) for an example of what we're looking for.

This doesn't need to be done by a single person in a single PR - feel free to grab a connector by leaving a comment below and submit the updated docs as its own PR.

Here's the current status of the connector docs (thanks to @rgriff23 for the audit!):

| Connector | Overview | Authentication Guide | Quickstart | Notes | 
| -------------- | :--------- | ----------: | ----------: | ----------: | 
| Action Kit | ✔️ | ✔️ | ✔️ | |
| Action Network |  ✔️| ✔️ | ✔️ | |
| Airtable | ✔️ | ✔️ |  ✔️ | |
| AWS: ALL |   ||| needs a global overview at the top to orient the reader |
| AWS: Lambda | to add | to add | to add | |
| AWS: S3 | ✔️ | ✔️| ✔️ | |
| AWS: Redshift | ✔️ |  ✔️| ✔️| |
| Azure: Blob Storage | ✔️ |  ✔️| ✔️| |
| Bill.com | ✔️ |  ✔️| ✔️| |
| Braintree |  ✔️ | ✔️ | ✔️ | |
| Civis |  ✔️ | ✔️ | ✔️ | |
| Copper |  ✔️ | ✔️ | ✔️ | |
| CrowdTangle |  ✔️ | ✔️ | ✔️ | |
| Databases: ALL | ✔️ | | |  |
| Databases: MySQL | ✔️ | ✔️ | ✔️ | |
| Databases: Postgres | ✔️ | ✔️ | ✔️ | |
| Databases: Database Sync | ✔️ | ✔️  | ✔️ | |
| Facebook Ads |  ✔️ | ✔️ | ✔️ | |
| FreshDesk |  ✔️ | ✔️ | ✔️ | |
| Github | ✔️ | ✔️  | ✔️ | |
| Google: ALL | | | | needs a global overview at the top to orient the reader |
| Google: Big Query | ✔️| to add | to add | |
| Google: CloudStorage | needs improvement| to add| to add | |
| Google: Civic | ✔️ | to add| to add| |
[...]

PR #342, opened 2020-08-10
Docs: Update ActionNetwork.

jburchard:
Improves overview and adds authentication.

#269

### DIFF
--- a/docs/action_network.rst
+++ b/docs/action_network.rst
@@ -20,7 +20,7 @@ To instantiate a class, you can either pass in the API token as an argument or s
 
 .. code-block:: python
 
-   from parsons import VAN
+   from parsons import ActionNetwork
 
    # First approach: Use API credentials via environmental variables
    an = ActionNetwork()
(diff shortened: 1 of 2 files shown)


docs/action_network.rst
```diff
@@ -1,18 +1,36 @@
...
 ***********
 
+To instantiate a class, you can either pass in the API token as an argument or set the ``AN_API_TOKEN`` environmental variable.
+
 .. code-block:: python
-	
-	from parsons import ActionNetwork, Table
 
-	an = ActionNetwork(API_TOKEN)
+   from parsons import VAN
```
rgriff23, line 23: Should import ActionNetwork, not VAN.

I'm guessing this was copied from ActionKit, where I previously introduced this exact typo. Maybe you can fix the ActionKit bug in your PR too, especially since it is referenced as an example for documentation?

jburchard: Good catch. And, yes, your theory is correct.

I've actually got this fixed in the S3 docs PR for AK, so I think that it should be fixed shortly there too.

PR #344, opened 2020-08-11
docs: update quickstart

eliotst:
This commit updates the Quick Start code in the README.md file
for Parsons so that the quick start can be executed without
credentials. The goal is to make it so that anyone could run
the Quick Start code to begin understanding how to use Parsons.

This commit also adds a `download_table` method to the `GitHub`
Connector that allows downloading CSV data from GitHub as a
Parsons `Table`.

Finally, this commit tweaks the `download_file` method on the
`GitHub` connector to support token-based authentication when
downloading files / tables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,41 +34,33 @@ We have a Parsons Docker container hosted on [DockerHub](https://cloud.docker.co
 
 ### Quickstart
 
+For this Quickstart, we are looking to generate a list of voters with cell phones using a [dummy data file](docs/quickstart.csv). We use the `assert` statements to verify that the data has been loaded correctly.
+
+```python
+# Download the Census data from the Parsons GitHub repository
+from parsons import GitHub
+github = GitHub()
+dummy_data = github.download_table('move-coop/parsons', 'docs/quickstart.csv')
+assert dummy_data.num_rows == 1000  # Check that we got all 1,000 people
+
+# Filter down to people with cell phones
+people_with_cell_phones = dummy_data.select_rows(lambda row: row['is_cell'] == 'true')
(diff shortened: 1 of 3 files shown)


parsons/github/github.py
```diff
@@ -365,18 +369,55 @@ def download_file(self, repo_name, path, branch=None, local_path=None):
...
+            branch: Optional[str]
+                Branch to download file from. Defaults to repo default branch
+            local_path: Optional[str]
+                Local file path to download file to. Will create a temp file if not supplied.
+            delimiter: Optional[str]
+                The CSV delimiter to use to parse the data. Defaults to ','
+
+        Returns:
+            Table
+                Table built with the CSV data
```
jburchard, line 419: Typically, we use this syntax for methods that return tables.
 
```
            Parsons Table
                See :ref:`parsons-table` for output options.
```

README.md
```diff
@@ -34,41 +34,33 @@ We have a Parsons Docker container hosted on [DockerHub](https://cloud.docker.co
 
 ### Quickstart
 
+For this Quickstart, we are looking to generate a list of voters with cell phones using a [dummy data file](docs/quickstart.csv). We use the `assert` statements to verify that the data has been loaded correctly.
+
```
shaunagm, line 38: Can we put a quick line in here about authentication?  We could do a short description with a link to the Github connector docs for more info.

shaunagm: Or does Github require no authentication whatsoever for public repositories?

eliotst: Yeah. GitHub doesn't require auth for public repos.

Issue #347, opened 2020-08-14
Add Cloud Storage Documentation


jburchard:
We've got a hidden utility called [cloud storage](https://github.com/move-coop/parsons/blob/master/parsons/utilities/cloud_storage.py) that tries to solve for the problem of POST methods that require access to a file on the internet.

This is currently being utilized for [VAN.upload_scores](https://move-coop.github.io/parsons/html/ngpvan.html#parsons.ngpvan.van.Scores.upload_scores) but will also be needed for VAN Bulk Upload endpoints. 

If this is used more extensively, it could benefit from some documentation.

PR #349, opened 2020-08-15
VAN: Bulk Upload Initial Methods

jburchard:
This PR contains the initial commits for the Bulk Upload functionality. The Bulk Import endpoint is pretty powerful and allow you to update contacts and contributions. For this initial functionality, I focused on a more basic type, activist codes.

**Public Methods**
- `get_bulk_import_job()`
- `get_bulk_import_mapping_types()`
- `get_bulk_import_mapping_type()`
- `bulk_apply_activist_codes()`

For future PRs:

- `bulk_create_update_person` - The ability to update contact data on a person level.
- Better validation prior to sending the request (checking that required columns are in the table.
- Convenience method to download the results of a bulk import job as a Parsons Table.

Additional Needs:
- Better documentation and additional options for cloud storage. Outlined in Issue #347.

### DIFF
--- a/parsons/ngpvan/bulk_import.py
+++ b/parsons/ngpvan/bulk_import.py
@@ -4,6 +4,7 @@ from parsons.utilities import cloud_storage
 
 import logging
 import uuid
+import csv
 
 logger = logging.getLogger(__name__)
 
@@ -79,8 +80,13 @@ class BulkImport(object):
 
         # Move to cloud storage
         file_name = str(uuid.uuid1())
-        url = cloud_storage.post_file(tbl, url_type, file_path=file_name + '.zip', **url_kwargs)
+        url = cloud_storage.post_file(tbl,
+                                      url_type,
+                                      file_path=file_name + '.zip',
+                                      quoting=csv.QUOTE_ALL,
+                                      **url_kwargs)
         logger.info(f'Table uploaded to {url_type}.')
+        print (url)
 
         # Generate request json
         json = {"description": description,
@@ -89,7 +95,7 @@ class BulkImport(object):
                     "columns": [{'name': c} for c in tbl.columns],
                     "fileName": file_name + '.csv',
                     "hasHeader": "True",
-                    "hasQuotes": "False",
+                    "hasQuotes": "True",
                     "sourceUrl": url},
                 "actions": [{"resultFileSizeKbLimit": 5000,
                              "resourceType": resource_type,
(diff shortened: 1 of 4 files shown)


parsons/ngpvan/bulk_import.py
```diff
@@ -24,3 +28,127 @@ def get_bulk_import_resources(self):
         r = self.connection.get_request(f'bulkImportJobs/resources')
         logger.info(f'Found {len(r)} bulk import resources.')
         return r
+
+    def get_bulk_import_job(self, job_id):
+        """
+        Get a bulk import job status.
+
+        .. note::
```
WillRaphaelson, line 36: This is somewhat surprising to me - have you experienced this?

test/test_van/test_bulkimport.py
```diff
@@ -19,8 +23,116 @@ def tearDown(self):
     @requests_mock.Mocker()
     def test_get_bulk_import_resources(self, m):
 
-        json = ['Contacts', 'Contributions', 'ActivistCodes', 'ContactsActivistCodes']
+        json = ['Contacts', 'Contributions',
```
WillRaphaelson, line 26: This will depend on the permissions the key has - it could also have RecurringCommitments for example.

parsons/ngpvan/bulk_import.py
jburchard: Well, I was unable to recreate it, so I removed. Thanks for lifting this up.

test/test_van/test_bulkimport.py
jburchard: Definitely. 

I think that this works for now, as it is just mocking the expected schema, rather than a live connection.

parsons/ngpvan/bulk_import.py
```diff
@@ -24,3 +28,123 @@ def get_bulk_import_resources(self):
...
+        logger.info(f'Table uploaded to {url_type}.')
+
+        # Generate request json
+        json = {"description": description,
+                "file": {
+                    "columnDelimiter": 'csv',
+                    "columns": [{'name': c} for c in tbl.columns],
+                    "fileName": file_name + '.csv',
+                    "hasHeader": "True",
+                    "hasQuotes": "False",
```
eliotst, line 92: It looks like the Python `csv.writer` (via `petl`) is putting quotes in when necessary.. which appears to be the default (`csv.QUOTE_MINIMAL`).

```python
from parsons import Table
tbl = Table([{'one': 'one', 'two': 'two,'}])
tbl.to_csv()
```

Produces:

```
one,two
one,"two,"
```

It might be best to write out the CSV file with `csv.QUOTE_ALL` and then set `hasQuotes` to `True` always.

`quoting` is a keyword arg for the `csv.writer`, which you can pass into the `petl.tocsv`.. I'm not sure though how to wire that up through the `post_file` call above.

jburchard: Fixed.

PR #351, opened 2020-08-18
phone2action: add update_advocate / create_advocate

### LINKED ISSUE #327: Phone2Action Update Advocate
Need to create endpoints for ```update_advocate()```  for both opt-in/opt-out and update phone number.
We need to be able to both push and pull data from the contacts ```phone``` , ```smsOptin```, ```advocateid ```, ```smsOptout```, and ```campaigns``` in order to update the contact's phone and or opt-out  or opt-in status.

[API Docs}(http://docs.phone2action.com/#overview)

eliotst:
This commit adds the `update_advocate` and `create_advocate`
methods to the `Phone2Action` Connector class.

This commit also updates the `APIConnector` class to allow users
to only have to pass in the relative path to an endpoint on
request calls to simplify the API. It also updates any classes
that were passing in the full URL to only pass in the relative
path. *Note:* It is still possible to pass in a full URL; the
full URL will override the base URL in the connector.

Finally, this commit tweaks the existing date utilities by
a) adding a new `parse_date` function for parsing semi-arbitrary
date values, and b) to rework the `iso_to_unix` function to the
`date_to_timestamp` function that can take any date-like value
and convert it into a UNIX timestamp.

---
Fixes #327 

I need to do some integration testing before merging, but wanted to get the PR up for review.

cc @abarrett9


parsons/phone2action/p2a.py
```diff
@@ -29,33 +30,26 @@ def __init__(self, app_id=None, app_key=None):
         self.app_id = check_env.check('PHONE2ACTION_APP_ID', app_id)
         self.app_key = check_env.check('PHONE2ACTION_APP_KEY', app_key)
         self.auth = HTTPBasicAuth(self.app_id, self.app_key)
-        self.uri = PHONE2ACTION_URI
-
-    def _request(self, url, args=None):
-        # Internal request method
-
-        r = requests.get(url, auth=self.auth, params=args)
-        r.raise_for_status()
```
jburchard, line 38: Any issues here with removing the raise for status and the validate response?

parsons/phone2action/p2a.py
```diff
@@ -152,17 +147,216 @@ def get_campaigns(self, state=None, zip=None, include_generic=False, include_pri
...
+        url = 'campaigns'
 
         args = {'state': state,
                 'zip': zip,
                 'includeGeneric': str(include_generic),
                 'includePrivate': str(include_private)
                 }
 
-        tbl = Table(self._request(url, args=args).json())
+        tbl = Table(self.client.get_request(url, params=args))
```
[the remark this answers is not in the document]
eliotst: Good question. I think line 150 used to include building out the URL, but that was simplified with the new API Connector API. I can just get rid of the local variable.

docs/p2a.rst
```diff
@@ -4,5 +4,39 @@ Phone2Action
...
+   p2a = Phone2Action()
+
+   # Get all advocates updated in the last day
+   import datetime
+   today = datetime.datetime.utcnow()
+   yesterday = today - datetime.timedelta(days=1)
+
+   # get_advocates returns a dictionary that maps the advocate data (e.g. phones) to a parsons
+   # Table with the data for each advocate
+   advocates_data = p2a.get_advocates(updated_since=yesterday)
```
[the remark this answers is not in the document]
eliotst: It can be either a `datetime`, `str`, or an int (UNIX timestamp). I can update the arg docstring accordingly.

parsons/phone2action/p2a.py
eliotst: Since we are now leveraging the `APIConnector`, we get that functionality from the `APIConnector`'s `get_request` / `post_request` methods.

parsons/phone2action/p2a.py
```diff
@@ -152,17 +147,216 @@ def get_campaigns(self, state=None, zip=None, include_generic=False, include_pri
...
+                The ID(s) of campaigns to add the advocate to
+            firstname: str
+                `Optional`: The first name of the advocate
+            lastname: str
+                `Optional`: The last name of the advocate
+            email: str
+                `Optional`: An email address to add for the advocate. One of ``email`` or ``phone``
+                is required.
+            phone: str
+                `Optional`: An phone # to add for the advocate. One of ``email`` or ``phone`` is
```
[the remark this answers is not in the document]
eliotst: I copied this from VAN (though I see VAN uses semicolons after `Optional`). If we're good with that format, I can go through in this PR and clean up the more obvious non-conforming docstrings.

https://github.com/move-coop/parsons/blob/8a7cb4a829a334ff7361b6f8e1500170f016da8f/parsons/ngpvan/people.py#L387-L392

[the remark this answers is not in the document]
jburchard: Let's just leave as is. I think that in the future, we are going to need a giant find/replace across the library to just standardize it.

parsons/phone2action/p2a.py
jburchard: My mistake. All good!

parsons/action_network/action_network.py
```diff
@@ -209,7 +209,7 @@ def get_tag(self, tag_id):
             A  JSON of the entry. If the entry doesn't exist, Action Network returns
             "{'error': 'Couldn't find tag with id = <id>'}"
         """
-        return self.api.get_request(url=f"{self.api_url}/tags/{tag_id}")
+        return self.api.get_request(url=f"tags/{tag_id}")
```
jburchard, line 212: Could you do this for all of the methods in the AN class as well?

eliotst: I thought I got all of the methods in the ActionNetwork class. Could you put out an example of what you are seeing needs to be fixed?

jburchard: My mistake. Misread the code.

PR #353, opened 2020-08-25
Docs: Airtable

jburchard:
Add a quickstart and update the name of the client to be `client`.

Issue #269.


jburchard:
It probably should be? I felt as though it was a little bit more clear. I might file an issue for discussion.

Issue #297, opened 2020-07-09
Build SurveyGizmo Class


elyse-weiss:
TMC wrote some code that we know works: https://gist.github.com/elyse-weiss/101d251eeed2e4f2fb25e8754ca493bd

Eager to finalize as Parsons connector.

coreyhaines:
Have started converting it over. Need to write tests for the functions there.

shaunagm:
Addressed in https://github.com/move-coop/parsons/pull/354

PR #354, opened 2020-08-27
SurveyGizmo connector

coreyhaines:
Initial creation of SurveyGizmo connector based on the code in the issue https://github.com/move-coop/parsons/issues/297

[gist](https://gist.github.com/elyse-weiss/101d251eeed2e4f2fb25e8754ca493bd)

Have done manual testing
```python
>>> sg = SurveyGizmo(api_token=api_token, api_token_secret=api_token_secret)
>>> sg.get_surveys()
surveygizmo INFO Found 1 surveys.
{'id': '5820711', 'team': '1018675', 'type': 'Standard Survey', 'status': 'Launched', 'created_on': '2020-09-02 16:45:03', 'modified_on': '2020-09-02 16:46:46', 'title': 'Test Gizmo Connector'}
```

I would like to punt the automated microtests to a follow-up PR, in order to get this into availability for folks.

@elyse-weiss 
There are 3 methods in the provided gist that are not directly API wrappers, instead appearing to be helpers that were written for a specific case. Not sure whether we want to move these over. Or perhaps into a `SurveyGizmo.Helpers` or something.

- _strip_html
- parse_responses
- parse_respondents

### DIFF
--- a/parsons/surveygizmo/surveygizmo.py
+++ b/parsons/surveygizmo/surveygizmo.py
@@ -1,8 +1,10 @@
 import logging
 import surveygizmo
+from parsons.etl import Table
 from parsons.utilities import check_env
 
 logger = logging.getLogger(__name__)
+
 
 class SurveyGizmo(object):
     """
@@ -22,10 +24,14 @@ class SurveyGizmo(object):
     """
 
     def __init__(self, api_token, api_token_secret):
-        self.api_token=check_env('SURVEYGIZMO_API_TOKEN', api_token)
-        self.api_token_secret=check_env('SURVEYGIZMO_API_TOKEN_SECRET', api_token_secret)
+        self.api_token = check_env('SURVEYGIZMO_API_TOKEN', api_token)
+        self.api_token_secret = check_env('SURVEYGIZMO_API_TOKEN_SECRET', api_token_secret)
 
-        self.client = surveygizmo.SurveyGizmo(api_version='v5', api_token=self.api_token api_token_secret=self.api_token_secret)
+        self.client = surveygizmo.SurveyGizmo(
+                api_version='v5',
+                api_token=self.api_token,
+                api_token_secret=self.api_token_secret
+                )
 
     def get_surveys(self):
         """
--- a/parsons/surveygizmo/surveygizmo.py
+++ b/parsons/surveygizmo/surveygizmo.py
@@ -39,7 +39,7 @@ class SurveyGizmo(object):
                 api_token_secret=self.api_token_secret
                 )
 
-    def get_surveys(self):
+    def get_surveys(self, page=None):
         """
         Get a table of lists under the account.
 
@@ -50,12 +50,13 @@ class SurveyGizmo(object):
             Table Class
         """
 
-        r = self._client.api.survey.list()
+        r = self._client.api.survey.list(page)
         data = r['data']
 
-        while r['page'] < r['total_pages']:
-            r = self._client.api.survey.list(page=(r['page']+1))
-            data.extend(r['data'])
+        if not page:
+            while r['page'] < r['total_pages']:
+                r = self._client.api.survey.list(page=(r['page']+1))
+                data.extend(r['data'])
 
         tbl = Table(data).remove_column('links')
         tbl.unpack_dict('statistics', prepend=False)
@@ -76,7 +77,7 @@ class SurveyGizmo(object):
             Table Class
         """
 
-        r = self._client.api.surveyresponse.list(survey_id)
+        r = self._client.api.surveyresponse.list(survey_id, page)
         logger.info(f"{survey_id}: {r['total_count']} responses.")
         data = r['data']
 
--- a/parsons/surveygizmo/surveygizmo.py
+++ b/parsons/surveygizmo/surveygizmo.py
@@ -85,8 +85,9 @@ class SurveyGizmo(object):
             while r['page'] < r['total_pages']:
                 r = self._client.api.surveyresponse.list(survey_id, page=(r['page']+1))
                 data.extend(r['data'])
-                logger.info(f"{survey_id}: Retrieving {r['page']} of {r['total_count']}.")
 
         tbl = Table(data).add_column('survey_id', survey_id, index=1)
+
+        logger.info(f"Found #{tbl.num_rows} responses.")
 
         return tbl
--- a/parsons/surveygizmo/surveygizmo.py
+++ b/parsons/surveygizmo/surveygizmo.py
@@ -23,12 +23,13 @@ class SurveyGizmo(object):
         SurveyGizmo Class
     """
 
-    def __init__(self, api_token, api_token_secret):
-        self.api_token = check_env('SURVEYGIZMO_API_TOKEN', api_token)
-        self.api_token_secret = check_env('SURVEYGIZMO_API_TOKEN_SECRET', api_token_secret)
+    def __init__(self, api_token=None, api_token_secret=None, api_version='v5'):
+        self.api_token = check_env.check('SURVEYGIZMO_API_TOKEN', api_token)
+        self.api_token_secret = check_env.check('SURVEYGIZMO_API_TOKEN_SECRET', api_token_secret)
+        self.api_version = check_env.check('SURVEYGIZMO_API_VERSION', api_version)
 
         self.client = surveygizmo.SurveyGizmo(
-                api_version='v5',
+                api_version=self.api_version,
                 api_token=self.api_token,
                 api_token_secret=self.api_token_secret
                 )
--- a/parsons/surveygizmo/surveygizmo.py
+++ b/parsons/surveygizmo/surveygizmo.py
@@ -44,7 +44,9 @@ class SurveyGizmo(object):
         Get a table of lists under the account.
 
         `Args:`
-            None
+            page : int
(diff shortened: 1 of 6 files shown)


coreyhaines:
@elyse-weiss @SorenSpicknall @pjsier Which surveygizmo library is required?

I thought maybe `surveygizmo==1.2.3` https://github.com/rpkilby/SurveyGizmo

SorenSpicknall:
> @elyse-weiss @SorenSpicknall @pjsier Which surveygizmo library is required?
> 
> I thought maybe `surveygizmo==1.2.3` https://pypi.org/project/SurveyGizmo/

Corey, that makes sense for me. Since this is fresh, we can rely on the most up-to-date stable version in PyPI.

shaunagm:
What's the status of this?  Really excited by the progress!

coreyhaines:
@shaunagm 

I tested it out manually on Wednesday

```python
>>> sg = SurveyGizmo(api_token=api_token, api_token_secret=api_token_secret)
>>> sg.get_surveys()
surveygizmo INFO Found 1 surveys.
{'id': '5820711', 'team': '1018675', 'type': 'Standard Survey', 'status': 'Launched', 'created_on': '2020-09-02 16:45:03', 'modified_on': '2020-09-02 16:46:46', 'title': 'Test Gizmo Connector'}
```

I want to write some automated tests for it, but I think it could be merged in with a follow-up to write the tests.

Then a matter of adding more endpoints as requested.

shaunagm:
We don't necessarily need to merge it for me to use it today - I can install from your fork, I believe.

coreyhaines:
@shaunagm 

Purrfect! I'll be around for the working session, so I can help fix any issues, add any other endpoints, etc.

eliotst:
Thanks for this, @coreyhaines ! My preference would be that we get at least one ["happy path"](https://en.wikipedia.org/wiki/Happy_path) test for each method in the class in this PR before merging. I can provide some context here, and then I am happy to answer any follow up questions you have.

(Note: I am working on a doc to explain how to write unit tests in Parsons, and much of the text below is taken from that.. so I apologize if it is a bit over-explain-y).

Most of our Parsons Connector classes fall into two categories:

1. Classes that call directly into HTTP API's using the Python `requests` library
2. Classes that wrap around third party Python libraries, often using their own "client" class

SurveyGizmo is the second type, since you are pulling in the `surveygizmo` library.

The category that the class falls into generally determines how one will write unit tests for the class. No matter which category it is, we are looking to use [mocks](https://en.wikipedia.org/wiki/Mock_object) to simulate the class' interactions with the external API. However, the category of Connector determines *how* we simulate the external API.

For classes that wrap third party libraries (eg SurveyGizmo), we are looking to simulate whatever client object that serves as the Python API for the calls to the external service. To simulate the client object, we can use the [`unittest.mock.MagicMock`](https://docs.python.org/3/library/unittest.mock.html) class from the Python standard library. 

The `MagicMock` class allows us to build a fake version of the client object by simulating functionality we expect our Connector class to use -- namely by returning canned responses to method calls on the client. As well, the `MagicMock` class is recording those calls as they happen, so that after we have run through our tests, we can check that things happened in the way we expected.

The following gives a simple example of initializing, setting up, and using a `MagicMock` class:

```
from unittest.mock import MagicMock

# Initialize a MagicMock object to serve as our fake client
mock_client = MagicMock()
# Tell the client to return a canned list of people from its list_people method
mock_client.list_people.return_value = [
    {'id': 1, 'name': 'Nicole Jackson'},
]

# Call our mock client's list_people method
people = mock_client.list_people()
# Check that we got our expected data
assert len(people) == 1
assert people[0]['id'] == 1

# Check that the list_people method was called
mock_client.list_people.assert_called()
```

The [`Salesforce` class](https://github.com/move-coop/parsons/blob/master/parsons/salesforce/salesforce.py) is a good example for writing tests for Connector classes written against a third party library. The `Salesforce` Parsons Connector class wraps around the `simple-salesforce` library's `Salesforce` client. When testing the `Salesforce` Parsons class, we will  need to swap out its reference to the `simple-salesforce` client with a mock client.

In the `SalesforceTest` class, this is done in the [`setUp` method](https://github.com/move-coop/parsons/blob/0e0e607eb706012f2a285116f8af139c2e8cd0c8/test/test_salesforce/test_salesforce.py#L8-L32)  of the test class:

```
        self.sf = Salesforce()
        self.sf._client = mock.MagicMock()
```

The `_client` attribute on the `Salesforce` Connector class holds the class' reference to the underlying third party client object. By overriding it with our `MagicMock` object, the `Salesforce` Parsons class will be calling methods on our mock client instead of an actual `simple-salesforce` client.

We can then set up our mock client's `query_all` method:

```
        self.sf._client.query_all.return_value = [{'Id': 1, 'value': 'FAKE'}]
```

Now, we can test our `Salesforce` Parsons Connector's `query` method:

```
        # Call the query method with a fake value
        response = self.sf.query('FAKESOQL') 
        # Check that our mock client's query_all method was also called with the fake value
        assert self.sf._client.query_all.called_with('FAKESOQL')
        # Check that the response from our query method is what we expect
        self.assertEqual(response[0]['value'], 'FAKE')
```

In the first line, we call the method we are testing (`query`) with a fake value. In the next line, we check to make sure our mock client's `query_all` method was called with the same fake value. Finally, we test to make sure that our `Salesforce` Connector returned the expected response, which is based on the return value of the mock client's `query_all` method (which we set up in the previous block).

That's pretty much all there is to it. When writing tests for a Connector wrapping a third party library, we will almost always:

  1. Create a mock client using the `MagicMock` class, and wire up the methods that our Connector will need
  2. Replace the actual third party library's client on our Connector class with our mock
  3. Call the method(s) on the Connector that we are looking to test
  4. Verify the return value of the method calls is what we expect
  5. Verify that the Connector called the expected methods on our mock client
...

Hopefully that's useful. Again, sorry if it's a bit long. Let me know if you have any questions!

coreyhaines:
@eliotst totally makes sense. My goal with rushing was to get it so that sunrise could use it when we were writing the script, but it turns out we could do it without it being merged. I'll get the tests written this week.

Thanks for the detailed response!

eliotst:
Sure! Also, let me know if you run into issues. Thanks!

parsons/surveygizmo/surveygizmo.py
```diff
@@ -0,0 +1,91 @@
...
+
+        `Args:`
+            survey_id: string
+                The id of survey for which to retrieve the responses.
+
+        `Returns:`
+            Table Class
+        """
+
+        r = self._client.api.surveyresponse.list(survey_id)
```
eliotst, line 79: You need to pass the `page` argument through here.

parsons/surveygizmo/surveygizmo.py
```diff
@@ -0,0 +1,91 @@
...
+
+        return tbl
+
+    def get_survey_responses(self, survey_id, page=None):
+        """
+        Get the responses for a given survey.
+
+        `Args:`
+            survey_id: string
+                The id of survey for which to retrieve the responses.
```
eliotst, line 73: Could you add the `page` argument to the `Args` documentation here?

parsons/surveygizmo/surveygizmo.py
```diff
@@ -0,0 +1,91 @@
...
+        self.api_token_secret = check_env.check('SURVEYGIZMO_API_TOKEN_SECRET', api_token_secret)
+        self.api_version = check_env.check('SURVEYGIZMO_API_VERSION', api_version)
+
+        self._client = surveygizmo.SurveyGizmo(
+                api_version=self.api_version,
+                api_token=self.api_token,
+                api_token_secret=self.api_token_secret
+                )
+
+    def get_surveys(self):
```
eliotst, line 42: Could you add `page` parameter to this method in the name of consistency across methods?

eliotst (changes requested):
Thanks for all of the work on the connector and the tests! I just had a couple comments around the pagination parameters.

parsons/surveygizmo/surveygizmo.py
coreyhaines: Done

coreyhaines: Done

parsons/surveygizmo/surveygizmo.py
coreyhaines: Done

parsons/surveygizmo/surveygizmo.py
coreyhaines: Done

coreyhaines:
@eliotst Thanks for the reviews. I can't seem to figure out how to set up the mocks to return different values for subsequent calls to support testing the pagination code, though. I'm happy to put that in, though.

I think otherwise there are some reasonable tests here.

eliotst:
@coreyhaines no worries on testing pagination right now. Looking in the [mock docs](https://docs.python.org/3/library/unittest.mock-examples.html#side-effect-functions-and-iterables), you would use `side_effect` on the mock and set it to a list.

eliotst (approved):
This is really great. Thanks so much for your work on this!

coreyhaines:
Thanks @eliotst . I'll take a look at `side_effect` for future reference.

shaunagm:
:tada: :tada: :tada: Congrats and thank you, Corey, for your first connector!

PR #359, opened 2020-09-01
Redshift: get_columns_list()

elyse-weiss:


parsons/databases/redshift/rs_table_utilities.py
```diff
@@ -389,6 +389,23 @@ def get_columns(self, schema, table_name):
...
+        `Args:`
+            schema: str
+                The schema name
+            table_name: str
+                The table name
+        `Returns:`
+            A list of column names.
+        """
+
+        first_row = rs.query(f"select * from {schema}.{table_name} limit 1")
```
jburchard, line 406: Should be self.query(...

elyse-weiss: Fixed!

PR #360, opened 2020-09-01
redshift: add region parameter to copy from S3

### LINKED ISSUE #356: Specify S3 Region in DBSync
We need a way to specify the s3 region of the destination bucket in DBSync class.

Received this error: 
image

Per conversation in Slack:
In the parsons docs it looks like `parsons.Redshift.unload` supports aws_region in the arguments, but I don’t see a corresponding argument in `parsons.Redshift.copy`

eliotst:
This commit adds the `aws_region` parameter to the `copy_s3`
method in the `Redshift` connector class to support loading keys
from an S3 bucket outside the current Redshift cluster's region.
The `aws_region` populates the `region` parameter in the `copy`
statement used to load the S3 data.

cc @elyse-weiss 

Fixes #356 

---
I don't have a good set up to test this. @elyse-weiss would you be able to test this locally for your use case?

### DIFF
--- a/parsons/aws/s3.py
+++ b/parsons/aws/s3.py
@@ -52,16 +52,6 @@ class S3(object):
         self.client = self.s3.meta.client
         """Boto3 API Session client object. Use for more advanced boto3 features."""
 
-    @property
-    def region_name(self):
-        """
-        Get the name of the AWS region for the S3 client.
-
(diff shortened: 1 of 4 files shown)


elyse-weiss:
this is working for me locally!

PR #366, opened 2020-09-04
RockTheVote: add support for extended report

dannyboy15:
This adds the option to request the extended report which add a few extra fields. In addition, it adds support for the testing url.

h/t @natashamathur

### DIFF
--- a/parsons/rockthevote/rtv.py
+++ b/parsons/rockthevote/rtv.py
@@ -41,18 +41,24 @@ class RockTheVote:
             The RockTheVote partner ID for the RTV account
         partner_api_key: str
             The API Key for the partner
+        testing: bool
+            Whether or not to use the staging instance
     `Returns`:
         RockTheVote class
     """
 
-    def __init__(self, partner_id=None, partner_api_key=None):
+    def __init__(self, partner_id=None, partner_api_key=None, testing=False):
         self.partner_id = check_env.check('RTV_PARTNER_ID', partner_id)
         self.partner_api_key = check_env.check('RTV_PARTNER_API_KEY', partner_api_key)
 
-        self.client = APIConnector('https://register.rockthevote.com/api/v4/',
-                                   headers=REQUEST_HEADERS)
+        if testing:
+            self.client = APIConnector(
+                "https://staging.rocky.rockthevote.com/api/v4", headers=REQUEST_HEADERS)
+        else:
+            self.client = APIConnector(
+                "https://register.rockthevote.com/api/v4", headers=REQUEST_HEADERS)
 
-    def create_registration_report(self, before=None, since=None):
+    def create_registration_report(self, before=None, since=None, report_type="extended"):
         """
         Create a new registration report.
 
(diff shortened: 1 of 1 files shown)


natashamathur:
For context, the reason I set up the report_type parameter the way it is was because

The existing code did not include a report_type parameter, and I wanted to make sure requests without that parameter still went through to minimize downstream effects and
I wasn't sure if RTV was going to release more report types, or if it was always going to be just "extended" vs "normal/default".

Overall, this should add the ability to use the testing environment, and to pull an "extended" report.

parsons/rockthevote/rtv.py
```diff
@@ -41,26 +44,36 @@ class RockTheVote:
...
         self.partner_api_key = check_env.check('RTV_PARTNER_API_KEY', partner_api_key)
 
-        self.client = APIConnector('https://register.rockthevote.com/api/v4/',
-                                   headers=REQUEST_HEADERS)
+        if testing:
+            self.client = APIConnector(
+                "https://staging.rocky.rockthevote.com/api/v4", headers=REQUEST_HEADERS)
+        else:
+            self.client = APIConnector(
+                "https://register.rockthevote.com/api/v4", headers=REQUEST_HEADERS)
```
jburchard, line 62: Super picky, but can you move these URIs to the top of the script?

```
TESTING_URI = "https://staging.rocky.rockthevote.com/api/v4"
PRODUCTION_URI = "https://register.rockthevote.com/api/v4"
```

parsons/rockthevote/rtv.py
```diff
@@ -41,26 +44,36 @@ class RockTheVote:
...
             before: str
-                Date before which to return registrations for
+                Limit to registrations that were started before this date, in
+                ISO format (e.g. 2020-01-01)
             since: str
-                Date for filtering registrations
+                Limit to registrations that were started since this date, in
+                ISO format (e.g. 2020-01-01)
+            report_type: str
+                The type of report to create. Currently only accepts "extended".
```
jburchard, line 76:
```
Currently only accepts ``extended``.
```
Also, could you explain what happens in the docs when it is left as ``None``?

parsons/rockthevote/rtv.py
```diff
@@ -196,18 +223,27 @@ def run_registration_report(self, before=None, since=None, poll_interval_seconds
...
             before: str
-                Date before which to return registrations for
+                Limit to registrations that were started before this date, in
+                ISO format (e.g. 2020-01-01)
             since: str
-                Date for filtering registrations
+                Limit to registrations that were started since this date, in
+                ISO format (e.g. 2020-01-01)
+            report_type: str
+                The type of report to run. Currently only accepts "extended".
```
jburchard, line 232: Currently only accepts ``extended``.
```
Also, could you explain what happens in the docs when it is left as ``None``?

jburchard (changes requested):
This is really great. Just a few minor comments.

jburchard (approved):

Issue #248, opened 2020-05-18
Better SFTP 'helpers'


schuyler1d:
I have a few use-cases where some 'helper-ish' functions or restructuring in the SFTP driver would be useful.  Some use-cases:

* Using a connection for several steps at once (e.g. making create_connection() public and maybe each method takes an optional conn= argument)
* File and directory patterns -- instead of dir listing, then file listing, it would be nice to pass some regex patterns for a directory and file name, and then either download all files that match or at least list all the files that match.
* When downloading a file, sometimes there are some additional processing steps, but mostly after the steps, there's a CSV to turn into a table -- it would be nice to get a function (possibly with a lambda processing arg in-between), that downloads and returns the Table object straight-away

eliotst:
Two of the three completed (1 and 3).

tiburona:
@eliotst I could try to close out this issue by working on (2) if no one else is in progress on it.

jburchard:
That would be great!

tiburona:
As long as I'm in the `sftp` module, I'll note that `get_file` and `put_file` do not have the if/else structure I'd expect (and that other methods in the class do): 

```
if connection: 
    do something 
else: 
    with self.create_connection() as connection:
        do something else
```
 I can't tell whether this is an inconsequential bug or intentional for a reason I can't see.  If it is intentional, it would be nice to have a comment explaining it just so it doesn't distract future contributors.  I'd be happy to add it if I knew what it should say.

PR #368, opened 2020-09-07
Issue 248 regex sftp helper

tiburona:
I am opening this PR as a draft to solicit comments before I fuss about writing tests.  
I have written created four new public methods on SFTP:

- `list_files`
- `list_subdirectories`
- `get_files`
- `walk_tree`

All of these methods take a regex pattern to match file name or directory path as appropriate, and `walk_tree` takes both.  These regex patterns can always be None, in which case everything is retrieved.  

`walk_tree` returns a tuple, a list of directories traversed and a list of files touched.  `walk_tree` also takes a max_depth parameter which defaults to 2, i.e., search the provided directory and its subdirectories.  

`get_files` can take a list of files as well as a remote path or a list of remote paths, so the user could use some of these other methods to generate a file list and then feed it to get_files.  Note: I haven't really tested `get_files `yet; it's probably broken.

All the public methods have doc strings that spell this out in a bit more detail.

I also wrote a @connect decorator that appears to work and hides this kind of logic under the covers.
```
if connection:
     return connection.do_something()
else:
     with self.create_connection() as connection:
         return connection.do_something()

```
It was getting kind of cumbersome given that I had methods that called other methods and they all used connection. If people don't like it I can get rid of it, but if people do like it I could apply it to other methods in the module.  

Anyway, I just wanted to check in about the new functionality of the module before I got really detail oriented about testing.  If people wanted these methods to do different things, or return things in different formats, or wanted something eliminated or added, it would be better to make those changes before I polish.


tiburona:
Hi @eliotst @jburchard I would welcome comments on this draft PR before I polish it -- i.e. feedback on the methods I created, what they take and return, anything I should add or take away, etc.  See my introductory comments above.  Thanks!

jburchard:
This is really great. Thank for you this. I really appreciate the @connect decorator.

A few questions and ideas:

1. It might make sense to use the decorators for all of the existing methods, though perhaps that would make sense as a separate PR.

2. Is `list_files()` thought to be a replacement for the existing `list_directory()` method?

tiburona:
> Is list_files() thought to be a replacement for the existing list_directory() method?

That's not how I was thinking of it.  The current list_directory() method gets all entries (files and dirs) and doesn't take a regex pattern.  I think it's at least reasonable to have a method that exactly mirrors `.listdir` and others to accomplish more complicated search (the same way `os.listdir` and `os.path.walk` both exist).

I did just slightly refactor, though, because I realized that my recursive method was going to take twice as long as it needed to.  Now `walk_tree` returns a tuple with a list of directories and a list of files, and it no longer calls those methods.  So it's arguable that because the `list_subdirectories` and `list_files` functionality is achievable with `walk_tree` I should get rid of them.  `list_subdirectories` and `list_files` and  are the equivalent of taking the first and second returned value of `walk_tree` respectively with max_depth=1.  I guess it's just a question of whether there's some additional user friendliness in having `list_files` and `list_subdirectories` methods.  I could just add a lot of examples to the `walk_tree` docstring.

Sorry to be indecisive!

eliotst:
@schuyler1d Do you have any feedback so far as the author issue #248?

jburchard:
@tiburona 

1. I'm all for having convenience methods that overlap a little bit. So, I'd leave in the `list_subdirectories` and `list_files` methods.

2. Would it be possible to clarify the docstrings with `list_directory()` to state that both files and directories are returned?

3. Lastly, is it possible to move the `get_files()` method directly after the `get_file()` method?

Overall, this looks really great. Let me know when it is ready for a full review and thank you for this!

tiburona:
Thanks for all the feedback!  I can make those changes, write tests, etc. and I'll ping you when it's ready.

tiburona:
@eliotst @jburchard -- I finally marked this ready for review.

@shaunagm can testify that I struggled mightily to rebase so this had a nice commit history, and I do think in the course of my struggles I learned something about what I did wrong but I got to a point where past mistakes had made fixing this more energy than it was worth since you would just squash it all anyway.  So I have a very strange commit history.  You will see that there's a commit that claims to incorporate a commit from another contributor.  I don't even know how that happened.

jburchard (approved):
This is really great. Thank you so much!

Issue #281, opened 2020-06-22
Remove Empty Lists from Action Network add_person() POST request


ydamit:
Right now, all arguments aside from `email_address` have default values and those default values that are empty lists `[]` cause the call to return an error. Filtering out any arguments that haven't been supplied in the final call should solve for that pretty simply.

PR #369, opened 2020-09-07
fix actionnetwork empty list default value bug

rgriff23:
This is an attempt to close issue #281. I followed the suggestion to filter for supplied arguments when creating the data dictionary. Not sure if there is a cleaner way to do this. 

Working on this PR raised several questions for me:

1. Is there an acceptable default value that can be used in place of the empty lists, thereby avoiding all the ugly if statements? 

2. The Action Network API docs say that only `email_address` is required to POST a new person, so why does the `ActionNetwork` class have arguments for these particular non-required fields rather than having all the optional fields passed in as `**kwargs`? https://actionnetwork.org/docs/v2/post-people/

3. Unrelated to the specific bug in this PR, I noticed that the ActionNetwork Quick Start example in the RST fails (line 29 in action_network.rst) because it does not specify an 'api_url' argument. My question is, why does the class require a user-provided 'api_url'? Won't this argument always be 'https://actionnetwork.org/api/v2'? Something needs to be changed, but I'm not sure whether it is the Quick Start example or the class itself.

### DIFF
--- a/parsons/action_network/action_network.py
+++ b/parsons/action_network/action_network.py
@@ -6,6 +6,8 @@ import logging
 
 logger = logging.getLogger(__name__)
 
+API_URL = 'https://actionnetwork.org/api/v2'
+
 
 class ActionNetwork(object):
     """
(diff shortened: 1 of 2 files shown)


jburchard:
@ydamit

ydamit:
I can complete this review by the end of tomorrow (Wed 9/9)!

ydamit (approved):
Looks right to me!

eliotst:
I don't know for sure.. but the ActionNetwork API docs make it look like it would always just be https://actionnetwork.org/api/v2 . Unless @ydamit has a good reason not to, I think we can just have it as a constant somewhere in the module. I think you could definitely do that as part of this PR.

ydamit:
Agreed! The person who initially wrote this thought leaving this exposed would make it more future-proof, but I think it will be a quite a long time before AN changes their API version.

rgriff23:
Thanks @ydamit @eliotst, I think this should be ready to merge

PR #372, opened 2020-09-08
Redshift:  update statupdate and compupdate copy options

dannyboy15:
This would change the default behavior of copying tables into Redshift. It would align more with the defaults that Redshift uses.


eliotst:
Thanks for this, @dannyboy15 ! This change seems reasonable to me, but I am not super familiar with the trade offs for these flags.

@schuyler1d or @jburchard do you have any thoughts on this? I know you two have done a lot of thinking around `compupdate` specifically, especially with regards to `upsert`. Do either of you have thoughts on defaults for `compupdate`?

eliotst:
I also want to flag the merge conflict on `test/test_redshift.py`

dannyboy15:
I think at the very least we should merge in the change that can handle `statupdate=None`.

dannyboy15:
As for the point about leaving the default as `True`, I wonder if there's any precedent there. Have other methods been implemented in such a way that they deviate from the underlying services default to encourage best practices or achieve optimization?

parsons/databases/redshift/redshift.py
```diff
@@ -401,9 +401,12 @@ def copy(self, tbl, table_name, if_exists='fail', max_errors=0, distkey=None,
                 characters.
             statupate: boolean
                 Governs automatic computation and refresh of optimizer statistics at the end
-                of a successful COPY command.
+                of a successful COPY command. If ``True`` explicitly sets ``statupate`` to on, if
+                ``False`` explicitly sets ``statupate`` to off. Defaults to ``None``.
```
jburchard, line 405: Please include a note telling the user that  

``If STATUPDATE is used, the current user must be either the table owner or a superuser.``

parsons/databases/redshift/redshift.py
```diff
@@ -401,9 +401,12 @@ def copy(self, tbl, table_name, if_exists='fail', max_errors=0, distkey=None,
...
             compupdate: boolean
-                Controls whether compression encodings are automatically applied during a COPY.
+                Controls whether compression encodings are automatically applied during a COPY. If
+                ``True`` explicitly sets ``compupdate`` to on, if ``False`` explicitly sets
+                ``compupdate`` to off. Defaults to ``None``.
```
jburchard, line 409: Perhaps explain to folks what happens if it is set to `None`. Or at least link to the redshift documentation?

parsons/databases/redshift/redshift.py
jburchard: Perhaps explain to folks what happens if it is set to None. Or at least link to the redshift documentation?

jburchard (changes requested):
Just a couple of requests around documentation?

SorenSpicknall (approved):
This looks good to go now after the documentation additions.

schuyler1d:
Sorry I missed this prompt -- I saw a new github message after this was merged.
I noted observed compupdate performance here:
https://github.com/dannyboy15/parsons/blob/332c3a8ab82836d24fa5419c69b2126f55340ca7/parsons/databases/redshift/redshift.py#L775-L780
In this case, it looks like @dannyboy15 is helping to fix a bad past default in Parsons for everyone where I merely changed it for the upsert() use-case.  So I retroactively support this PR :-)

I do think Parsons should include best-practices rather than just API defaults -- an API is only as good as its implementation, and to make parsons performant we should help users around performance cliffs when we can (and often API implementations from vendors are dumb -- defaults aren't policy, but sometimes defaults for their own legacy/backwards-compatibility requirements)

In terms of precedent, there are a few optimizations in the action_kit driver (upload_table) depending on the shape of the data uploaded.

PR #373, opened 2020-09-09
Redshift: Add dist/sort keys to upsert.

jburchard:
This PR does the following:

- Adds arguments that allow the user to set the dist/sort keys for an upsert. If the table does not previously exist, this will set the dist/sort keys. If the table exists, it will set the dist/sort keys for the staging table to the provided arguments.

- If the dist/sort key areguments are `None`, then it will set the primary_key as the dist/sort keys for the staging table.

The benefit of all of this should be that upserts should run faster whether the argument is passed or not, since the staging table will have a dist/sort key which will most likely be the same as the destination table.

### DIFF
--- a/parsons/databases/redshift/redshift.py
+++ b/parsons/databases/redshift/redshift.py
@@ -690,9 +690,8 @@ class Redshift(RedshiftCreateTable, RedshiftCopyTable, RedshiftTableUtilities, R
     def upsert(self, table_obj, target_table, primary_key, vacuum=True, distinct_check=True,
                cleanup_temp_table=True, **copy_args):
         """
-        Preform an upsert on an existing table. An upsert is a function in which records
-        in a table are updated and inserted at the same time. Unlike other SQL databases,
-        it does not exist natively in Redshift.
+        Preform an upsert on an existing table. An upsert is a function in which rows
+        in a table are updated and inserted at the same time.
 
         `Args:`
             table_obj: obj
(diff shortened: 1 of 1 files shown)


parsons/databases/redshift/redshift.py
```diff
@@ -762,6 +770,8 @@ def upsert(self, table_obj, target_table, primary_key, vacuum=True, distinct_che
                     copy_args = dict(copy_args, compupdate=False)
                 self.copy(table_obj, staging_tbl,
                           template_table=target_table,
+                          distkey=primary_key,
```
eliotst, line 773: `primary_key` can be a list, but Redshift only allows for a single `distkey`. You could just pick one if there's more than one.. so `primary_keys[0]` (`primary_keys` is generated in line 738 and will always be a list)

parsons/databases/redshift/redshift.py
```diff
@@ -762,6 +770,8 @@ def upsert(self, table_obj, target_table, primary_key, vacuum=True, distinct_che
...
+                          sortkey=primary_key,
```
eliotst, line 774: Redshift allows multiple `sortkey`s, but the `RedshiftCreateTable` doesn't.. could you change line 206/207 in `parsons/databases/redshift/rs_create_table.py` to support a possible list while we're here?

eliotst (changes requested):
Looks good.. just a couple comments about handling multiple primary keys.

ChrisC:
Updated in #588

PR #374, opened 2020-09-09
Docs: Update Zoom

rgriff23:
This PR adds an Authentication Guide and improves the Overview and Quick Start for the Zoom connector, checking the boxes for Zoom in #269. 

In addition, I made the following changes after reviewing the docstrings and functions in the Zoom class:

- Corrected the return value description in the docstring for `get_past_meeting` from `dict` to `Table`
- Prepended an underscore to the private `get_request` method, following the convention used in most other connectors

### DIFF
--- a/docs/zoom.rst
+++ b/docs/zoom.rst
@@ -5,24 +5,40 @@ Zoom
 Overview
 ********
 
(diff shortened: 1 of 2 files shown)


jburchard (approved):
Awesome. Thank you.

PR #388, opened 2020-09-12
VAN: Bulk Upsert Contact

jburchard:
There are a few more follow up PRs that I'd like to do, but I think that this is enough for now.

- Adds some fuzziness to the `map_columns()` utility that ignores spaces, underscores and capitalization.
- Add `VAN.get_bulk_import_results()` method.
- Add `VAN.get_bulk_import_mapping_type_fields()` method.
- Add `VAN.bulk_upsert_concacts()` method.
- Adds an unused at the moment method to automatically discover mapping types. This is not being used at the moment, but I hope to in a future PR.


parsons/ngpvan/bulk_import.py
```diff
@@ -46,6 +46,29 @@ def get_bulk_import_job(self, job_id):
         logger.info(f'Found bulk import job {job_id}.')
         return r
 
+    def get_bulk_import_job_results(self, job_id):
```
eliotst, line 49: It might be nice to provide some polling functionality in addition to the quick check here. Could you add another method to the class that polls this method every X seconds and returns when it gets a non-`None` result (or after a specified timeout)?

What I did in Rock the Vote was have three methods when downloading data.. one that kicks off the export (and returns the job ID), one that will check for results, and one that wraps both by kicking off the job and then polling the results until the data was available to download.

It's a non-trivial amount of work (and so might be outside the scope of this PR), but what if the import jobs returned an object that had methods for getting results (rather than the job ID).. so we could have a `ImportResult` that has `get_results` and `wait_on_results` (or some such). It adds another object, which isn't necessarily Parson-ic, but it might make for a more intuitive API than having to call back into the Connector with the `job_id`..?

parsons/ngpvan/bulk_import.py
```diff
@@ -153,3 +197,165 @@ def bulk_apply_activist_codes(self, tbl, url_type, **url_kwargs):
...
                                      'Activist Code Upload',
                                      **url_kwargs)
+
+    def bulk_upsert_contacts(self, tbl, url_type, result_fields=None, **url_kwargs):
+        """
+        Bulk create or update contact records. Provide a Parsons table of contact data to
+        create or update records.
+
+        .. note::
+            * The first column of the table must be VANID.
```
eliotst, line 207: In the name of making this more user-friendly, could we move the VAN ID column to the beginning of the `Table` for the user?

eliotst (changes requested):
This is exciting! I just had a few comments / questions.

parsons/ngpvan/bulk_import.py
```diff
@@ -46,6 +46,29 @@ def get_bulk_import_job(self, job_id):
...
+    def get_bulk_import_job_results(self, job_id):
+        """
+        Get result file of a bulk upload job. This will include one row
+        per record processed as well as the status of each.
+
+        If the job results have not been generated, either due to an error in the
+        process or the fact the job is still processing, it will return ``None``.
+
+        `Args:`
+            job_id : int
```
coreyhaines, line 58:
```suggestion
            job_id: int
```

ChrisC (approved):

parsons/ngpvan/bulk_import.py
ChrisC: Will file this as a separate issue

PR #392, opened 2020-09-13
Docs: Update Azure Blob Storage

rgriff23:
- add Authentication Guide, improve Overview, and improve Quick Start 
- should check the boxes for Azure: Blob Storage in #269

**Question about the class:** 

@eliotst 

I believe the `account_domain` argument / env variable is unnecessary in `AzureBlogStorage` since it will always be `blob.core.windows.net` (this is why I don't mention it in the Authentication Guide or Quickstart). The argument currently defaults to  `blob.core.windows.net` and it doesn't cause any issues AFAIK; it just looks like clutter in the `__init__`. 

I haven't changed the class yet since I'm not 100% sure there isn't a good reason for keeping the argument, but if someone can confirm that `account_domain` should be a constant in the class, I'm happy to make the change in this PR.

### DIFF
--- a/docs/azure.rst
+++ b/docs/azure.rst
@@ -8,43 +8,50 @@ Blob Storage
 ========
 Overview
 ========
-Azure Blob Storage is a cloud file storage system. It uses storage accounts to organize containers (similar to
-"buckets" for other storage providers) in which to store arbitrary files referred to as blobs.
 
-This connector currently only implements block blobs and not page or append blobs.
+`Azure Blob Storage <https://azure.microsoft.com/en-us/services/storage/blobs/>`_ is a cloud file storage system that
+uses storage accounts to organize containers (similar to "buckets" for other storage providers) in which to store
+arbitrary files referred to as 'blobs'. This Parsons integration currently only implements
+`block blobs, not page or append blobs <https://docs.microsoft.com/en-us/rest/api/storageservices/understanding-block-blobs--append-blobs--and-page-blobs>`_.
 
-You'll need credentials for an Azure Blob Storage storage account to use this connector. The ``azure-storage-blob``
-library is used for this connector, and `examples of how to create and use multiple types of credentials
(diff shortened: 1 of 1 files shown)


docs/azure.rst
```diff
@@ -8,43 +8,50 @@ Blob Storage
...
 .. code-block:: python
+   from parsons import AzureBlobStorage
 
-  from parsons import AzureBlobStorage
+   # First approach: Use API credentials via environmental variables
+   azure_blob = AzureBlobStorage()
 
-  azure_blob = AzureBlobStorage()
+   # Second approach: Pass API credentials as arguments
+   azure_blob = AzureBlobStorage(account_name='my_account_name', credential='1234')
```
jburchard, line 41: Is this missing the `account_url`?

jburchard (changes requested):
This is great. Just a minor comment.

In one of the examples that you didn't touch, I had a couple of suggestions:

```
from parsons import AzureBlobStorage

azure_blob = AzureBlobStorage()

container_name = 'testcontainer'

# Download to a temporary file path
temp_file_path = azure_blob.download_blob(container_name, 'test.csv')

# Download to a specific file path
azure_blob.download_blob(container_name, 'test.csv', local_path='/tmp/test.csv')
```

It seems like the container name is just a string which is the blob name. Perhaps just get rid of the declaration of the `container_name` variable. Pass the string directly into the `azure_blob.download_blob` method call? Also, call is `blob_name` rather than `container_name`?

Finally, in keeping with the other examples, remove the code (68-70) that imports and instantiates the class?

docs/azure.rst
rgriff23: I don't think so, the class requires *either* `account_name` or `account_url`, since `account_url` is ultimately `f"https://{account_name}.blob.core.windows.net/"`

rgriff23:
@jburchard thanks for the review, I addressed your comments in my last commit, so I think this should be good to merge

jburchard (approved):

PR #393, opened 2020-09-15
Docs: Update Facebook Ads

rgriff23:
- Add Quick Start, add Authentication Guide, improve Overview
- Move some material from the class docstring to the RST Overview
- Should check the boxes for Facebook Ads in #269


eliotst (approved):
This is great. Facebook has a lot of configuration pieces, so it's good to have it all written down like this. Thanks!

PR #394, opened 2020-09-15
Docs: Update Mobilize America

rgriff23:
- Add Quickstart, improve Overview and Authentication Guide 
- Should check the boxes for Mobilize America in #269

Also, I made some minor edits to the class:
- Add underscore prefix to the private class methods `request` and `request_paginate`, so they are now `_request` and `_request_paginate`)
- Remove some unused function arguments


eliotst (approved):
Thanks!

PR #398, opened 2020-09-15
Docs: Update SFTP

rgriff23:
- Add Authentication Guide and Quickstart, improve Overview
- Move example code from the `create_connection` docstring to the Quickstart
- Should check the boxes for SFTP in #269

### DIFF
--- a/docs/sftp.rst
+++ b/docs/sftp.rst
@@ -1,10 +1,51 @@
 SFTP
 ====
 
-The SFTP class allows you to interact with SFTP services.
+The ``SFTP`` class allows you to interact with `SFTP services<https://en.wikipedia.org/wiki/SSH_File_Transfer_Protocol>`_,
+using the  `Paramiko SFTP library <http://docs.paramiko.org/en/2.4/api/sftp.html>`_ under the hood.
 
-It uses the `Paramiko SFTP library <http://docs.paramiko.org/en/2.4/api/sftp.html>`_ under the hood.
+The class provides methods to:
 
+- Create SFTP connections
+- Make, remove, and list the contents of directories
+- Get, put, remove, and check the size of files
+
+.. note::
+  Authentication
+    Depending on the server provider, SFTP may require either password or public key authentication.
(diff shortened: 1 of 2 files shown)


docs/sftp.rst
```diff
@@ -1,10 +1,51 @@
 SFTP
 ====
 
-The SFTP class allows you to interact with SFTP services.
+The ``SFTP`` class allows you to interact with `SFTP services<https://en.wikipedia.org/wiki/SSH_File_Transfer_Protocol>`_,
+using the  `Paramiko SFTP library <http://docs.paramiko.org/en/2.4/api/sftp.html>`_ under the hood.
```
shaunagm, line 5: It looks like we're actually using version 2.7.1 of Paramiko (http://docs.paramiko.org/en/2.7/api/sftp.html), can you update the link?

docs/sftp.rst
```diff
@@ -1,10 +1,51 @@
...
+.. note::
+  Authentication
+    Depending on the server provider, SFTP may require either password or public key authentication.
+    The ``SFTP`` class supports both methods via ``password`` and ``rsa_private_key_file`` arguments.
+
+**********
+Quickstart
+**********
+
+To instantiate ``SFTP``, pass your host name, user name, and either a password or an authentication
```
shaunagm, line 22: I was going to tell you "remind people that they can use environmental variables!" but it looks like they can't with the SFTP connector.  @eliotst is there a reason for that, or should we make a new ticket to add support for environmental variable config in the SFTP connector?

shaunagm (changes requested):
Looks good to me, let's just get that link to the paramiko docs updated.

docs/sftp.rst
jburchard: There is not a reason for it. It was just forgotten when we moved to the envs + args model. Most everything else was upgraded. I think that an issue for this would be appropriate.

rgriff23: FYI `Bill.com` also lacks support for environmental variables

shaunagm (approved):
Sorry for the delay, I thought this PR was still waiting for changes.

PR #402, opened 2020-09-17
redshift: move distkey/sortkey warning

eliotst:
This commit moves the warning for missing distkey / sortkeys when
copying data into Redshift to only fire when a new table is being
created, instead of whenever `copy` (or `upsert`) is called.

cc @tiburona


jburchard (approved):
Awesome!

PR #407, opened 2020-09-20
Docs: Update Civis

rgriff23:
This should check the boxes for Civis in #269:

- Add Authentication Guide, Overview, and Quickstart

Additional changes:

- Remove some unused function arguments
- Align order of args in the function definition, docstring, and usage in the init

### DIFF
--- a/parsons/civis/civisclient.py
+++ b/parsons/civis/civisclient.py
@@ -44,10 +44,6 @@ class CivisClient(object):
             preview_rows: int, optional
                 The maximum number of rows to return. No more than 100 rows can be
                 returned at once.
-            polling_interval: int or float, optional
-                Number of seconds to wait between checks for query completion.
-            hidden: bool, optional
-                If ``True`` (the default), this job will not appear in the Civis UI.
             wait: boolean
                 If ``True``, will wait for query to finish executing before exiting
                 the method.
--- a/parsons/civis/civisclient.py
+++ b/parsons/civis/civisclient.py
@@ -8,11 +8,11 @@ class CivisClient(object):
     Instantiate the Civis class.
 
     `Args:`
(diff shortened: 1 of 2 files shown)


jburchard (changes requested):
Do you mind removing hidden and hidden from the doc strings?

parsons/civis/civisclient.py
```diff
@@ -21,17 +21,16 @@ class CivisClient(object):
...
         self.client = civis.APIClient(api_key=api_key, **kwargs)
         """
         The Civis API client. Utilize this attribute to access to lower level and more
         advanced methods which might not be surfaced in Parsons. A list of the methods
         can be found by reading the Civis API client `documentation <https://civis-python.readthedocs.io/en/stable/client.html>`_.
         """  # noqa: E501
 
-    def query(self, sql, preview_rows=10, polling_interval=None, hidden=True,
-              wait=True):
+    def query(self, sql, preview_rows=10, wait=True):
```
eliotst, line 33: Good catch on the unused variables. It looks like they are valid arguments to the `query_civis` method on the Civis client (https://civis-python.readthedocs.io/en/stable/generated/civis.io.query_civis.html#civis.io.query_civis). Instead of removing them from the Parsons method, could you update the call to `query_civis` to pass them through?

So, on line 60, something like:
```
fut = civis.io.query_civis(sql, self.db, preview_rows=preview_rows, polling_interval=polling_inverval, hidden=hidden)
```

eliotst (changes requested):
This is great. I just had the one comment re: unused arguments.

Thanks for all of the work you are putting in on Parsons UI!

eliotst (approved):
Thanks!

PR #417, opened 2020-09-25
Issue 269

davidpablocohn:
Filled out the google.rst doc with quickstarts and enhanced overviews.


rgriff23:
One little thing, can we add an ‘Overview’ header for all the classes, for consistency? It looks like Sheets and Cloud Storage have it, but Civis and BigQuery don’t.

davidpablocohn:
If I'm reading things correctly,  Randi's observation is because there were obsolete separate google_sheets.html and google_civic.html files and, for reasons I don't understand, the docs/html files are .gitignored, so the new google.html wasn't getting pushed. I've force-added it, and I *think* everything now looks right here, with overviews and all: https://github.com/davidpablocohn/parsons/blob/issue_269/docs/google.rst

rgriff23:
@davidpablocohn actually it was a totally silly mistake on my part, I read over the RST doc on my phone and the view was just clipping out the: 

```
=======
Overview
=======
```

So... everything was fine. :)

eliotst:
One thing I learned a good 6+ months into working the project: the HTML files checked into Github are not the ones served up on the documentation website. The deploy to the doc site re-builds the HTML files from whatever RST files are in Github.

eliotst:
Thanks for this @davidpablocohn ! I should be able to review this soon! I'll have comments to you in a couple of days.

davidpablocohn:
Ah - thanks!

docs/google.rst
```diff
@@ -17,12 +19,74 @@ Google BigQuery is a cloud data warehouse solution. Data is stored in tables, an
 BigQuery uses datasets as top level containers for tables, and datasets are themselves contained within
 Google Cloud projects.
 
+==========
+Quickstart
+==========
+
+To instantiate the GoogleBigQuery class, you can pass the constructor a string containing either the name of the Google Drive credentials file or a JSON string encoding those credentials. Alternatively, you can set the environment variable ``GOOGLE_APPLICATION_CREDENTIALS`` to be either of those strings and call the constructor without that argument.
```
eliotst, line 26: I think "Google service account credentials file" is better than "Google Drive credentials file".

docs/google.rst
```diff
@@ -31,14 +95,52 @@ Cloud Storage
 Overview
 ========
 Google Cloud Storage is a cloud file storage system. It uses buckets in which to
-store arbitrary files referred to as blobs.
+store arbitrary files referred to as blobs. You may use this connector to upload Parsons tables as blobs, download them to files, and list available blobs.
+
+To use the GoogleCloudStorage class, you will need Google Drive credentials. If you are the administrator of your Google Cloud account, you can generate them in the `Google Cloud Console APIs and Services <https://console.cloud.google.com/apis/credentials/serviceaccountkey?_ga=2.116342342.-1334320118.1565013288>`_.
```
eliotst, line 100: Same as above.. for Cloud Storage and BigQuery, it's not "Google Drive credentials".. I think "Google service account credentials" is more accurate.

docs/google.rst
```diff
@@ -31,14 +95,52 @@ Cloud Storage
...
+
+==========
+Quickstart
+==========
+
+To instantiate the GoogleBigQuery class, you can pass the constructor a string containing either the name of your Google Drive credentials file or a JSON string encoding those credentials. Alternatively, you can set the environment variable ``GOOGLE_APPLICATION_CREDENTIALS`` to be either of those strings and call the constructor without that argument.
```
eliotst, line 106: Again, "Google service account credentials" instead of "Google Drive credentials".

eliotst (changes requested):
Thank you so much. The code examples here are really helpful, and it's great to have more info on credentials for Google. I had a few comments about terminology, but overall this looks great!

docs/google.rst
davidpablocohn: Good catch! Changed and pushed

docs/google.rst
davidpablocohn: Fixed! Good catch.

eliotst (approved):
Thanks so much!

PR #422, opened 2020-09-27
docs: unit test doc

eliotst:
This commit adds a HOW TO doc for writing unit tests. The doc
is meant to give an overview of writing unit tests to
contributors.


rgriff23 (commented):
I think this looks great. One thought I have is that it might be helpful to mention the utility function `test.utils.assert_matching_tables` so that people know it exists and are aware that:

```
from parsons import Table
from test.utils import assert_matching_tables

a = Table() 
b = Table() 

# This fails because it actually tests whether a and b are the same instance
assert(a == b)

# But this works
assert(list(a) == list(b))

# And this works
assert_matching_tables(a, b)
``` 

Maybe more experienced Python users would not be too surprised by this, but it was definitely a sticking point for me on my first connector.

docs/build_a_connector.rst
```diff
@@ -242,34 +242,34 @@ Adding automated tests
  * Add one `“Happy Path” <https://en.wikipedia.org/wiki/Happy_path>`_ test per public method of your connector
  * When possible mock out any external integrations, otherwise mark your test using the
```
shaunagm, line 243: Using the what?

shaunagm, line 243: Using the what?

docs/write_tests.rst
```diff
@@ -0,0 +1,131 @@
+How to Write Tests for Parsons Connectors
```
shaunagm, line 1: Is this doc linked from How to Build a Connector?  We should probably stick a quick link in the 'adding tests' section.

docs/write_tests.rst
```diff
@@ -0,0 +1,131 @@
...
+******************
+How to Write Tests
+******************
+
+Most of our Parsons Connector classes fall into two categories:
+
+ * Classes that call directly into HTTP API's using the Parsons ``APIConnector`` class
+ * Classes that wrap around third party Python libraries, often using their own "client" class
+
+The category that the class falls into generally determines how one will write unit tests for the class.  However, the category of Connector determines how we simulate the external API.
```
shaunagm, line 13: I'm having trouble parsing this, specifically that second sentence. What are you trying to say here?

docs/write_tests.rst
```diff
@@ -0,0 +1,131 @@
...
+
+For classes built using the ``APIConnector`` class, we can leverage the ``requests-mock`` library, since the ``APIConnector`` uses the ``requests`` library.
+
+The ``requests-mock`` library allows us to simulate the API we are calling into by pretending to be an HTTP server. We can create canned responses, and introspect on the endpoints that our Connector called to ensure that the method we are testing is acting the way we expect.
+
+For an example of testing a class built on the ``APIConnector``, we can look at the ``MailChimp`` connector.
+
+In the code below, we have our ``TestMailchimp`` that serves as our test case. We have one test method to test calls to our ``get_campaigns`` method in our ``Mailchimp`` Connector class.
+
+We decorate ``test_get_campaigns`` with the `requests_mock.Mocker`` class, which passes in our requests mocker as an argument ``m``. We can then use this mock to configure our mock server.
```
shaunagm, line 37: Missing a tick around requests_mock.Mocker

shaunagm (commented):
Looks fantastic overall, just a few minor issues/requests for clarification.

docs/write_tests.rst
```diff
@@ -0,0 +1,131 @@
...
+            mc = MailChimp(api_key='12345')
+
+            # Test that campaigns are returned correctly.
+            m.get(self.mc.uri + 'campaigns', json=expected_json.test_campaigns)
+            tbl = self.mc.get_campaigns()
+
+            self.assertEqual(tbl.num_rows, 2)
+
+
+After wiring up the mock HTTP API, we are ready to call the ``get_campaigns`` method on our ``Mailchimp`` connector. Since we are using the ``requests_mock.Mocker`` class, our Connector will not actually hit the Mailchimp API; our call will be intercepted and passed to our mock HTTP server.
```
coreyhaines, line 59: It might be a bit confusing to say "our mock HTTP server" as it isn't really a mock http server, but a canned response. Perhaps something like
```suggestion
After wiring up the mock HTTP API, we are ready to call the ``get_campaigns`` method on our ``Mailchimp`` connector. Since we are using the ``requests_mock.Mocker`` class, our Connector will not actually hit the Mailchimp API; our call will be intercepted and the configured canned response will be returned.
```

docs/write_tests.rst
```diff
@@ -0,0 +1,131 @@
...
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Tests for Connectors Built on Third Party Libraries
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+For classes that wrap third party libraries, we are looking to simulate whatever client object that serves as the Python API for the calls to the external service. To simulate the client object, we can use the ``unittest.mock.MagicMock`` class from the Python standard library.
+
+The ``MagicMock`` class allows us to build a fake version of the client object by simulating functionality we expect our Connector class to use -- namely by returning canned responses to method calls on the client. As well, the ``MagicMock`` class is recording those calls as they happen, so that after we have run through our tests, we can check that things happened in the way we expected.
+
+The following gives a simple example of initializing, setting up, and using a ``MagicMock`` class:
```
coreyhaines, line 69: This might lead to some confusion about "using" the mock, since the example code here is focused on setting up and checking expectations. Calling the mock directly like this is just for illustrative value.

```suggestion
The following gives a simple example of initializing, setting up, and checking expectations on a ``MagicMock`` class. Please see the following, SalesforceTest example on integrating it with your Connector class.
```

docs/write_tests.rst
```diff
@@ -0,0 +1,131 @@
...
+    # Tell the client to return a canned list of people from its list_people method
+    mock_client.list_people.return_value = [
+        {'id': 1, 'name': 'Nicole Jackson'},
+    ]
+
+    # Call our mock client's list_people method
+    people = mock_client.list_people()
+    # Check that we got our expected data
+    assert len(people) == 1
+    assert people[0]['id'] == 1
```
coreyhaines, line 86: I would put some spacing here to highlight that this moves into the assertions.
```suggestion


    # Check that we got our expected data
    assert len(people) == 1
    assert people[0]['id'] == 1
```

docs/write_tests.rst
```diff
@@ -0,0 +1,131 @@
...
+
+    # Initialize a MagicMock object to serve as our fake client
+    mock_client = MagicMock()
+    # Tell the client to return a canned list of people from its list_people method
+    mock_client.list_people.return_value = [
+        {'id': 1, 'name': 'Nicole Jackson'},
+    ]
+
+    # Call our mock client's list_people method
+    people = mock_client.list_people()
```
coreyhaines, line 83: Emphasize that this is where your Connector would call it.
```suggestion
    # Your Connector would call the client like normal
    people = mock_client.list_people()
```

shaunagm (commented):
This largely looks good, except I'm not clear on where the configured data for the Connectors that connect directly to HTTP APIs live. It's referenced by `expected_json.test_campaigns` correct?  But where is that being imported from? From looking at the repo, it should be in an `expected_json.py` file at the same level as the test.  Can you put instructions for this into the docs?

PR #423, opened 2020-09-30
New connector: Sisense / Periscope 

### LINKED ISSUE #270: Periscope Connector
https://developer.sisense.com/display/API2/REST+API+Reference+-+v1.0

There are some key endpoints that would be super helpful:

[Dashboards](https://developer.sisense.com/display/API2/REST+API+Reference+-+v1.0#RESTAPIReference-v1.0-Dashboards)

MVP:
- get/dashboards
- get/dashboards/{id}
V2 & beyond:
- get/dashboards/{dashboardId}/shares
- get/dashboards/{dashboardId}/shares/{id}
- post/dashboards/{id}/publish

For the publish piece, I want to make sure you can set filter values:
https://dtdocs.sisense.com/article/embed-api#PublishDashboard

rgriff23:
This PR adds an MVP Sisense/Periscope connector and should close #270. 

Summary of class methods:

- `get_dashboards` fetches a specific dashboard by ID or set of dashboards, depending on whether the argument `dashboard_id` is passed
- `get_dashboard_shares` fetches shares for a specific dashboard by ID or set of dashboards, depending on whether the argument `share_id` is passed
- `publish_dashboard` publishes a dashboard 

I don't have credentials so I have not tested the methods against the API.


eliotst:
Thanks @rgriff23 . I tested the code using an API key for our Periscope instance, and I reviewed the documentation above .. I think the endpoints you are calling for `get_dashboards` and `get_dashboard_shares` are actually for a different product. I think we thought they were for our Periscope dashboards ("Sisense for Cloud Data Teams"), but I think they are for a different Sisense dashboarding product. Sorry for the confusion!

The endpoint for publishing dashboards is still valid, so the `publish_dashboard` endpoint is still useful. I am reviewing the code and testing that more extensively. I should have comments in the next day or two.

parsons/sisense/sisense.py
```diff
@@ -0,0 +1,106 @@
...
+        `Returns:`
+            A Table or an error.
+        """
+        if dashboard_id is None:
+            response = self.api.get_request('dashboards', params=params)
+        else:
+            response = list(self.api.get_request(f'dashboards/{dashboard_id}', params=params))
+        return Table(response)
+
+    def get_dashboard_shares(self, dashboard_id, share_id=None, params=None):
```
eliotst, line 76: So, it looks like there is an endpoint to list all shares for a given dashboard ID even in the Sisense for Cloud Teams API: https://dtdocs.sisense.com/article/embed-api-options#List

It looks like the API URL is https://app.periscopedata.com/api/v1/shared_dashboard/list

parsons/sisense/sisense.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        This method publishes a dashboard based on a URL encoded JSON blob. For available options, see the `API documentation <https://dtdocs.sisense.com/article/embed-api-options>`_. # noqa
+
+        `Args:`
+            payload: dict
+                A JSON blob defining the dashboard to publish.
+        `Returns:`
+            Response or an error
+        """
+        return requests.post(self._create_dashboard_url(payload))
```
eliotst, line 106: I don't think you need this `post`.. the `_create_dashboard_url` method does the work of creating the dashboard, and returns the URL that I think the user will want.

eliotst (changes requested):
Thanks for this. Sorry again for the confusion. It turns out there are two additional endpoints for the Sisense for Cloud Data Teams API.. List (https://app.periscopedata.com/api/v1/shared_dashboard/list) and delete (https://dtdocs.sisense.com/article/embed-api-options#Delete)

They look fairly straightforward, so if you think you can add them here, that'd be great.

Otherwise, I had a few comments about how the current method works. I was able to use it to create a new share, but there is an additional `post` call in the `publish_dashboard` method that I think is unnecessary.

rgriff23:
@eliotst ready for another review. Some notes:

I got rid of `_create_dashboard_url` to simplify things, so all three functions (create, list, and delete shared dashboards) now use a similar structure.  

I wasn't sure what the expected response is when deleting a dashboard, so I went with `{'success': True}`.

eliotst (approved):
This looks great. Thanks!

PR #424, opened 2020-09-30
Clean up docs args

rgriff23:
Minor fixes for #269:

- Bloomerang: rm some extraneous backticks from args in docstrings
- Twilio: escape hyperlink formatting using backslash in the `from_` arg

### DIFF
--- a/parsons/bloomerang/bloomerang.py
+++ b/parsons/bloomerang/bloomerang.py
@@ -99,7 +99,7 @@ class Bloomerang(object):
     def create_constituent(self, **kwargs):
         """
         `Args:`
-            **kwargs:`
+            **kwargs:
                 Fields to include, e.g., FirstName = 'Rachel'.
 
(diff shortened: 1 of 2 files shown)


rgriff23:
@shaunagm hmmm I thought this backslash would work to escape the hyperlink formatting based on reading this:

https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#escaping-mechanism 

I also got an answer on Stack Overflow which told me it would work.

shaunagm:
Looks good to me:

twilio

FYI if you're having trouble building the docs yourself, there are some proto-instructions [here](https://github.com/move-coop/parsons/issues/425).  (I'll be tackling the issue itself tomorrow with a new Parsons contributor.)

shaunagm (approved):

shaunagm:
Oh it looks like your change causes a linting error:

Screenshot from 2020-10-01 10-00-06

My inclination is to [ignore the linting error](https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html) but we can see what Eliot says.

PR #427, opened 2020-10-02
Add VAN script for random sample of a saved list

pjsier:
Creates a template script based on some of the conversations in Slack that demonstrates creating a new saved list in VAN based on a random sample of an existing saved list.

This is my first sample script, so I wanted to try something relatively simple to make sure I've got all the right pieces together and that I'm using the VAN API correctly. Let me know what I can change, thanks!


eliotst:
Thanks for this, @pjsier ! @ydamit is going to take a look.

useful_resources/sample_code/ngpvan_sample_list.py
```diff
@@ -0,0 +1,73 @@
+# ### METADATA
+
+# Connectors: VAN
+# Description: Creates a new saved list from a random sample of an existing saved list in VAN
+
+# ### CONFIGURATION
+
+# Set the configuration variables below or set environmental variables of the same name and leave
+# these with empty strings.  We recommend using environmental variables if possible.
```
ydamit, line 9: Not sure we should change anything here, but worth noting that items that are likely to change often (as opposed to those that are evergreen), such as `VAN_FOLDER_ID`, `VAN_SAVED_LIST_ID`, etc. are more annoying to move in and out of env vars.

useful_resources/sample_code/ngpvan_sample_list.py
```diff
@@ -0,0 +1,73 @@
...
+    "S3_BUCKET": "",  # Bucket that the sample list data should be uploaded to
+}
+
+
+# ### CODE
+
+from parsons import VAN  # noqa: E402
+from parsons import logger  # noqa: E402
+import os  # noqa: E402
+import random  # noqa: E402
```
ydamit, line 31: I've been told that convention is to put all the simple `import`s at the top, and more complicated `from`s at the bottom. That's a bit more nitpicky than I like to be (not to mention I don't even fully understand the reasoning behind it), but as a sample script, I figure we should model the convention.

useful_resources/sample_code/ngpvan_sample_list.py
```diff
@@ -0,0 +1,73 @@
...
+    if value.strip() != "":
+        os.environ[name] = value
+
+van = VAN(db=os.environ["VAN_DB_NAME"])
+
+# Get details on the saved list we're pulling from and the folder we're uploading to
+saved_list = van.get_saved_list(os.environ["VAN_SAVED_LIST_ID"])
+folder = van.get_folder(os.environ["VAN_FOLDER_ID"])
+
+saved_list_download = van.download_saved_list(os.environ["VAN_SAVED_LIST_ID"])
```
ydamit, line 46: A note here that using this method requires certain non-default permissions be turned on for an API key. I _believe_ all Parsons type VAN API keys have this turned on, but may be worth a comment to reach out to apidevelopers@ngpvan.com if struggling to get this to work.

useful_resources/sample_code/ngpvan_sample_list.py
```diff
@@ -0,0 +1,73 @@
...
+
+# Generate a random sample of VAN IDs from the list
+saved_list_sample_ids = random.sample(
+    saved_list_download["VanID"],
```
ydamit, line 50: I'm not 100% positive if this would be true on the results of a saved list download, but I believe VAN defaults to camelCase.

useful_resources/sample_code/ngpvan_sample_list.py
```diff
@@ -0,0 +1,73 @@
...
+    int(os.environ["VAN_SAMPLE_LIST_SIZE"]),
+)
+
+# Create a new sample list from the sampled IDs that only includes the VanID column
+saved_list_sample = saved_list_download.cut("VanID").select_rows(
+    lambda row: row["VanID"] in saved_list_sample_ids
```
ydamit, line 56: Again, I think column names will be camelCased.

ydamit (changes requested):
This is great stuff! All my requested changes are, I think, very minor.

Thanks for your contribution!

useful_resources/sample_code/ngpvan_sample_list.py
shaunagm: This is a very valid point, and something I ran into when writing my own script today. What @pjsier has done here is in line with the templates I made, but I think you're right that we shouldn't be storing *all* the configuration variables as environmental variables, just the authentication ones.  If we make this change, we should do it for all scripts though. So maybe keep this as is, and we can create a separate issue to switch all the scripts + the template at once?

useful_resources/sample_code/ngpvan_sample_list.py
pjsier: Agreed with you here as well as the above point, but I was going off of another script `apply_activist_code.py` so not sure if this is something we should apply across all scripts first?

useful_resources/sample_code/ngpvan_sample_list.py
pjsier: This worked when I ran it against VAN in our test instance so not sure if there's an issue with different versions?

ydamit: I don't think there should be issues with versions. If this casing worked, then we should probably ignore any of my comments on casing!

shaunagm: Is the VAN connector, or Parsons in general, case sensitive when it comes to column names? I was trying to figure this out yesterday when working with `map_and_coalesce_columns` and ended up guessing it was case-insensitive. Whichever it is, we should say so prominently in the documentation. @eliotst?

useful_resources/sample_code/ngpvan_sample_list.py
pjsier: Updated for now because it's inconsistent across scripts!

useful_resources/sample_code/ngpvan_sample_list.py
pjsier: Just added a comment in the latest commit

ydamit (approved):
Can't see any reason to hold this up any longer!

shaunagm (approved):

PR #428, opened 2020-10-04
Issue 217

davidpablocohn:
First pass implementation of Box connector.


eliotst:
@davidpablocohn This is awesome. Thanks for the contribution. I think this is going to be a really helpful tool for us and for others when we need to share files. We will take a look at this over the next couple of days.

In the meantime, could you clear out the generated doc files from the PR? The docs get auto-generated on the doc server at deploy time, so we don't need them updated in Github. Also, it adds a decent amount of noise to the PR diff. Thanks!

davidpablocohn:
Ah criminy - I forgot about those! Will do shortly.

eliotst:
Thanks!

docs/box.rst
```diff
@@ -0,0 +1,80 @@
...
+
+**********
+Quickstart
+**********
+
+This class requires credentials in the form three strings to be either passed into the constructor or made available as environment variables:
+
+.. code-block:: bash
+
+  export BOX_CLIENT_ID=dp4rqi0cz5qckz361fziavdtdwxz
```
shaunagm, line 22: Are these fake keys?

docs/box.rst
```diff
@@ -0,0 +1,80 @@
...
+**List folders and files**
+
+.. code-block:: python
+
+  # List default folder
+  folder_list = box.list_folders()
+
+  # Subfolders of folder id 533944
+  subfolder_list = box.list_folders(folder_id='533944')
+
```
shaunagm, line 50: The indentation looks slightly off here to me. Were you able to build the docs to check how it actually looks?

parsons/box/box.py
```diff
@@ -0,0 +1,200 @@
...
+        url = 'https://api.box.com/2.0/folders/' + folder_id
+        json_response = self.client.make_request('GET', url)
+
+        items = Table(json_response.json()['item_collection']['entries'])
+        if item_type:
+            items = items.select_rows(lambda row: row.type == item_type)
+        return items
+
+    # In what formats can we upload/save Tables to Box? For now, just csv
+    ALLOWED_FILE_FORMATS = ['csv', 'json']
```
shaunagm, line 133: Your comment says we can only save as csv but the allowed file formats includes json?

shaunagm (commented):
This looks like a really great start, @davidpablocohn! The only substantive comment I have so far is around increasing the test coverage, but I imagine you're waiting on that to get review on the code itself.

docs/box.rst
davidpablocohn: Yes, fake keys. Shall I add a note in the doc saying "These are fake - go generate your own"?

docs/box.rst
davidpablocohn: Good catch!

parsons/box/box.py
davidpablocohn: Again, good catch - I added JSON in a subsequent CL, and missed some of the doc updates

parsons/box/box.py
```diff
@@ -0,0 +1,200 @@
...
+from parsons.etl.table import Table
+from parsons.utilities.check_env import check as check_env
+from parsons.utilities.files import create_temp_file
+
+import tempfile
+
+logger = logging.getLogger(__name__)
+
+
+class Box(object):
```
coreyhaines, line 33: Notice that some of the functions have the default id of "0" in the signature, some don't. Can we make it consistent across all functions.

parsons/box/box.py
```diff
@@ -0,0 +1,200 @@
...
+    """Box is a file storage provider.
```
coreyhaines, line 34: In the docstring for some functions, we have a space after the `:` , such as
> client_id: str
and some don't have the space
> client_id:str

Also, some of the argument descriptions end in a `.` and some don't.

Would be good to have these inconsistencies addressed.

coreyhaines (changes requested):
Mostly consistency requests.

parsons/box/box.py
davidpablocohn: I think I've put in defaults (specifically, using the default folder) wherever possible. This makes sense for some methods, e.g. for list_files() and upload_table() - if they've not otherwise specified list files in the default folder, or upload to the default folder, unless otherwise specified.

With other things like delete_file(), it doesn't make sense to have a default file_id.

I've created a global `DEFAULT_FOLDER_ID = '0'` and swapped it in on the function signatures to clarify the semantics.

parsons/box/box.py
davidpablocohn: Good point - I got sloppy.

parsons/box/box.py
```diff
@@ -0,0 +1,200 @@
...
+            parent_folder: str
+               Folder id of the parent folder in which to create the new folder. If
+               omitted, the default folder will be used.
+
+        `Returns`:
+            str: The Box id of the newly-created folder
+        """
+        parent_folder = parent_folder or '0'
+        subfolder = self.client.folder(parent_folder).create_subfolder(folder_name)
+        return subfolder.id
```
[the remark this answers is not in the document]
davidpablocohn: Good point - fixed!

parsons/box/box.py
coreyhaines: Was looking at line 77, which sets the default at the signature to `None` then do a parent_folder_id = parent_folder_id or '0' inside the body.

davidpablocohn: Ah - yup! I'm sure I had a reason for doing that, but have no idea what it was, and replaced it with parent_folder_id=DEFAULT_FOLDER_ID in the signature.

coreyhaines (approved):
Looks good to me. Don't have a Box account to test, but the comments I had on the code are resolved. Would look to wait for someone to test it who has an account.

eliotst:
Hey @davidpablocohn .. I just tested the endpoints using my Box account, and it all works!

Before I start making line-level comments, I wanted to ask a high level question: do you think it is possible (or worth it) to have the Parsons methods talk about folders and files as paths instead of as file/folder IDs? So, instead of `create_folder('bar', parent_folder_id=12345)`, instead saying `create_folder('foo/bar')`.

Looking at the Box API, it seems like it might be non-trivial to implement. I suspect you would have to walk paths by making multiple API calls. One implementation of `list_files('foo/bar/baz')` would have to call `list_items` for `foo`, `bar`, and then `baz` to find the appropriate parent IDs along the way.. I'm not sure if there's a higher level endpoint on the API to leverage (the Box API is large and confusing). It looks like there's some search endpoints, but I can't tell if they support searching by path.

What you have here is great, and this would definitely more of a nice-to-have.. but I thought I'd run it by you before diving too far into the code.

davidpablocohn:
I did ponder that, and decided I might be overreaching my spec. If you
think that would be useful (more useful than, e.g. me starting to poke at
the Action Network stuff), I can definitely dive into it. I'm sure there
are gotchas, and yes, it would take a couple of API calls per higher level
method, but I don't think it would be all that tough. If we're going to
eventually go that route, though, I should prob rename the get_table()
method to get_table_by_id(). Maybe I should do that, anyway.

Regardless - let me know if you'd like me to implement paths instead of (or
in addition to) ids?

Thanks,
 -p

eliotst:
I think ID's are fine for now. I think it makes sense for you to move on to ActionNetwork and you or someone else can circle back to this later. There is precedence for going back to add the flexibility in the connector to handle either an ID or a name where we initially expected an ID (we did this with Google Sheets, etc). Do you think it would be worth it to rename some of the args from like `folder_id` to `folder` to future proof for that?

davidpablocohn:
Hmm - my previous comment/question seems to have evaporated with the conflict fix. Repeating here:

While waiting for Action Network access, I implemented get_item_id(path), which makes it straightforward to convert all the other methods to use paths rather than ids. Only problem is that it appears the only way to navigate a path is via one list_items() API call per path level, so with deep paths (or very large intermediate folders), get_item_id() could be slow.

Because of that, I'm hesitant to lock power users into using paths only. So I see a couple of alternatives:

1. Have separate path-based and id-based methods:
```
upload_table_to_folder_id(self, table, file_name, folder_id=DEFAULT_FOLDER_ID, format='csv')
upload_table(self, table, file_name, path, format='csv')

get_table_by_file_id(self, file_id, format='csv')
get_table(self, path, format='csv')
```

2. Have optional parameters
```
upload_table(self, table, file_name, path=None, folder_id=None, format='csv')
get_table(self, path=None, file_id=None, format='csv')
```
And specify that only one of path/file_id may be provided.

Thoughts? Preferences?

Thanks!
 -p

eliotst:
I think if you were to have an argument named `folder` and the method could suss out whether it was a path or an ID, that would be great. I guess I'm not sure: is the ID always an integer? If so, we could do a type check (and document it in the args) and say "if type(folder) == str` then it's a path, `elif type(folder) == int` then it's an ID. Thoughts?

For me, the optional parameters is tough because then we have to check every time that you provided one or the other but not both, which can be tedious. Having separate methods is definitely more clear, but it makes the API a little more clunky. If it's not easy to disambiguate between an ID and a path in a single argument, then my preference is probably separate methods.

It did look to me that the only way to support path would be repetitive calls to `list_items`, which is a bummer .. but I'm guessing it's pretty uncommon to have to navigate more than a few layers..? And in that case, users could fall back to providing the folder ID. Maybe we could note in the documentation of the class that deeply nested folder structures don't perform well when referencing folders/files by path..?

davidpablocohn:
Alas, it appears that Box IDs are integers encoded as strings. So there's
no reliable (or relatively foolproof) way of distinguishing. Too much
opportunity for minor user error to cause inscrutable confusion. I ended up
leaning toward separate methods, as well. Coding them up was almost trivial
- the docstring longer than the code, e.g.

    def upload_table(self, table, file_name, path='', format='csv') ->
boxsdk.object.file.File:

        """Save the passed table to Box.


        `Args`:

            table:Table

               The Parsons table to be saved.

            file_name: str

               The filename under which it should be saved in Box.

            path: str

               Optionally, file path of the folder in which it should be
saved.

            format: str

               For now, only 'csv' and 'json'; format in which to save
table.


        `Returns`: BoxFile

            A Box File object

        """

        if path:

            folder_id = self.get_item_id(path)

        else:

            folder_id = DEFAULT_FOLDER_ID

        return self.upload_table_to_folder_id(table=table,
file_name=file_name,

                                              folder_id=folder_id,
format=format)

I'll clean things up just a touch before adding them to the pull request.

shaunagm:
Out of curiosity, couldn't you distinguish between an ID and path by doing something like:

```
try:
  folder_id = int(folder)
except ValueError:
  folder_path = folder
```

davidpablocohn:
I started to go that route, but realized that sometimes (like me) people
give numeric folder or file names.

I think having separate methods, while a bit more verbose, is probably the
only foolproof approach. And codewise it's also the simplest.

I will go ahead and finish up coding that way to see whether I'm happy with
how it looks. Will keep you posted.

eliotst (approved):
This looks good to me! I appreciate your willingness to talk through things and rework the code. Thanks!

PR #431, opened 2020-10-06
Add zoom_to_van.py sample script

shaunagm:


useful_resources/sample_code/zoom_to_van.py
```diff
@@ -0,0 +1,71 @@
...
+# Coalesce the columns into something VAN expects
+column_map = {"first_name": ["fn", "first", "firstname", "first name"],
+              "last_name": ["ln", "last", "lastname", "last name"],
+              "date_of_birth": ["dob", "date of birth", "birthday"],
+              "email": ["email address", "email_address"],
+              "street_number": ["street number", "street no.", "street no", "street #"],
+              "street_name": ["street name", "street"],
+              "phone": ["phone_number", "phone #", "phone_#", "phone no.", "phone no"],
+              "zip": ["zip5", "zipcode", "zip code"]}
+filtered_participants.map_and_coalesce_columns(column_map)
```
eliotst, line 54: This is an interesting pattern.. I wonder if we should think about moving some of these into Parsons.. like this could be our "Person map"..?

eliotst (approved):
Looks good!

useful_resources/sample_code/zoom_to_van.py
shaunagm: There's a Slack thread from a few days ago where I proposed something vaguely like this, and Elyse seemed approving. It definitely feels like there's a couple things here that could be abstracted out.  I like the idea of maps, though they might need to be connector-specific (the keys in that dict are all the variable names accepted/required by find_person, upsert_person, etc **in the VAN connector**). The other element is this bit: `params = {col: participant[col] for col in column_map.keys() if participant[col]}` which is trying to make it easy to supply row data as parameters to a method call.

Issue #416, opened 2020-09-24
Add missing import information to How to Build a Connector doc


shaunagm:
The [How to Build a Connector](https://move-coop.github.io/parsons/html/build_a_connector.html) doc currently doesn't remind the user to hook up the connector to Parson's quirky import system.  See [this comment](https://github.com/move-coop/parsons/issues/217#issuecomment-698567117) (step 1) for the missing information.

davidpablocohn:
I can take this one.

shaunagm:
Thanks for offering, @davidpablocohn! Do you mind if we hold onto it for someone for whom it would be their first issue?

davidpablocohn:
That sounds like a fine idea!

I notice there's some broken stuff in the formatting of that doc as well.
I'll go ahead and leave that, but add notes to the issue as guidance for
whomever hops onto it.

davidpablocohn:
Some other things about the doc that could use a little attention:

```
/opt/ragtag/parsons/docs/build_a_connector.rst:245: WARNING: Inline literal start-string without end-string.
/opt/ragtag/parsons/docs/build_a_connector.rst:245: WARNING: Inline interpreted text or phrase reference start-string without end-string.
/opt/ragtag/parsons/docs/build_a_connector.rst:272: WARNING: Block quote ends without a blank line; unexpected unindent.
/opt/ragtag/parsons/docs/build_a_connector.rst:272: WARNING: Inline literal start-string without end-string.
/opt/ragtag/parsons/docs/build_a_connector.rst:272: WARNING: Inline interpreted text or phrase reference start-string without end-string.
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] zoom                                                                                              
/opt/ragtag/parsons/docs/build_a_connector.rst:36: WARNING: Could not lex literal_block as "python". Highlighting skipped.
/opt/ragtag/parsons/docs/build_a_connector.rst:54: WARNING: Could not lex literal_block as "python". Highlighting skipped.
```

It looks like the doc is trying to use a triple-backquote for code instead of the rst directive ``.. code-block:: python`` used elsewhere in the docbase.

Also, given the length of the doc and its internal structure, I think it could benefit from a ``.. contents::`` directive right before "Getting Started"

shepardjma:
I will work on this update.

PR #437, opened 2020-10-08
add instructions to modify parsons/_init_.py to build_a_connector.rst

shepardjma:
Documentation updates for #416

### DIFF
--- a/docs/build_a_connector.rst
+++ b/docs/build_a_connector.rst
@@ -16,11 +16,19 @@ The first thing you’ll need to do is create a new folder for your connector.  
 
 Inside the folder, create two files.  The first should be named __init__.py and should be empty.  The second will have the same name as your folder - this is the file which will have your connector’s code.  For example, in the airtable folder this file is called airtable.py and in the hustle folder it’s called hustle.py.
 
-The directory should look like this::
+The directory should look like this:
 
+.. code-block:: python
     yourconnectorname/
     __init__.py
         yourconnectorname.py
+		
+Next, add the reference to your connector to parsons/__init__.py.  Specifically, open parsons/__init__.py, scroll to the end of the other imports, and add the following:
+
+.. code-block:: python
+	from parsons.yourconnectorname.yourconnectorname import yourconnectorname
+	
+Also, in parsons/__init__.py add 'yourconnectorname' to the end of the list __all__.
 
 Once this is done, open the yourconnectorname.py file.  At the top of the file, add the following code to enable logging for our connector:
 


shaunagm:
Looking pretty good! Couple things:

- You'll need to add an extra space between the `.. code-block:: python` directive and the thing we want to display, otherwise it won't show up.
- Can you wrap the references to "parsons/\_\_init__.py" and "\_\_all__" in backticks, so they show up as code, ie like this: `__all__`?

You should be able to just add a new commit and push it to your import_docs branch, and it will show up in the PR. Let me know if you have any issues getting that to work.

docs/build_a_connector.rst
```diff
@@ -16,11 +16,19 @@ The first thing you’ll need to do is create a new folder for your connector.
 
 Inside the folder, create two files.  The first should be named __init__.py and should be empty.  The second will have the same name as your folder - this is the file which will have your connector’s code.  For example, in the airtable folder this file is called airtable.py and in the hustle folder it’s called hustle.py.
 
-The directory should look like this::
+The directory should look like this:
 
+.. code-block:: python
     yourconnectorname/
```
shaunagm, line 22: You'll need to add a blank line between the `.. code-block:: python` directive and the thing we want to display, otherwise it won't show up. I just learned that fact today!

docs/build_a_connector.rst
```diff
@@ -16,11 +16,19 @@ The first thing you’ll need to do is create a new folder for your connector.
...
     __init__.py
         yourconnectorname.py
+		
+Next, add the reference to your connector to parsons/__init__.py.  Specifically, open parsons/__init__.py, scroll to the end of the other imports, and add the following:
+
+.. code-block:: python
+	from parsons.yourconnectorname.yourconnectorname import yourconnectorname
```
shaunagm, line 29: See above re: needing a blank line here.

docs/build_a_connector.rst
```diff
@@ -16,11 +16,19 @@ The first thing you’ll need to do is create a new folder for your connector.
...
 
-The directory should look like this::
+The directory should look like this:
 
+.. code-block:: python
     yourconnectorname/
     __init__.py
         yourconnectorname.py
+		
+Next, add the reference to your connector to parsons/__init__.py.  Specifically, open parsons/__init__.py, scroll to the end of the other imports, and add the following:
```
shaunagm, line 26: Can you put the references to parsons/__init__.py in backticks `` so they show up as code?  Github works the same way, fyi, here's `parsons/__init__.py` using backticks.  (Except Github uses single backticks, and our docs use double.)

docs/build_a_connector.rst
```diff
@@ -16,11 +16,19 @@ The first thing you’ll need to do is create a new folder for your connector.
...
+
+.. code-block:: python
+	from parsons.yourconnectorname.yourconnectorname import yourconnectorname
+	
+Also, in parsons/__init__.py add 'yourconnectorname' to the end of the list __all__.
```
shaunagm, line 31: If you can use backticks here as well, for the __init__.py reference and the __all__ reference.

shaunagm (changes requested):
See my comment in the thread for requested changes.

Issue #282, opened 2020-06-23
Unit test failures on Windows


eliotst:
A user saw test failures when running `pytest` on Windows against the Parsons test package. Specifically, they saw 8 failures in the `test.test_smtp.TestSMTP` and `test.test_gmail.TestGmail`.

sarah-russell12:
I was the user who saw test failures.  I have a few screenshots of what the primary problems are.  The test failures in TestGmail all seem to be in the tests for adding attachments to emails, and essentially all looked like this:
![TestGmail-test_create_message_attachments-failed](https://user-images.githubusercontent.com/10925364/85423765-181ac580-b545-11ea-9a23-bcfb61f376fd.png)
The test failure in from TestSMP was this:
![TestSMP-failed-test](https://user-images.githubusercontent.com/10925364/85423897-3f719280-b545-11ea-8dde-07249d327bfe.png)

sarah-russell12:
I also saw this error catching message in between most of the test failure readouts, though that may just be a problem on my end:
![TestGmail-stderr-output](https://user-images.githubusercontent.com/10925364/85424131-88c1e200-b545-11ea-8cb5-967749ee6f82.png)

dannyboy15:
For the Gmail test failures it appears to be an issue with the new lines characters `\n`. I think we may have to change how we test the those specific components are there.

sarah-russell12:
As an update, I had to do a clean install of Windows recently.  To get parsons back on my computer, I followed all the documentation for forking, cloning, and then setting up a virtual environment for installing parsons dependencies.  After doing that, I ran the tests again and I got only one test failure in test_create_message_attachments in TestGmail.
![TestGmail-test_create_message_attachments-Failure](https://user-images.githubusercontent.com/10925364/91481526-93b12f80-e872-11ea-9218-6bf2f7c62138.png)
Here is the Log Call I got after it as well.
![Captured-Log-Call](https://user-images.githubusercontent.com/10925364/91481599-b2172b00-e872-11ea-8463-f90a5406295b.png)

According to Elliot, it is still an issue with newlines on Windows.

sarah-russell12:
I looked further into the issue and it is an issue with Windows newlines.  I first created a function in TestGmail that compares the output of the `Gmail._create_message_attachments` encryption and decryption against the text of the file the test opens for comparison.
[...]

PR #438, opened 2020-10-08
Fixing test_gmail.py

sarah-russell12:
This is my fix for Issue #282 Unit test failures on Windows
TestGmail.test_create_message_attachments() uses the file loremipsum_b64_txt.txt to compare against the base64 encoding of the loremipsum.txt file.  The test and file assume you are working on a non-windows machine, where reading the file would show newlines as '\n' rather than '\r\n'.  This difference in the line separator results in a different base64 encoding of the file on windows machines.
So I created loremipsum_b64_win_txt.txt and adjusted the test to use this file if the user's operating system uses '\r\n' for line separators.


eliotst (approved):
Thank you for this. Windows support sometimes gets overlooked, and I appreciate you digging in to get this fixed!

Issue #430, opened 2020-10-06
Make sure examples in docs are valid Python


shaunagm:
I noticed that some of the scripts in [this docs page](https://move-coop.github.io/parsons/html/table.html#parsons.etl.etl.ETL.coalesce_columns) are not valid Python - in particular, they're using non-strings as dictionary keys, for example:

![Screenshot from 2020-10-06 15-45-21](https://user-images.githubusercontent.com/1179362/95252436-fb9c5380-07ea-11eb-8869-a374b23184bf.png)

While this would be valid in Javascript, it's not in Python, and we don't want to mess anyone who is copying and pasting from the docs.  We should fix this and maybe also do a sweep of the docs to find and fix other examples.

ndelrossi7:
Happy to work on this!

SorenSpicknall:
Closing since the relevant documentation has since changed significantly.

PR #441, opened 2020-10-15
update parsons table docs

ndelrossi7:
Edited table docs with valid python issue #430


shaunagm (changes requested):
Hi Natalie, thank you for this! The changes in `etl.py` look good, though I haven't re-built the docs yet to double-check.

This PR has a bunch of extra files in it. I think maybe you added all the changed files in your repository, instead of just the `etl.py` file?  We don't want to include "built" files in our PR, which includes anything in `docs/html`.  We also don't want to include your personal vscode settings.

You can use a `.gitignore` file ([more on gitignore](https://www.atlassian.com/git/tutorials/saving-changes/gitignore)) to prevent things like this from getting accidentally included in a PR. Or you can just keep an eye out when adding files to a commit.

To fix this PR, you'll want to do some slightly complex stuff with git. First, you can undo your local commit with `git reset HEAD^`.  You should then be able to redo your commit, this time only adding the `etl.py` file using `git add etl.py` and committing that single addition.  To update your PR, you'll need to do a force push to overwrite the previous commit, which you undid locally. To do this, type `git push -f $yourforkname master`.  (Your fork name is whatever remote goes to your fork, to see a list of remotes type `git remote -v`.)

shaunagm (approved):

shaunagm:
Looks good to me!

Issue #412, opened 2020-09-23
Posting Events in ActionNetwork


ydamit:
I successfully extended the ActionNetwork class with the following method in some code, so I thought I'd share it here in case someone else has more capacity to improve it and write the unit tests:
```
def create_event(self, title, start_date=None, location=None):
    """
    Create an event in Action Network
    
    `Args:`
        title: str
            The public title of the event
        start_date: str OR datetime
            OPTIONAL: The starting date & time. If a string, use format "YYYY-MM-DD HH:MM:SS"
            (hint: the default format you get when you use `str()` on a datetime)
        location: dict
            OPTIONAL: A dict of location details. Can include any combination of the types of
            values in the following example:
            .. code-block:: python
            
                my_location = {
                    "venue": "White House",
                    "address_lines": [
                        "1600 Pennsylvania Ave"
                    ],
                    "locality": "Washington",
                    "region": "DC",
                    "postal_code": "20009",
                    "country": "US"
                }
    
    `Returns:`
        Dict of Action Network Event data. 
    """
    
    data = {
        "title": title
    }
    
    if start_date:
        start_date = str(start_date)
        data["start_date"] = start_date
        
    if isinstance(location, dict):
        data["location"] = location
    
    event_dict = self.api.post_request(url=f"{self.api_url}/events", data=json.dumps(data))

    an_event_id = event_dict["_links"]["self"]["href"].split('/')[-1]
    event_dict["event_id"] = an_event_id
    
    return event_dict
```

schlich:
what does this issue need before pull request?  would love to contribute to this specifically

ydamit:
@schlich I think primarily unit tests! The method itself has been working in multiple applications by TMC staff.

ydamit:
Looks like this was merged ages ago! All done!

PR #446, opened 2020-10-19
Action network create_event method

schlich:
First pull request, be gentle 😅 

Took the code shared from #412 and wrote a unit test based off the previous unit tests written for the AN module.  Please double check my testing logic as i'm still fairly new to writing unit tests especially with mock API requests.  Let me know if anything could be improved!


parsons/action_network/action_network.py
```diff
@@ -231,3 +231,51 @@ def add_tag(self, name):
...
+        data = {
+            "title": title
+        }
+
+        if start_date:
+            start_date = str(start_date)
+            data["start_date"] = start_date
+
+        if isinstance(location, dict):
+            data["location"] = location
```
jburchard, line 274: Looks like this might fail silently. I would instead suggest an if/else that raises and error if it isn't a `dict`.

parsons/action_network/action_network.py
```diff
@@ -231,3 +231,51 @@ def add_tag(self, name):
...
+
+        event_dict = self.api.post_request(url=f"{self.api_url}/events", data=json.dumps(data))
+
+        an_event_id = event_dict["_links"]["self"]["href"].split('/')[-1]
```
jburchard, line 278: This seems like a painful extraction, but it is nice that we are doing this for the user.

jburchard (changes requested):
This is really awesome. Just one minor comment. Really nice work!

parsons/action_network/action_network.py
schlich: Action Network's API is full of painful extractions in my experience! 😄 That's why I'm here after all!!

jburchard (approved):

PR #451, opened 2020-10-28
redshift: upsert(from_s3:bool) and s3_copy(template_table:str)

schuyler1d:
Instead of using a local table, this allows using the upsert command to upload an existing S3 file.

cc: @crayolakat

### DIFF
--- a/parsons/databases/redshift/redshift.py
+++ b/parsons/databases/redshift/redshift.py
@@ -800,7 +800,7 @@ class Redshift(RedshiftCreateTable, RedshiftCopyTable, RedshiftTableUtilities, R
                 if from_s3:
                     if table_obj is not None:
                         raise ValueError(
-                            'upsert() using from_s3= requires the first argument (table_obj)'
+                            'upsert(... from_s3=True) requires the first argument (table_obj)'
                             ' to be None. from_s3 and table_obj are mutually exclusive.'
                         )
                     self.copy_s3(staging_tbl,
--- a/parsons/databases/redshift/redshift.py
+++ b/parsons/databases/redshift/redshift.py
@@ -236,7 +236,7 @@ class Redshift(RedshiftCreateTable, RedshiftCopyTable, RedshiftTableUtilities, R
                 blanksasnull=True, nullas=None, acceptinvchars=True, truncatecolumns=False,
                 columntypes=None, specifycols=None,
                 aws_access_key_id=None, aws_secret_access_key=None, bucket_region=None,
-                strict_length=True):
+                strict_length=True, template_table=None):
         """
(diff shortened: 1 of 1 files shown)


eliotst:
This looks great, @schuyler1d . Thanks!

We are holding off on merging Redshift changes until a little while after the election, but I should be able to get you a review in the couple of days.

ydamit (changes requested):
This seems like a great idea. The only question in terms of usability might be whether we should have `upsert()` raise some sort of helpful error if `table_obj` is not None _and_ `from_s3` is True. Alternatively, we could make it a warning instead of an error, and allow one to override the other.

My reasoning is that I don't think we have any other Parsons methods to date where one might be expected to submit None as a positional argument, so I'm worried this would confuse people. In an ideal world, this would actually be a separate method of the `S3()` class, but that might be unnecessarily complicated to make possible just now.

ydamit:
I submitted a review yesterday and am awaiting a response.

schuyler1d:
Sorry to take so long to get back to you on this -- I've rebased and added a ValueError when table_obj AND from_s3 are present.

parsons/databases/redshift/redshift.py
```diff
@@ -785,10 +797,20 @@ def upsert(self, table_obj, target_table, primary_key, vacuum=True, distinct_che
...
                     # https://docs.aws.amazon.com/redshift/latest/dg/c_Loading_tables_auto_compress.html
                     copy_args = dict(copy_args, compupdate=False)
-                self.copy(table_obj, staging_tbl,
-                          template_table=target_table,
-                          alter_table=False,  # We just did our own alter table above
-                          **copy_args)
+                if from_s3:
+                    if table_obj is not None:
+                        raise ValueError(
+                            'upsert() using from_s3= requires the first argument (table_obj)'
```
ydamit, line 803: Did you mean to have `using from_s3=True` here?

ydamit (approved):
Looks good! Maybe just one cosmetic issue in the raised error on line 803, but that's all I could find!

parsons/databases/redshift/redshift.py
schuyler1d: I didn't, but I'm happy to make the error text whatever you think is best (and that's a good suggestion!)

ydamit: Honestly, entirely up to you. Feel free to make the change, or merge as you see fit!

schuyler1d: Ok, I tweaked it.  I don't think I have power to merge, but I'm happy if you are :pray:

ydamit (approved):

Issue #454, opened 2020-11-05
Airtable get_records column issue


mkwoods927:
On larger Airtable tables columns that are sparsely populated may get skipped in get_records unless sort is provided and the column is sorted descending.

The specific example came up in a table with 1,299 records where the column only had data in one record. When we tried to run get_records without any arguments the column doesn't come up in the returned Parsons Table. It also didn't come up when we provided a parameter for fields. Sorting by the column in descending order did work, and using a formula setting the column equal to the exact value worked as well.

We speculate that creating a Parsons Table from a list of dictionaries must take some sort of sample of the first x number of dictionaries to determine the columns and if only one dictionary has a column and it's buried deep in the list it does not get sampled.

Two likely solutions would 1. be able to specify in Table how many rows to sample or 2. have some lines in this specific Airtable method that figures out what all the fields are and fills out all the dictionaries so they all have the same number of items.

bzupnick:
I was able to successfully replicate this bug.

I created my own table with 1512 records and 2 columns; `name` and `notes`. 

I found the magic number. If there's a `note` on or below row 445, then `notes` exists in all the later rows. If there are no notes on or below row 445, then the bug is reproduced. 

I don't fully understand why yet. I've been focussing on `petl.io.json.iterdicts` which takes in `sample` and sends it to `iterpeek`. When debugging, `sample` is 1000 so I'm not sure why the magic number is still 445 but I still think that's a viable suspect for where our problem is.

bzupnick:
Docs I foudn in `petl.io.json`:

> If no `header` is specified, fields will be discovered by sampling keys
    from the first `sample` objects in `source`. The header will be
    constructed from keys in the order discovered. Note that this
    ordering may not be stable, and therefore it may be advisable to specify
    an explicit `header` or to use another function like
    :func:`petl.transform.headers.sortheader` on the resulting table to
    guarantee stability.

Again, such that this is the issue is unconfirmed. Assuming it is, though, I see 2 solutions:

1. give the `header` manually. 
2. Make the `sample` larger. 

Obviously option 1 is best for both stability and efficiency, but as I'm only on this project all of 1 day, I'm not yet confident enough to know if option 1 is possible. 

I'll keep digging!

bzupnick:
[...]

PR #462, opened 2020-12-06
enabled fixes for Airtables missing columns that are sparsely populated

bzupnick:
PR for issue #454 

The notes in the issue linked above go through the investigation. In their I laid out 2 different solutions:

1. give the `header` manually.
2. Make the `sample` larger.

This PR enables the developer to do both of those. 

The default `sample_size` remains 1000 but the developer can make that larger (or smaller) by sending that parameter to `get_records`.

Additionally, the `fields` parameter will now be sent to `petl` and it will no longer need to use sampling at all but rather will take these headers as they are.

### DIFF
--- a/parsons/airtable/airtable.py
+++ b/parsons/airtable/airtable.py
@@ -37,7 +37,7 @@ class Airtable(object):
 
         return self.client.get(record_id)
 
-    def get_records(self, fields=None, max_records=None, view=None, formula=None, sort=None):
+    def get_records(self, fields=None, max_records=None, view=None, formula=None, sort=None, sample_size=None):
         """
         `Args:`
             fields: str or lst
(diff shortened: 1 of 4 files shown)


jburchard:
Thanks for digging into this one!

I might consider setting the default parameter as 5,000. That is the [row limit](https://support.airtable.com/hc/en-us/articles/115010928147-Airtable-plans) of Airtable Paid. I would just set the default kwarg for it `sample_size=5000` to provide greater visibility for the user.

I think that this is in keeping with the ethos of Parsons in which we prioritize working under the hood rather than tuning for performance. While there might be performance hit, I think that the data sizes of Airtable is small enough that it is worth the tradeoff for the user.

bzupnick:
Got it. Okay! Yeah obviously the default `sample_size` is a `efficiency vs. usability` but TIL that here at Parsons we should lean towards the latter. I'll push that change!

jburchard (approved):

PR #463, opened 2020-12-07
copper: fix copper tests

eliotst:
This commit fixes a few bugs in the copper tests. Namely, it
fixes the pagination being done by the test code when simulating
the server, and it aligns the the expected results for one test
with what is being given to the `Copper` connector in the test.

---

cc @crayolakat

This fixes an issue where tests seem to be passing, but adding an explicit check on the number of lines when comparing tables causes things to fail.

### DIFF
--- a/parsons/etl/etl.py
+++ b/parsons/etl/etl.py
@@ -237,7 +237,7 @@ class ETL(object):
             `Parsons Table` and also updates self
 
         .. code-block:: python
-        
+
             tbl = [{'fn': 'Jane'},
                    {'lastname': 'Doe'},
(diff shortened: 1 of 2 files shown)


parsons/etl/etl.py
```diff
@@ -237,7 +237,7 @@ def map_columns(self, column_map):
             `Parsons Table` and also updates self
 
         .. code-block:: python
-
```
bzupnick, line 240: this change has to be in each PR until one of them is merged to master :joy:

SorenSpicknall (approved):

PR #467, opened 2020-12-15
tests: fix more tests

eliotst:
This commit fixes a few more tests that started failing when
adding a row count check as part of `assert_matching_tables`.

### DIFF
--- a/test/test_etl.py
+++ b/test/test_etl.py
@@ -576,22 +576,22 @@ class TestParsonsTable(unittest.TestCase):
         self.assertEqual(tbl.row_data(1), row)
 
     def test_stack(self):
-        tbl1 = self.tbl
+        tbl1 = self.tbl.select_rows(lambda x: x)
         tbl2 = Table([{'first': 'Mary', 'last': 'Nichols'}])
         # Different column names shouldn't matter for stack()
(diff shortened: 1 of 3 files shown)


test/test_etl.py
```diff
@@ -576,22 +576,22 @@ def test_row_data(self):
...
     def test_stack(self):
-        tbl1 = self.tbl
+        tbl1 = self.tbl.select_rows(lambda x: x)
         tbl2 = Table([{'first': 'Mary', 'last': 'Nichols'}])
         # Different column names shouldn't matter for stack()
         tbl3 = Table([{'f': 'Lucy', 'l': 'Peterson'}])
         tbl1.stack(tbl2, tbl3)
 
-        expected_tbl = Table(petl.stack(tbl1.table, tbl2.table, tbl3.table))
+        expected_tbl = Table(petl.stack(self.tbl.table, tbl2.table, tbl3.table))
```
SorenSpicknall, line 585: I'm interested in understanding this change. You can't reference this through tbl1?

SorenSpicknall (approved):
Approving, just have a question on one piece.

test/test_etl.py
eliotst: `tbl1` gets modified above with the `stack` method.. but I want to access the original contents from `self.tbl`.

Issue #469, opened 2021-01-04
SurveyGizmo now called Alchemer


elyse-weiss:
We should probably update things!

https://www.alchemer.com/

PR #470, opened 2021-01-04
Alchemer: Update Class Name

jburchard:
Changes name of the class from SurveyGizmo to Alchemer.

- Has some functions for backwards compatibility. Not sure if it is the most elegant way to do it.
- Adds in the documentation, which was never created.

Addresses #469

### DIFF
--- a/docs/alchemer.rst
+++ b/docs/alchemer.rst
@@ -1,0 +1,25 @@
+Alchemer
+========
+
+********
+Overview
+********
+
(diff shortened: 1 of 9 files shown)


eliotst (approved):
I think the approach to dealing with backwards compatibility is good. The other possibility that comes to mind is to just create a variable named `SurveyGizmo` and assign its value as `Alchemer`.. so:

```
SurveyGizmo = Alchemer
```

It's pretty much the same for our purposes, so no need to change.

PR #475, opened 2021-01-25
dbsync drop if needed

### LINKED ISSUE #474: add drop_if_error to Parsons DBSync
I had a DBSync set to `truncate` and it failed because of a varchar length or something equally benign. In this case, I am also more than happy to drop when needed. Seems like we could easily add a `try/except` block where the parameter option is `drop_if_error`.

elyse-weiss:
Closes #474

### DIFF
--- a/parsons/databases/db_sync.py
+++ b/parsons/databases/db_sync.py
@@ -101,9 +101,6 @@ class DBSync:
                 Full table path (e.g. ``my_schema.my_table``)
             destination_table: str
                 Full table path (e.g. ``my_schema.my_table``)
-            if_exists: str
-                If destination table exists either ``drop`` or ``truncate``. Truncate is
-                useful when there are dependent views associated with the table.
             primary_key: str
                 The name of the primary key. This must be the same for the source and
                 destination table.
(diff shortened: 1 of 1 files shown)


parsons/databases/db_sync.py
```diff
@@ -56,9 +58,13 @@ def table_sync_full(self, source_table, destination_table, if_exists='drop',
...
             if if_exists == 'drop':
                 destination_tbl.drop()
-            elif if_exists == 'truncate':
-                self._check_column_match(source_tbl, destination_tbl)
-                destination_tbl.truncate()
+            elif if_exists in ('truncate','drop_if_needed'):
+                try:
+                    self._check_column_match(source_tbl, destination_tbl)
+                    destination_tbl.truncate()
+                except:
```
SorenSpicknall, line 65: Doesn't this except clause make `truncate` and `drop_if_needed` exactly the same? i.e. I'd only do this for `drop_if_needed`, and allow `truncate` to fail if truncation doesn't work.

SorenSpicknall (changes requested):
One note on logic. Also, might be worth it to change your description to link an issue, since you're accidentally saying that this PR closes this PR and linking to the PR itself.

parsons/databases/db_sync.py
elyse-weiss: You are right! Fixing!

SorenSpicknall (approved):
Looks ready to go. Unrelated tests are failing due a known Zoom connector bug, but this functionality is passing.

PR #478, opened 2021-01-26
Zoom JWT Encryption Fix

ydamit:
Turns out it was a comedy of errors in which both the desired package, PyJWT and another similar one, python-jwt, are both imported with `import jwt`. The latter took precedence for some reason, until we added PyJWT explicitly into requirements.txt.

That worked locally, but I'd also like to try to test this live once before merging.

### DIFF
--- a/parsons/zoom/zoom.py
+++ b/parsons/zoom/zoom.py
@@ -34,7 +34,7 @@ class Zoom:
         # on JWT generation using Zoom API: https://marketplace.zoom.us/docs/guides/auth/jwt
 
         payload = {"iss": self.api_key, "exp": int(datetime.datetime.now().timestamp() + 30)}
-        token = jwt.encode(payload, self.api_secret, algorithm='HS256').decode("utf-8")
+        token = jwt.encode(payload, self.api_secret, algorithm='HS256')
         self.client.headers = {'authorization': f"Bearer {token}",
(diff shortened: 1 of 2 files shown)


parsons/zoom/zoom.py
```diff
@@ -34,7 +34,7 @@ def refresh_header_token(self):
         # on JWT generation using Zoom API: https://marketplace.zoom.us/docs/guides/auth/jwt
 
         payload = {"iss": self.api_key, "exp": int(datetime.datetime.now().timestamp() + 30)}
-        token = jwt.encode(payload, self.api_secret, algorithm='HS256').decode("utf-8")
+        token = jwt.encode(payload, self.api_secret, algorithm='HS256')
```
SorenSpicknall, line 37: Did you still intend to take the `decode` away, or was that a leftover from debugging?

SorenSpicknall (approved):

parsons/zoom/zoom.py
ydamit: That's intentional! I think the original error came from the fact that, sometime over the summer, the PyJWT package upgraded its `encode()` method to return a string instead of bytes. Then the whole package ambiguity obscured that very simple fix.

eliotst (approved):

Issue #432, opened 2020-10-06
Table.from_json use petl when line delimited


eliotst:
The `Table.from_json` method has a `lines_delimited` argument which indicates that the JSON file you are loading is a series of JSON objects that are newline separated, as so:

```json
{"one": 1, "two": 2}
{"one": 3, "two": 4}
```

As opposed to a JSON file that is a single JSON array, like:

```json
[{"one": 1, "two": 2}, {"one": 3, "two": 4}]
```

Right now, if you specify `lines_delimited=True`, the `from_json` method attempts to load the whole file and parse the JSON data all at once, which takes up a lot of memory.

petl's `fromjson` method has a `lines` argument which can be used to load a file with new-line separated JSON objects (or new line separated lists) in compliance with the JSON lines specification. When petl loads this data, it does it incrementally, so the whole file does not need to be loaded into memory at once.

We should swap the implementation of `Table.from_json` to use petl's `fromjson` function with the `lines=True` argument.

As a workaround currently, I am doing the following:

```
import petl

local_file = '/tmp/file.json'
tbl = Table(petl.fromjson(local_file, lines=True))
```

eliotst:
Update: it looks like this would require moving to perl 1.6.6

glqstrauss:
Hi! Came across this project and interested in contributing. Taking a look at this issue, I noticed a couple things 1) `petl` is not currently used for `lines_delimited=True`, so I'm assuming it `petl` t

glqstrauss:
Hi! Came across this project and interested in contributing. Taking a look at this issue, I noticed that petl is not currently used for the `lines_delimited=True` case, so I'm assuming using petl is not required to implement this enhancement?
Since this package requires python>=3.6, seems like it might be a matter of just changing the list comprehension to a generator comprehension? You'd have to keep the file descriptor open and either keep a handle on it to manually close or just let python garbage collect it.
```python
            rows = (json.loads(line) for line in open_fn(file))
```

PR #488, opened 2021-02-13
refactor(Table): Use petl to load json lines files

dannyboy15:
As mentioned in #432, `petl` supports json lines in more recent versions. Nows that the `petl` version req has been bumped to `1.6.8` it's possible to have `petl` load the data directly.


SorenSpicknall (approved):
Looks good to me!

PR #490, opened 2021-02-16
dbsync: add retries and configurable chunking

eliotst:
This commit makes a few fixes to the Parsons DBSync class
to improve the performance and stability of the functionality.

 * Add a `retries` argument to the `DBSync` class to configure
 the number of times the sync will retry reading / writing data
 if there is a failure. This should make the sync process less
 brittle.
 * Add `read_chunk_size` and `write_chunk_size` arguments to
 the `DBSync` class to allow users to configure how much data is
 downloading from the source DB or written to the destination DB
 at a time.
 * Added a `strict_length` argument to the Redshift `copy_s3` and
 all DB `copy` functions that allows users to opt out or into of new
 functionality when creating tables to use a step function when
 determining what length to specify on a `varchar` column.

---
This is one of those "went to do 1 or 2 things and ended up doing 5" kind of Pull Requests. I'm happy to rework or break up if it seems useful. I tested the DB sync script against Postgres to Postgres, Postgres to Redshift, and Redshift to Postgres.

### DIFF
--- a/parsons/databases/db_sync.py
+++ b/parsons/databases/db_sync.py
@@ -1,5 +1,7 @@
 import logging
 
+from parsons.etl.table import Table
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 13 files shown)


parsons/databases/redshift/rs_create_table.py
```diff
@@ -91,10 +101,8 @@ def data_type(self, val, current_type):
                 if not self.is_valid_integer(val):
                     return 'varchar'
 
-                # Use smallest possible int type
-                if (-32768 < t < 32767) and current_type not in ['int', 'bigint']:
-                    return 'smallint'
-                elif (-2147483648 < t < 2147483647) and current_type not in ['bigint']:
+                # Use smallest possible int type (but don't bother with smallint)
+                if (-2147483648 < t < 2147483647) and current_type not in ['bigint']:
```
SorenSpicknall, line 105: Is there any chance this would run in an environment where these comparison numbers would cause overflows during execution?

SorenSpicknall (changes requested):
Just a few small changes and questions.

parsons/databases/db_sync.py
```diff
@@ -13,21 +15,30 @@ class DBSync:
...
+            defaults to whatever the read_chunk_size is.
+        retries: int
+            The number of times to retry if there is an error processing a
+            chunk of data. The default value is 0.
     `Returns:`
         A DBSync object.
     """
 
-    def __init__(self, source_db, destination_db, chunk_size=100000):
+    def __init__(self, source_db, destination_db, read_chunk_size=100_000, write_chunk_size=None,
```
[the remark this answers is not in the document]
eliotst: Yeah. It is semantically neutral `100000 == 100_000 == 10_00_00 == 1_0_0_0_0_0` .. I just think it makes it that much easier to parse as a human to not have to count all the 0's.

parsons/databases/redshift/rs_create_table.py
eliotst: So, my understanding is that [Python handles that under the hood](https://mortada.net/can-integer-operations-overflow-in-python.html).

jburchard (approved):

SorenSpicknall (approved):
Looks ready to go now!

PR #504, opened 2021-03-16
New Upload Saved List Method for VAN

WillRaphaelson:
**Summary**
The SOAP endpoint used in [parsons.VAN.upload_saved_list()](https://github.com/move-coop/parsons/blob/a49dba54053b35bddf4f3d9f7881f5ece8f82cf4/parsons/ngpvan/saved_lists.py#L70) is using a deprecated endpoint. The new v4 endpoint has additional required arguments, only works with uploads based on VAN ID, and supports multiple column uploads. This PR Introduces:
- A new upload_saved_list_rest() method that supports VAN ID uploads with more than one column. 
- Adds a deprecation warning on the old method.

**Deprecation Plan**
- Create a Github issue for the future. When NGPVAN allows external ids, route all calls through the correct endpoint, flag upload_saved_list as deprecated with deprecation warning.
- After deprecation period - remove the function entirely or just continue to have it route to the correct endpoint, with validation on the new required arguments of the rest endpoint (callback url, for example)

**Other Potential Features Discovered During Development**
- Support the other listener types on all jobs that hit the /fileLoadingJobs endpoint? https://developers.ngpvan.com/van-api#file-loading-jobs
- Support other file upload methods than S3?
- Check if files are zipped prior to uploading to cloud storage If not, Zip em ourselves? If so, just go ahead and upload.

### DIFF
--- a/parsons/ngpvan/saved_lists.py
+++ b/parsons/ngpvan/saved_lists.py
@@ -67,7 +67,7 @@ class SavedLists(object):
         else:
             return Table.from_csv(job['downloadUrl'])
 
-    def upload_saved_list_rest(self, tbl, url_type, folder_id, list_name,
+    def upload_saved_list_rest(self, tbl, folder_id, list_name,
                                description, callback_url, columns, id_column,
                                delimiter='csv', header=True, quotes=True,
                                overwrite=None, **url_kwargs):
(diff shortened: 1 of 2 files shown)


jburchard:
This seems really great and I'm excited to see the deprecation plan for this. Let me know when you want me to dig into the code for review.

Thanks again for this!

jburchard (approved):
This is great! Thank you!

PR #507, opened 2021-03-23
Temporarily roll back JSON loading changes

SorenSpicknall:
We merged #488 yesterday with the intent of reducing memory use when loading multi-line JSON files, but something isn't being handled as expected and is causing errors in certain TMC jobs. It seems we have an edge case not caught by testing, so this PR will roll back the changes of #488 in order to stabilize things while we investigate the issue.


elyse-weiss (approved):
LGTM

PR #508, opened 2021-03-26
feat(Database): Add support for parsing booleans

dannyboy15:
Add the ability to parse (sql) booleans. By default the option is turned off, to ensure current workflows don't unintentionally get altered. 

To turn on boolean parsing set
```python
DatabaseCreateStatement.DO_PARSE_BOOLS = True
```


test/test_databases/test_database.py
```diff
@@ -8,7 +8,9 @@
 
 @pytest.fixture
 def dcs():
-    return DatabaseCreateStatement()
+    db = DatabaseCreateStatement()
+    db.DO_PARSE_BOOLS = True
```
ChrisC, line 12: I think it would be helpful to have a separate set of test cases for when this config flag is set to False (which is the default setting for it). So that we can confirm that when this is set to False, we're not parsing '0' and '1' values as Booleans. Same goes for the mysql, postgres and redshift tests.

parsons/databases/database/database.py
```diff
@@ -107,6 +115,28 @@ def is_valid_sql_num(self, val):
...
+        if not self.DO_PARSE_BOOLS:
+            return
+
+        if isinstance(val, bool) or (
+                type(val) in (int, str) and
+                str(val).upper() in self.TRUE_VALS + self.FALSE_VALS):
+            return True
+        return False
+
     def detect_data_type(self, value, cmp_type=None):
```
ChrisC, line 140: It might be easier to toggle the Boolean parsing on or off if we pass it as an optional argument in this function, instead of the config setting on the base class. Something like: 
`def detect_data_type(self, value, cmp_type=None, do_parse_bools=False):`

ChrisC (changes requested):

ChrisC (approved):

ChrisC:
@dannyboy15 I added the requested changes about separate test cases for the DB boolean parsing... I think you might have had some other ideas about how to approach but merging this in now to keep it current. Feel free to submit another PR if you want to modify the approach!

Issue #440, opened 2020-10-12
Deal with dependent views and alter_table()


elyse-weiss:
You cannot run an alter_table if there are dependent views.

dannyboy15:
I may be able to help with this. Do you have additional details, I wasn't able to find an `alter_table()` method.

elyse-weiss:
hey @dannyboy15 - `alter_table` is within `copy`. It's used for extending varchar length where necessary. However, if there are dependent views, it fails. My thought was to maybe create `alter_table_cascade` and if that is used, it would drop dependent views?

PR #511, opened 2021-03-28
feat(Redshift): Add option to alter table cascade when copying a table

dannyboy15:
This addresses #440.


jburchard (commented):
@elyse-weiss - Mind testing since I believe you first identified the issue?

SorenSpicknall (approved):

Issue #509, opened 2021-03-26
Update to Google Sheets create_spreadsheet()


mkwoods927:
We recently updated the requirements to a newer version of gspread which allows for passing a folder ID when creating a new Google Sheet. The function below updates the existing create_spreadsheet to incorporate this parameter. I've tested this code, it just needs unit tests to get over the finish line!

Note for documentation: if you create a spreadsheet in a shared folder it will be shared with people who are shared on the folder.

def create_spreadsheet(self, title, editor_email=None, folder_id=None):
  """
  Create a Google spreadsheet from a Parsons table. Optionally shares the new doc with
  the given email address. Optionally creates Google spreadsheet in specified folder.
  `Args:`
      title: str
          The human-readable title of the new spreadsheet
      editor_email: str (optional)
          Email address which should be given permissions on this spreadsheet
      folder_id: str (optional)
          ID of the folder where the Google Sheet should be created (tip: get this from the folder URL)
  `Returns:`
      str
          The spreadsheet ID
  """

  spreadsheet = self.gspread_client.create(title=title, folder_id=folder_id)

  if editor_email:
      self.gspread_client.insert_permission(
          spreadsheet.id,
          editor_email,
          perm_type='user',
          role='writer',
      )

  logger.info(f'Created spreadsheet {spreadsheet.id}')
  return spreadsheet.id

jburchard:
Want to go ahead and submit a PR for this? Looks great to me.

PR #514, opened 2021-04-08
Google Sheets: Add optional folder_id to create_spreadsheet

mkwoods927:
Added optional parameter of folder_id to the create_spreadsheet() method in the Google Sheets class.

Addresses issue #509


ChrisC:
@mkwoods927 Were you able to confirm that the live tests still work? It might also be worth adding a test similar to this one that also tests that extra field (tho I think you'll only be able to confirm that the spreadsheet exists to make the test pass). See https://github.com/move-coop/parsons/blob/master/test/test_google/test_google_sheets.py#L51-L53

ChrisC (approved):

PR #515, opened 2021-04-15
Add ActionKit mailer methods

crayolakat:
Add some methods from ActionKit's Mailer API


ChrisC (approved):
This looks great! Thanks!

PR #523, opened 2021-05-19
feat(Table): Add use_petl method

dannyboy15:
Add a method that exposes the petl functions to `parsons.Table`.

### DIFF
--- a/docs/table.rst
+++ b/docs/table.rst
@@ -249,7 +249,26 @@ If you need to iterate over the data, make sure to use the python iterator synta
 PETL
 ====
 
-The Parsons ``Table`` relies heavily on the `petl <https://petl.readthedocs.io/en/stable/index.html>`_ Python package. You can always access the underlying petl table with ``my_parsons_table.table``, which will allow you to perform any petl-supported ETL operations.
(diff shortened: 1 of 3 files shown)


jburchard:
Fwiw, I like the idea of clearly defined method. The `Table.table` will still exist. Would it be possible to update [this](https://move-coop.github.io/parsons/html/table.html#petl) documentation (show that you can do it both ways) as part of the PR to link to the new method?

dannyboy15:
@SorenSpicknall, This is more of a convenience for using `petl` functions. It allows for functions to be called directly on the `parsons.Table` as oppose to having to reassign it to the underlying `parsons.Table.table`.

Instead of this:

```python
import petl
...

tbl = Table()
tbl.table = petl.skipcomments(tbl.table, '#')
```

you would be able to do something like this

```python
tbl = Table()
tbl.use_petl('skipcomments', '#', update_table=True)
```

@jburchard, yeah I can update the docs

SorenSpicknall:
With the docs changes, I'm in support of merging this. It makes sense to do as a shortcut to using petl functions, for sure.

SorenSpicknall (approved):
Looks ready to go for me!

Issue #546, opened 2021-07-21
Redshift Create Table Float Bug


ydamit:
*Steps to Reproduce:*
```
float_tbl = Table([
    {"col1": 7.239829837, "col2": 7},
    {"col1": 23.2391287, "col2": 23},
    {"col1": 2309.47, "col2": None}
]).to_redshift('schema.table_name')

rs.query("SELECT * FROM schema.table_name")
```
This will show that the floats have been converted to ints, such that the first two rows appear to have identical values for `col1` and `col2`. Querying DDL, we find that `col1` is actually `DECIMAL(18, 0)`.

*Possible Source:*
I believe the problem might be that the constant passed to [the Redshift create_sql() method](https://github.com/move-coop/parsons/blob/914f41752c5e487cfe7d1574f912bf773794912b/parsons/databases/redshift/constants.py) for python `float` types is `decimal`. However, without specified precision or scale, the defaults of 18 and 0 are used.

*Workaround until Fixed:*
Supplying `columntypes` arg when loading to Redsihft

ydamit:
Update: just confirmed that the workaround doesn't actually help. It seems `columntypes` are somehow overridden.

ydamit:
Proposed fix: change the `DECIMAL` constant for Redshift to be `FLOAT` - will submit a PR soon.

Also, it _does_ look like the `columntypes` workaround was fine under controlled circumstances, so it's possible there was something else interfering when I claimed it wasn't working.

PR #548, opened 2021-07-22
Switch default handling of py floats as RS floats

ydamit:
Definitely want a second pair of eyes to see if there are any terrible data implications to switching from `DECIMAL` as the default data type to `FLOAT`.

### DIFF
--- a/parsons/databases/redshift/constants.py
+++ b/parsons/databases/redshift/constants.py
@@ -1,6 +1,6 @@
 COL_NAME_MAX_LEN = 120
 
-DECIMAL = 'decimal'
+DECIMAL = 'float'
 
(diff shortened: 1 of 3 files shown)


parsons/databases/redshift/constants.py
```diff
@@ -1,6 +1,6 @@
 COL_NAME_MAX_LEN = 120
 
-DECIMAL = 'decimal'
+DECIMAL = 'float'
```
ChrisC, line 3: If we're going to go down this route, let's change the constant name to `FLOAT` as well.

ChrisC (changes requested):

ChrisC (approved):

ChrisC:
closes #546

PR #550, opened 2021-07-28
URGENT: fix van connector because of ngpvan api change

bxjw:
The NGPVAN API just had a [change to get saved lists pagination](https://docs.ngpvan.com/changelog) which causes the `get_saved_lists()` function to keep paginating forever. This is a quick fix that should correct the issue, although I will admit I had to solve it pretty quickly this morning and there might be a more elegant way to do it/I haven't checked whether any other paginations are weird.

### DIFF
--- a/parsons/ngpvan/van_connector.py
+++ b/parsons/ngpvan/van_connector.py
@@ -73,7 +73,9 @@ class VANConnector(object):
         data = self.api.data_parse(r)
 
         # Paginate
-        while isinstance(r, dict) and len(r['items']) > 0:
+        while isinstance(r, dict) and self.api.next_page_check_url(r):
+            if endpoint == 'savedLists' and not r['items']:
+                break
             r = self.api.get_request(r[self.pagination_key], **kwargs)
             data.extend(self.api.data_parse(r))
 
--- a/parsons/ngpvan/van_connector.py
+++ b/parsons/ngpvan/van_connector.py
@@ -73,7 +73,7 @@ class VANConnector(object):
         data = self.api.data_parse(r)
 
         # Paginate
-        while isinstance(r, dict) and self.api.next_page_check_url(r):
+        while isinstance(r, dict) and len(r['items']) > 0:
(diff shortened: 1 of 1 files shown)


parsons/ngpvan/van_connector.py
```diff
@@ -73,7 +73,7 @@ def get_request(self, endpoint, **kwargs):
         data = self.api.data_parse(r)
 
         # Paginate
-        while isinstance(r, dict) and self.api.next_page_check_url(r):
+        while isinstance(r, dict) and len(r['items']) > 0:
```
ydamit, line 76: I believe this is a temporary error on VAN's side. Unfortunately, I think your suggestion would interfere with all sorts of other methods relying on `next_page_check_url()`, and I can confirm that it's causing some unit tests to fail.

May I recommend switching it to the following hot fix, that is more easily ignored/removable once we get VAN to fix their bug?
```
while isinstance(r, dict) and self.api.next_page_check_url(r):
    if endpoint == 'savedLists' and not r['items']:
        break
```

ydamit (changes requested):
Hi there! Sorry about the trouble you're experiencing with this!

Once I was able to reproduce the error you're encountering, I was able to determine that this is a bug on VAN's side. The response data continues to contain escalating `nextPageLink` URLs even after there aren't any actual results in `items` - which of course is why your solution worked. Thing is, this problem doesn't seem to be affecting other pagination-enabled endpoints (at least not in the spot-checking I was able to do).

I'll reach out to VAN, providing them with the evidence I collected to try convince them to correct this issue as soon as possible. Since this is urgent, however, I suggested a slightly more modular approach to your edit that I'm hoping will prevent this fix from interfering with other pagination methods, and won't make the unit tests quite so angry.

parsons/ngpvan/van_connector.py
bxjw: that works, thank you!! and appreciate you taking the time to look into this more carefully than I did

ydamit (approved):
Team work!

PR #586, opened 2021-10-20
Quickbase connector - some initial endpoint implementations

SorenSpicknall:
Not ready to merge yet - a connector that implements some basic Quickbase methods. This will need changes to testing and probably expansion of some of its functions before merging. Opening a PR right now to get easy access in one place to current results from CI routines.

### DIFF
--- a/docs/quickbase.rst
+++ b/docs/quickbase.rst
@@ -1,3 +1,4 @@
+===
 Quickbase
 ===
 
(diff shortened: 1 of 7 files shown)


SorenSpicknall:
I think I'm prepared for this to be merged now. It'll benefit from expansion later, since QuickBase has an API that enables more than what I'm doing here, but I'd rather see this functionality widely available for folks right away than delay until I have the time to implement functions for more of the API.

ChrisC (approved):

PR #588, opened 2021-10-29
Redshift: updated PR for #373 - Add dist/sort keys to upsert

ChrisC:
This cherry-picks original changes from #373 on top of the latest so it's easier to read. Also implements the original requested changes for handling primary key and sort key lists.


SorenSpicknall (approved):
Thanks for implementing the changes to handle primary key lists, Chris, and for reconciling this with newer changes. Looks ready to go to me!

PR #592, opened 2021-11-03
reverts etl column mapper to run exact match

ChrisC:
Reverts ETL column mapper fuzzy matching default (introduced [here](https://github.com/move-coop/parsons/commit/245fb922feb5975e6987fcd2ca45ffae34afaccb)). This change introduced new bugs in other connectors using this method (for example, the hustle connector).

### DIFF
--- a/parsons/etl/etl.py
+++ b/parsons/etl/etl.py
@@ -224,7 +224,7 @@ class ETL(object):
 
         return self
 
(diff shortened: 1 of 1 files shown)


ChrisC:
cc @retacg

elyse-weiss (approved):
LGTM!

Issue #537, opened 2021-06-11
table.fillna_column relocates column


cmdelrio:
Hi!

I was conflicted about whether or not this issue was serious enough to report, but I noticed that the Table.fillna_column method not only fills empty values in a column, but it also reposition the column to the last position. For me, one line of code fixes the issue, but maybe it would cause issues in a more complicated script where column position is important. 
<img width="1349" alt="Screen Shot 2021-06-10 at 5 27 52 PM" src="https://user-images.githubusercontent.com/66973815/121706142-f5a5f300-ca9a-11eb-9445-53328ab64a55.png">

Thanks!
Cormac

shaunagm:
Fixed by #603

PR #603, opened 2021-12-16
Don't move columns on Table.fill_column and Table.fillna_column

Jason94:
Fix for [issue 537](https://github.com/move-coop/parsons/issues/537).

The existing `Table.fill_column` and `Table.fillna_column` methods create a temp column with the new value, drop the old column, and then rename the existing column. This results in the "old" column being moved to the end of the table.

I changed both methods to use the petl `convert` and `update` functions to update the column in-place. This has the added benefit of allowing a calculated value to be passed to `Table.fillna_column`, which previously could only handle a fixed value. This does change the API of the function, but only in the case that someone wanted to store a reference to a function object in a Parsons `Table`.

An alternative, less invasive solution would have been to note the index of the old column before it was dropped, and then move the new column to that index. I believe that is not a trivial operation, so I think this solution will be more performant.

### DIFF
--- a/parsons/etl/etl.py
+++ b/parsons/etl/etl.py
@@ -80,9 +80,11 @@ class ETL(object):
             `Parsons Table` and also updates self
         """
 
-        self.add_column(column_name + '_column_fill_temp', fill_value)
-        self.remove_column(column_name)
-        self.rename_column(column_name + '_column_fill_temp', column_name)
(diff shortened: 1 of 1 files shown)


ChrisC:
This looks great @Jason94 ! Thanks for submitting 👍🏽

PR #604, opened 2021-12-21
Add support for temporary S3 credentials.

Jason94:
The S3 API supports creating temporary credentials for one-off operations, such as pushing a file to a particular key in a particular bucket. For example in our use case, the Mapbox API uses this to take new data for website map plugins. When S3 returns a set of temporary credentials it also returns a _session token_ that needs to be included with the standard credentials for them to be accepted.

The Parsons S3 connector uses the Boto library under the hood, which supports such session tokens. All we need to do to add support in Parsons is look for it in the env variables at the right time and pass it along to the Boto object.

Amazon documentation on session tokens & temporary credentials:
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html

### DIFF
--- a/docs/aws.rst
+++ b/docs/aws.rst
@@ -138,6 +138,26 @@ API
 .. autoclass:: parsons.S3
    :inherited-members:
 
+=====================
+Temporary Credentials
+=====================
+
+The S3 API supports creating temporary credentials for one-off operations, such as pushing a file to a particular key in a particular bucket.
+For example, the Mapbox API allows you to request temporary credentials that grant you access to a bucket where you can upload map data. 
+When S3 returns a set of temporary credentials it also returns a session token that needs to be included with the standard credentials for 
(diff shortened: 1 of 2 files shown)


ChrisC (changes requested):
@Jason94 This looks great! Would you be able to add some documentation for this to the AWS S3 doc page? See https://github.com/move-coop/parsons/blob/master/docs/aws.rst#s3

Jason94:
Thanks, Chris! I added some documentation in another commit. (https://github.com/Jason94/parsons/blob/temp-s3/docs/aws.rst#temporary-credentials)

Let me know if you want me to go into more detail.

ChrisC (approved):

ChrisC:
> Thanks, Chris! I added some documentation in another commit. (https://github.com/Jason94/parsons/blob/temp-s3/docs/aws.rst#temporary-credentials)
> 
> Let me know if you want me to go into more detail.

Awesome this looks great! Much appreciated!

PR #607, opened 2022-01-14
feat(Redshift): Add option to alter table cascade when upserting a table

neverett:
Similar to #511, except adds option to alter table cascade when upserting. When I tested with `alter_table=True` and `alter_table_cascade=False` when upserting to a test table with a dependent view, the test failed with the expected error `DETAIL:  rule _RETURN on view neverett_dev.upsert_cascade_test_view depends on column "col1"`, and when I tested with `alter_table=True` and `alter_table_cascade=True`, the dependent view was dropped and the test table was updated with new values as expected.

### DIFF
--- a/parsons/databases/redshift/redshift.py
+++ b/parsons/databases/redshift/redshift.py
@@ -790,7 +790,8 @@ class Redshift(RedshiftCreateTable, RedshiftCopyTable, RedshiftTableUtilities, R
 
         if alter_table and table_obj:
             # Make target table column widths match incoming table, if necessary
-            self.alter_varchar_column_widths(table_obj, target_table)
+            self.alter_varchar_column_widths(table_obj, target_table,
+                                             drop_dependencies=alter_table_cascade)
 
         noise = f'{random.randrange(0, 10000):04}'[:4]
         date_stamp = datetime.datetime.now().strftime('%Y%m%d_%H%M')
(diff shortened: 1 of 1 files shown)


parsons/databases/redshift/redshift.py
```diff
@@ -760,6 +760,9 @@ def upsert(self, table_obj, target_table, primary_key, vacuum=True, distinct_che
                 A temp table is dropped by default on cleanup. You can set to False for debugging.
             alter_table: boolean
                 Set to False to avoid automatic varchar column resizing to accomodate new data
+            alter_table_cascade: boolean
+                Will drop dependent objects when attempting to alter the table. If ``alter_table``
+                is not ``True``, this will be ignored.
```
ChrisC, line 765: Really minor quibble but I was a little confused reading this... 

It might be easier to say "if `alter_table` is `False`,  ..."?

ChrisC (approved):

parsons/databases/redshift/redshift.py
neverett: Yeah, that was confusing to me too -- I copied this language from the `copy()` docs, and I can change it in both places. I think what this is trying to say is "if `alter_table` is`False` or not explicitly set to `True`", since `alter_table` is False by default.

SorenSpicknall (approved):

PR #609, opened 2022-01-14
Updates Sphinx dependencies

### LINKED ISSUE #608: Sphinx docs builder no longer working
See for example the latest attempt at building our docs pages: https://app.circleci.com/pipelines/github/move-coop/parsons/3333/workflows/c0274c59-fc74-4808-9df7-0656794c97ae/jobs/3996

I did some digging into this and it appears the issue has to do with a known and unresolved issue with the `m2r` dependency. See https://github.com/miyakogi/m2r/issues/66 for more info.

ChrisC:
This also replaces the problematic markdown parser we were using (see #608) with the recommended extension for parsing markdown in the latest version of Sphinx. 

Also experimenting with adding a new CI job to test building our docs on Pull Request submissions, so we can hopefully catch issues with our docs-builder before stuff gets merged into the main branch. 

closes #608

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,16 +3,20 @@
 # Check https://circleci.com/docs/2.0/language-python/ for more details
 #
 workflows:
-  version: 2
+  version: 3
   build:
     jobs:
       - test
+      - docs-build:
+        filters:
+          branches:
+            ignore: master
       - docs-build-deploy:
           filters:
             branches:
               only: master
 
-version: 2
+version: 3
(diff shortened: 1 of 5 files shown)


ChrisC:
Confirmed that the docs are now building with these dependency updates. Also tested the sphinx `make html` command locally and confirmed that the markdown importing from https://github.com/move-coop/parsons/pull/290 is still working as well.

SorenSpicknall (approved):

PR #618, opened 2022-01-28
Create ActBlue Connector.

torimarbois:
The ActBlue Parsons connector allows the user to access the ActBlue CSV API. Authentication requires two keys: an `ACTBLUE_CLIENT_UUID` and an `ACTBLUE_CLIENT_SECRET`, which can be passed into the class constructor or set as environment variables. The user then should use the main helper function, `get_contributions()`, to make the request to the API and get the results within the Parsons table format.
This PR includes the code for the connector class, testing suite, documentation, and sample code.

The PR also adds '202' as a success code for returned status codes within the parsons/utilities/api_connector.py file, and fixes a bug preventing the provision of Google credentials via JSON file in the Google Sheets connector.


parsons/google/google_sheets.py
```diff
@@ -32,12 +32,17 @@ def __init__(self, google_keyfile_dict=None):
 
         if google_keyfile_dict is None:
             try:
-                keyfile_json = os.environ['GOOGLE_DRIVE_CREDENTIALS']
+                google_drive_credentials = os.environ['GOOGLE_DRIVE_CREDENTIALS']
```
ChrisC, line 35: For consistency and for optional handling of strings, I would recommend using this google app credential utility: https://github.com/move-coop/parsons/blob/master/parsons/google/utitities.py#L7-L20

eudaimonious: Thanks, I've modified the GoogleSheets connector to use that utility [here](https://github.com/move-coop/parsons/pull/618/commits/b2cc67cba7b66129b6e67ca4e03ec22587433d6f).

I wonder if using this utility might not be entirely backward compatible for Parsons users. It modifies the credntials environment variable, which is not something the GoogleSheets connector used to do.

SorenSpicknall (approved):

PR #623, opened 2022-02-14
Add ActBlue and QuickBase to docs sidebar

shaunagm:
Noticed that these didn't actually get linked to sidebar and are currently invisible to users.

Plus I moved shopify to after sftp because that's how the alphabet works.


neverett (approved):

eudaimonious (approved):
Thank you!

PR #627, opened 2022-03-02
Remove unused file.

cmc333333:
This was added in #581, but doesn't appear to be used by it. Probably
not super sensitive, but let's just not track it.


SorenSpicknall (approved):
Thanks for catching this!

PR #631, opened 2022-03-07
Add 'getting started with parsons' training guide

shaunagm:
This PR contains content created for the 'Getting Started With Parsons' Parsons Party written up as a first training guide.

### DIFF
--- a/docs/training_guides/getting_set_up.rst
+++ b/docs/training_guides/getting_set_up.rst
@@ -30,7 +30,7 @@ The commands you can use in the command line differ somewhat dependning on wheth
 * Use ``dir`` to list all the files and folders in a directory.
 * A `Windows command line cheat sheet <http://www.cs.columbia.edu/~sedwards/classes/2017/1102-spring/Command%20Prompt%20Cheatsheet.pdf>`_ can help you keep track of which commands are which.
 
-You do not have to type everything on the command line out by hand. You can auto-complete the names of of files/folders in your current directory by tapping the tab key. On Mac/Linux you can also tab-complete installed programs. And you can access previous commands by clicking the up and down arrows. Save your hands! Learn these tricks.
+You do not have to type everything on the command line out by hand. You can auto-complete the names of files/folders in your current directory by tapping the tab key. On Mac/Linux you can also tab-complete installed programs. And you can access previous commands by clicking the up and down arrows. Save your hands! Learn these tricks.
 
 ***************************************
 Step 2: Set Up Your Virtual Environment
--- a/docs/training_guides/getting_set_up.rst
+++ b/docs/training_guides/getting_set_up.rst
@@ -92,7 +92,7 @@ On Windows, absolute paths look a little different. They start with the letter o
 
 In these instructions we try to use absolute paths, even though they're a little wordier, because it's less likely to cause problems for you if you run them from an unexpected place.
 
-Confusingly, there's an important environmental variable called **$PATH**. The $PATH is a list of places your computer will check when searching for installed libraries and scripts. You can check what's currently in your $PATH by typing ``echo $PATH`` (Mac/Linux) or ``echo %PATH%`` (Windows).
+In addition to paths, there's an important environmental variable called **$PATH**. The $PATH is a list of absolute paths your computer will check when searching for installed libraries and scripts. You can check what's currently in your $PATH by typing ``echo $PATH`` (Mac/Linux) or ``echo %PATH%`` (Windows).
 
 When you activate your virtual environment, the path to the environment is placed as the first path. Paths are checked in order from first to last. You can check what packages have been installed in your virtualenv (and thus should be available on the path when the virtualenv is activated) by looking in ``lib/site-packages``.
 
--- a/docs/training_guides/getting_set_up.rst
+++ b/docs/training_guides/getting_set_up.rst
@@ -30,7 +30,7 @@ The commands you can use in the command line differ somewhat dependning on wheth
 * Use ``dir`` to list all the files and folders in a directory.
 * A `Windows command line cheat sheet <http://www.cs.columbia.edu/~sedwards/classes/2017/1102-spring/Command%20Prompt%20Cheatsheet.pdf>`_ can help you keep track of which commands are which.
 
-You do not have to type everything on the command line out by hand. You can auto-complete the names of files/folders in your current directory by tapping the tab key. On Mac/Linux you can also tab-complete installed programs. And you can access previous commands by clicking the up and down arrows. Save your hands! Learn these tricks.
+You do not have to type everything on the command line out by hand. You can auto-complete the names of files/folders in your current directory by tapping the tab key. On Mac/Linux you can also tab-complete installed programs. And you can access previous commands via the up and down arrows. Save your hands! Learn these tricks.
 
 ***************************************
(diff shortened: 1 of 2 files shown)


docs/training_guides/getting_set_up.rst
```diff
@@ -0,0 +1,172 @@
...
+* A `Mac/Linux command line cheat sheet <https://www.guru99.com/linux-commands-cheat-sheet.html>`_ can help you keep track of which commands are which.
+
+**Windows**:
+
+* You can use ``cd`` to figure out where you currently are.
+* To move around, use ``cd`` (for example ``cd ..`` means "go up one directory" or ``cd my_folder`` which means "go into ``my_folder``").
+* Use ``dir`` to list all the files and folders in a directory.
+* A `Windows command line cheat sheet <http://www.cs.columbia.edu/~sedwards/classes/2017/1102-spring/Command%20Prompt%20Cheatsheet.pdf>`_ can help you keep track of which commands are which.
+
+You do not have to type everything on the command line out by hand. You can auto-complete the names of installed programs as well as files/folders in your current directory by tapping the tab key. You can also access previous commands by clicking the up and down arrows. Save your hands! Learn these tricks.
```
Jason94, line 33: Windows auto-completion is a bit more limited than Unix systems. It only autocompletes the names of files/folders in the current directory, not programs. Perhaps we should have this paragraph under both OS sections here, and then modify the Windows paragraph accordingly.

docs/training_guides/getting_set_up.rst
```diff
@@ -0,0 +1,172 @@
...
+
+Regardless of what version you're on, you can activate your virtual environment with the command: ``source $path_to_your_env/$your_env_name/bin/activate``.
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Windows Virtual Environments
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Start by installing virtualenvwrappers from source::
+
+  	git clone git://github.com/davidmarble/virtualenvwrapper-win.git
```
Jason94, line 67: We haven't covered Git yet, so should we add a little note above saying `(See below section to install Git)` or something?

docs/training_guides/getting_set_up.rst
```diff
@@ -0,0 +1,172 @@
...
+Paths vs $PATHs
+^^^^^^^^^^^^^^^
+
+Paths are how we refer to a file or program's location in the file system. For example, ``home/janedoe/virtualenvs`` says that within the top-level directory ``home`` there is a directory named ``janedoe``, and within ``janedoe`` there is a directory ``virtualenvs``.
+
+``home/janedoe/virtualenvs`` is an **absolute path** because it specifies exactly how to get there no matter where you are in the system. The path ``janedoe/virtualenvs`` is a **relative path** because it only works if you use it from the home directory. Trying to use a relative path from the wrong location is a common source of command line errors!
+
+In these instructions we try to use absolute paths, even though they're a little wordier, because it's less likely to cause problems for you if you run them from an unexpected place.
+
+Confusingly, there's an important environmental variable called **$PATH**. The $PATH is a list of places your computer will check when searching for installed libraries and scripts. You can check what's currently in your $PATH by typing ``echo $PATH`` (Mac/Linux) or ``echo %PATH%`` (Windows).
```
Jason94, line 91: I don't think we should say it's confusing. I think it's generally good practice to be empowering in documentation, particularly for beginners.

Jason94 (changes requested):

docs/training_guides/getting_set_up.rst
shaunagm: Do you see that wording as disempowering? My instinct is that people find it empowering to hear that other people find confusing things confusing, but OTOH if I'm the only one who finds path vs $PATH confusing I definitely don't want to make others doubt themselves.

Jason94: Yeah, I think so. I think you're right - it is a confusing concept. I think a good analogy is GotV literature. We know that voting in a lot of places is confusing and difficult to do properly. But it's been found that it's important to have a positive message about voting ("Voting is easy", "Voting is fun", etc) even if we don't believe it, because people won't vote if we're telling them it's difficult. Even if we're also telling them how to do it.

In any case, I don't really think it's _that_ important - it's just one word. I'll defer to you, but did want to share my opinion.

shaunagm: No, you've convinced me. :)  I changed the wording.

docs/training_guides/getting_set_up.rst
```diff
@@ -0,0 +1,172 @@
...
+
+.. _path-explainer:
+
+^^^^^^^^^^^^^^^
+Paths vs $PATHs
+^^^^^^^^^^^^^^^
+
+Paths are how we refer to a file or program's location in the file system. For example, ``home/janedoe/virtualenvs`` says that within the top-level directory ``home`` there is a directory named ``janedoe``, and within ``janedoe`` there is a directory ``virtualenvs``.
+
+``home/janedoe/virtualenvs`` is an **absolute path** because it specifies exactly how to get there no matter where you are in the system. The path ``janedoe/virtualenvs`` is a **relative path** because it only works if you use it from the home directory. Trying to use a relative path from the wrong location is a common source of command line errors!
```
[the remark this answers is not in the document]
shaunagm: Whoops, great catch. Made both those changes.

docs/training_guides/getting_set_up.rst
shaunagm: Yes, great idea! Added.

docs/training_guides/getting_set_up.rst
shaunagm: I just re-wrote it so it was clear the tab-complete for installed programs was only for Mac/Linux. Thanks for catching that!

docs/training_guides/getting_set_up.rst
```diff
@@ -0,0 +1,178 @@
...
+* A `Mac/Linux command line cheat sheet <https://www.guru99.com/linux-commands-cheat-sheet.html>`_ can help you keep track of which commands are which.
+
+**Windows**:
+
+* You can use ``cd`` to figure out where you currently are.
+* To move around, use ``cd`` (for example ``cd ..`` means "go up one directory" or ``cd my_folder`` which means "go into my_folder").
+* Use ``dir`` to list all the files and folders in a directory.
+* A `Windows command line cheat sheet <http://www.cs.columbia.edu/~sedwards/classes/2017/1102-spring/Command%20Prompt%20Cheatsheet.pdf>`_ can help you keep track of which commands are which.
+
+You do not have to type everything on the command line out by hand. You can auto-complete the names of of files/folders in your current directory by tapping the tab key. On Mac/Linux you can also tab-complete installed programs. And you can access previous commands by clicking the up and down arrows. Save your hands! Learn these tricks.
```
Jason94, line 33: Two 'of'

Jason94 (changes requested):

Jason94 (approved):
Looks good to me! Per our ongoing discussion about PR review I don't know if I'm authoritative enough to allow it to get merged. But it looks good to me :)

shaunagm:
@neverett or @SorenSpicknall, it looks like Jason's review is not enough - can you take a look and approve?

neverett (approved):

Issue #517, opened 2021-04-22
Improve parsons experience when limiting needed dependencies


schuyler1d:
Currently there is a somewhat elaborate way to use the `parsons` python library without installing all the dependencies.
https://move-coop.github.io/parsons/html/index.html#minimizing-resource-utilization

As more people use Parsons in deployment contexts where packaging all possible dependencies increases, this approach is more awkward for a larger part of the community.

This original workaround came about after one of my (first) PRs (https://github.com/move-coop/parsons/pull/147) was merged at the same time as an additional connector was added from a second PR and the second one was broken by the first's new requirement for a different pattern in the `parsons/__init__.py` file.

This issue is meant to discuss possible ways forward to improve the library ergonomics of installing fewer dependencies.

1. I think it is still possible to do the approach of #147 as a first step -- with more planning and testing this time.
2. If we did that, then there's a further step that might make it even easier to support this use-case -- instead of having an environment variable, etc, we would create a second package (`parsons-core` maybe) that used the `extras_require=` feature in pip's run of setup.py files for different components and their dependencies.

I'm interested in whether maintainers would support or allow this work to move forward and/or if there are other ideas or preferences pro/con from the community.

apiology:
👋 

I'm the original author of https://github.com/bluelabsio/records-mover, which has a similar problem with a very large dependency pool that is not always needed by any one user.

Based on the advice of @cwegrzyn (the current maintainer), I moved to a system of named extras via `extras_require=` - you can see it here: https://github.com/bluelabsio/records-mover/blob/master/setup.py#L337-L360

Along with that, I added CI scenarios to make sure key core tasks could be done with the default dependencies, and changed imports throughout to be more lazy - meaning folks would generally not get an `ImportError` on a given rarely-needed package until they tried to run the requested scenario.

The downsides were the extra complexity in the CI scenarios and in setup.py, and the fact that novice users aren't generally aware of the concept of extras.  What happens in practice is that many users will `pip install records_mover`, run it, hit a missing dependency-related `ImportError`, and then `pip install` that dependency.  In some cases that means they get a version that's not (yet) compatible with the code, but that's generally been rare.

For more advanced users, the extras do provide a way to let Records Mover control the version constraints needed.  This has been useful when, for example, a version of Pandas got released which broke a use case we relied on.

[...]

PR #632, opened 2022-03-08
Prototype for modular dependencies.

cmc333333:
For #517 , I wanted to propose a backwards-compatible solution!

Currently, setup.py declares that Parson requires a ton of pinned
dependencies. This is problematic for incorportating Parsons into a
larger code base for a few reasons:
* Extra dependencies increase the attack surface, which is a bad
trade-off if they aren't ever used
* Extra dependencies increase the build size, which, in turn increases
testing and deployment time
* Pinning to specific versions often causes conflicts (dependency hell)

The docs recommend installing parsons separately from the rest of an
app's dependencies, using the "--no-deps" flag, but that doesn't play
well with package locking tools like pipenv and poetry.

This prototype tries to give us a backwards-compatible offramp.
Following the process used in python-slugify 1.x.x, it changes which
dependencies are available at install time, dependending on an
environment variable. So,

```sh
pip install parsons  # retain current behavior
PARSONS_LIMITED_DEPENDENCIES=true pip install parsons  # core dependencies only
PARSONS_LIMITED_DEPENDENCIES=true pip install parsons[ngpvan]  # core + ngpvan(suds)
PARSONS_LIMITED_DEPENDENCIES=true pip install parsons[google,ngpvan]  # core + google + ngpvan
PARSONS_LIMITED_DEPENDENCIES=true pip install parsons[all]  # core + all optional deps
```

In some future, major release we can remove the guard and recommend
`parsons[all]` where appropriate.

This does *not* attempt to find all optional dependencies -- I wanted to
throw the proof of concept up for review first.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,12 +2,38 @@ import os
 from setuptools import find_packages
 from distutils.core import setup
 
-THIS_DIR = os.path.abspath(os.path.dirname(__file__))
 
 
 def main():
-    with open(os.path.join(THIS_DIR, 'requirements.txt')) as reqs:
-        requirements = reqs.read().strip().split('\n')
+    limited_deps = os.environ.get("PARSONS_LIMITED_DEPENDENCIES", "")
+    if limited_deps.strip().upper() in ("1", "YES", "TRUE", "ON"):
+        install_requires = [
+            "petl",
+            "python-dateutil",
+            "requests",
+            "simplejson",
+        ]
+        extras_require = {
+            "google": [
+                "google-cloud-bigquery",
+                "google-cloud-storage",
+                "gspread",
+                "oauth2client",
+            ],
+            "ngpvan": [
+                "suds-py3",
+            ],
+        }
+        extras_require["all"] = sorted({
+            lib
+            for libs in extras_require.values()
+            for lib in libs
+        })
+    else:
+        THIS_DIR = os.path.abspath(os.path.dirname(__file__))
+        with open(os.path.join(THIS_DIR, 'requirements.txt')) as reqs:
+            install_requires = reqs.read().strip().split('\n')
+        extras_require = {}
 
     setup(
         name="parsons",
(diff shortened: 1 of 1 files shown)


SorenSpicknall:
Hey @schuyler1d, not sure if you're checking in on GitHub much these days, but if you want to give some feedback on an evolution of Parsons dependency management, this sounds like a formalization of some of the concepts you were talking about when we implemented the --no-deps flag.

schuyler1d:
@SorenSpicknall thanks for the ping.
I support this direction @cmc333333 !  I believe you will still need to do this kind of change in `parsons/__init__.py`
https://github.com/move-coop/parsons/compare/master...schuyler1d:try-imports?expand=1
and BE CAREFUL when releasing another Parsons version that any new modules are ALSO ADDED (this is how it got reverted in a past attempt).
I also think once we merge this we should immediately update the documentation on installations to `parsons[all]` for installations, so if/when we remove the env var and change the default to be more minimal, then we recent versions will already have included the `[all]` as the best practice.

cmc333333:
Thanks @schuyler1d ! I think the removal of the `PARSONS_SKIP_IMPORT_ALL` flag is a great idea, but doesn't seem strictly necessary for this (kind of like how `PARSONS_SKIP_IMPORT_ALL` is separate from the `--no-deps` guidance currently).

Adding a noop "all" even when `PARSONS_LIMITED_DEPENDENCIES` is not present is a great idea; I'll update the PR soonish!

schuyler1d:
Fair enough, but then we're requiring people navigate two obscure env bars for the same use case. If you are not replacing the old one being needed maybe you can use the same env variable?
It does seem like a good opportunity to do it....

cmc333333:
Great idea to use a single env variable! I don't want to break backwards compatibility, so I won't piggy back off the existing variable, though. Changesets shortly.

cmc333333:
Integrated comments! Can I get a :+1: on this being the right direction? Once I do, I'll work through the other integrations and add their requirements to setup.py.

cmc333333:
Alrighty, I've gone through and added all of the connectors with special dependencies as "extras"; I think this is ready for final review!

SorenSpicknall (approved):
This looks good to go to me. I'll give it a day or two before merging in case somebody else spots something I didn't.

PR #639, opened 2022-03-18
Multiversion docs

shaunagm:
Here's a first stab at setting up multiversioning for the docs using https://github.com/Holzhaus/sphinx-multiversion

Deployment is a bit more complicated, so I've stuck the steps in a `deploy_docs.sh` script. I couldn't figure out the Makefiles so I left them alone for now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,7 +91,7 @@ jobs:
           name: Build docs
           command: |
             . venv/bin/activate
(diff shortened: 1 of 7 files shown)


SorenSpicknall (approved):
We'll need to closely monitor once live and make relevant changes to any broken absolute links, but this looks good to go!

PR #641, opened 2022-03-22
Add filtering to Action Network connector

joe-irving:
As laid out in the [Searching and filtering (OData)](https://actionnetwork.org/docs/v2/) section of the Action Network documentation, you can filter resource lists by fields like `modified_date`, which is needed for any kind of database sync and avoids unnecessary API calls.

The `filter` argument has been added to `_get_page`, and then `_get_entry_list`, `get_people` and `get_tags`. 

This is my first pull request to parsons, but looking forward to making more! (Also be nice if I have made any silly mistake please)


SorenSpicknall:
Hey Joe, this looks like a great addition to the connector. Two questions for you:

1. Have you tested this using your local version of Parsons to see if the functions work as expected with the modified parameters? I'm particularly interested in whether there might be unintended side effects when the new parameter is left empty. It's likely safe, but good to test.
2. Speaking of testing, have you taken a look at [the way tests are structured for the Action Network connector](https://github.com/move-coop/parsons/blob/master/test/test_action_network/test_action_network.py)? Before merging, it likely makes sense to create a set of tests with mocked responses for use of this new parameter.

joe-irving:
@SorenSpicknall is this okay to merge?

SorenSpicknall (approved):

SorenSpicknall:
> @SorenSpicknall is this okay to merge?

Thanks for the ping! Just took care of it; sorry for the delay.

PR #658, opened 2022-04-05
Update README license to match docs.

cmc333333:
Also point to `main` instead of `master`.


SorenSpicknall (approved):
Thanks for catching this!

Issue #445, opened 2020-10-17
Adding a note to the NGP VAN apply_activist_code function


vischalla:
Can we add a note in the Parsons docs on the NGP VAN appy_activist_code function that using it will also apply a canvass status to the person? A member raised that this caused an issue in their canvassing counts. It wasn’t clear why the previously loaded activist codes were impacting canvass results without looking at the underlying VAN API docs.

Source docs: https://developers.ngpvan.com/van-api#activist-codes

jburchard:
What was the canvass mode that was listed?

bzupnick:
Is this issue closed from that merged PR? Or should we be adding something in the docs of this behavior?

shaunagm:
No, the issue is still open - there's nothing in the docs about this.

PR #663, opened 2022-04-12
VAN: Activist Code Apply - Omit Contact History

jburchard:
The ``van.apply_activist_code`` method currently applies a contact response (e.g. canvass status) when it is run. This is problematic in that many use cases for applying activist codes are not the result of contacts. Instead they are arbitrary flags applied to records. This can be a pain as it introduces fake contact attempts which muck up reporting for organizations (i.e. me).

Here is an example of what happens when you currently apply an activist code.

ss3

There was an old issue #445 which lifted up, but it seems like in the meantime, VAN has addressed it. This PR adds in a kwarg ``omit_contact`` that omits, by default, a contact response being added when an activist code is applied. 

I have coded this such that the default is to omit the contact, which, while it would represent a change in functionality compared to the previous version of the method, I believe is representative of the desired defaults of most use cases.

### DIFF
--- a/parsons/ngpvan/activist_codes.py
+++ b/parsons/ngpvan/activist_codes.py
@@ -45,7 +45,7 @@ class ActivistCodes(object):
     def toggle_activist_code(self, id, activist_code_id, action, id_type='vanid',
                              omit_contact=True):
         # Internal method to apply/remove activist codes. Was previously a public method,
-        # but for the sake of simplicity, breaking out into two public methods.
+        # but for the sake of simplicity, breaking out into two public methods. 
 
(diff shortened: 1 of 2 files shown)


SorenSpicknall (approved):
This looks good to me! The Docker build failed at the time this was submitted due to an infrastructure issue, and I'm going to see if I can get it to rerun successfully before we merge...

SorenSpicknall:
Hey @jburchard, in order to surmount our fight with Docker's build behavior after the breakage we had last week, could you push a minor functionally inconsequential update to your PR, like a tiny change to a docstring? I want to make sure we can get a successful build before merging.

PR #667, opened 2022-04-14
Fix for Google Dependencies

SorenSpicknall:
Corrects issues with Google dependencies that resulting in a build bug described in the Parsons Slack [here](https://tmc-parsons.slack.com/archives/C018CQD3G4E/p1649818619946079).


shaunagm (approved):
lgtm

Issue #642, opened 2022-03-22
Switch dependency to newer simple-salesforce version 


cmdelrio:
I've been getting emails that salesforce is retiring old versions of their api. The Parsons simple-salesforce dependency is a super outdated version. It's not clear in their documentation which salesforce api version they use, but to be safe I think we need to update to a newer version.

shaunagm:
@cmdelrio do those emails say when they old API will no longer be supported/will be retired? Trying to figure out how urgent a priority this is.

cmdelrio:
May 22nd @shaunagm

shaunagm:
Okay, that seems fairly high priority, then.

bzupnick:
I believe this issue should also be closed as resolved!

PR #682, opened 2022-05-06
Update simple-salesforce version

shaunagm:
Addresses #642 by bumping [simple-salesforce](https://github.com/simple-salesforce/simple-salesforce) version from 0.74.3 to 1.11.6. 

0.74.3 was published July 4 2019, so we've got about 3 years of [changes](https://github.com/simple-salesforce/simple-salesforce/blob/master/CHANGES). I went through and looked at the tests to make sure that the client interfaces hadn't introduced breaking changes and I don't believe they have, but I don't have a Salesforce account so I couldn't actually test that. 

I did note that we have two empty tests in `test_salesforce.py` and I considered trying to add them but I don't know enough about the API to mock the data.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,7 +26,7 @@ docutils<0.18,>=0.14
 urllib3==1.26.5
 simplejson==3.16.0
 twilio==6.30.0
-simple-salesforce==0.74.3
+simple-salesforce==1.11.6
 suds-py3==1.3.4.0
 newmode==0.1.6
 mysql-connector-python==8.0.18


SorenSpicknall (approved):

SorenSpicknall:
Since the old version is being deprecated in less than two weeks, I'm fine merging this without implementing new tests first. We should make note of the empty tests for a later PR, however.

cmdelrio:
Hey all! Sorry, I've been meaning to check whether this will break some of the script we have running. If it does, I'll figure it out since we don't really have any other option as Soren pointed out. Thanks all!

Issue #669, opened 2022-04-21
Re-name Phone2Action to Capitol Canary


elyse-weiss:
"now is the time to retire the Phone2Action name and move forward with a new, more focused identity as we continue to transform how policy is won. Call us Capitol Canary."

we should probably re-name the connector? check out their API docs to confirm?

shaunagm:
Moving my thoughts from Slack:

Proposal:
1 - create a new capitol canary class that P2A wraps around (so when P2A is eventually removed, it doesn't change capitol canary)
2 - when using the P2A, log a warning about the class being removed by a certain date, and instruction to switch to capitol canary
3 - the docs keep P2A but just as a brief statement that says its deprecated and you should use capitol canary
4 - at whatever the deprecation date is, we remove P2A connector and its associated docs, but leave a note on the capitol canary docs saying that it used to be the P2A connector and that you can see its docs in older versions of the docs

A PR to address this issue should handle steps 1-3, and then we can schedule a reminder near the deprecation date to do step 4.

bzupnick:
@shaunagm should this get closed?

shaunagm:
Yes! Thank you for helping keep the issue tracker up to date. And for the PR of course :)

PR #687, opened 2022-05-15
Re-named Phone2Action to Capitol Canary

bzupnick:
This is a PR for issue https://github.com/move-coop/parsons/issues/669.

I couldn't find API docs anywhere so this PR assumes all of the URLs and environment variables maintain the `phone2action` branding.

### DIFF
--- a/test/test_p2a.py
+++ b/test/test_p2a.py
@@ -127,7 +127,7 @@ class TestP2A(unittest.TestCase):
 
     def setUp(self):
 
-        self.p2a = Phone2Action(app_id='an_id', app_key='app_key').capitol_canary
+        self.p2a = Phone2Action(app_id='an_id', app_key='app_key')
 
     def tearDown(self):
 
@@ -146,8 +146,8 @@ class TestP2A(unittest.TestCase):
         os.environ['PHONE2ACTION_APP_KEY'] = 'key'
 
         p2a_envs = Phone2Action()
-        self.assertEqual(p2a_envs.capitol_canary.app_id, 'id')
-        self.assertEqual(p2a_envs.capitol_canary.app_key, 'key')
+        self.assertEqual(p2a_envs.app_id, 'id')
+        self.assertEqual(p2a_envs.app_key, 'key')
 
     @requests_mock.Mocker()
     def test_get_advocates(self, m):
--- a/test/test_p2a.py
+++ b/test/test_p2a.py
@@ -127,7 +127,7 @@ class TestP2A(unittest.TestCase):
 
     def setUp(self):
 
-        self.p2a = Phone2Action(app_id='an_id', app_key='app_key')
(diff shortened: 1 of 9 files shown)


shaunagm (changes requested):
These code changes look good.  The only hesitation I have is that this is technically a Capitol Canary class wrapping Phone2Action, which means that at the future moment when we remove Phone2Action, we'll have to do the work of moving over the methods, instead of just deleting it.

There's no particular reason to do that now vs later, except that it'll be easier for our future selves. Up to you @bzupnick if you want to go ahead and make that change now or leave it for later.

Also, it would be good to merge the docs changes at the same time. Probably the best approach is:

(1) rename the Phone2Action rst doc for Capitol Canary, and change references to Capitol Canary at the top. 
(2) create a new Phon2Action doc with the same name as the old name (so old links won't break) that says something like, "Phone2Action has been renamed Capitol Canary. All methods should be the same." with a link to the Capitol Canary docs.

shaunagm (approved):
Thanks for making all these changes! Enjoy your vacation. :)

test/test_p2a.py
```diff
@@ -127,7 +127,7 @@ class TestP2A(unittest.TestCase):
 
     def setUp(self):
 
-        self.p2a = Phone2Action(app_id='an_id', app_key='app_key')
+        self.p2a = Phone2Action(app_id='an_id', app_key='app_key').capitol_canary
```
shaunagm, line 130: Ahhh just noticed this. Sorry for assuming we were done!  😭 

@bzupnick, can you leave this as `self.p2a = Phone2Action(app_id='an_id', app_key='app_key')`? We want people to be able to keep using the Phone2Action class for now, with just a warning, which means we need to test that the wrapper is actually working by calling the class's methods directly, rather than testing `.capitol_canary`'s methods.

bzupnick: The reason I did that was since below we use things like `self.p2a.client.uri` which doesn't exist in p2a anymore.

But now that I write this out, I wonder if that means that my solution is incomplete and I should _also_ bring in not just the functions, but also the class variables 🤔

shaunagm: Let's try overriding `__getattr__`. `__getattr__` is a "magic method" that is only called when the attribute is not found. If we override it to look for the same attribute on CapitolCanary, that should move over all the attributes. 

Should look something like this (this is from memory, the syntax may be off and its definitely oversimplified):

```

class Phone2Action:

  self.__init__()
    self.capitolcanary = CapitolCanary

  self.__getattr__(self, name):
    return getattr(self.capitolcanary, name)
```

shaunagm:
Hi @bzupnick, want to try and pick this up again? Happy to pair on this last piece (and sorry again that I didn't notice it until after our last pairing session).

shaunagm (approved):
lgtm

Issue #691, opened 2022-05-18
Bug in GoogleBigQuery Copy NoneType Handling


ydamit:
The [internal method](https://github.com/move-coop/parsons/blob/main/parsons/google/google_bigquery.py#L289), `_generate_schema()`, uses the Parsons Table method `get_columns_type_stats()` and parses it with the `BIGQUERY_TYPE_MAP` reference, rather than going through the whole `detect_data_type()` [process](https://github.com/move-coop/parsons/blob/main/parsons/databases/database/database.py#L140) the other database types use.

Unfortunately, `BIGQUERY_TYPE_MAP` doesn't account for columns with NoneType under certain circumstances. AS a result, `copy()` can fail if there are any non-string columns that also contain NoneType values.

It's possible that the better long-term solution would be to incorporate the BigQuery process. But a quick workaround is also possible that filters out NoneTypes before any application of `BIGQUERY_TYPE_MAP`

ydamit:
@elyse-weiss & @zerlinac for awareness!

SorenSpicknall:
Fixed in #695

PR #695, opened 2022-05-31
Big Query NoneType Mapping Bug

scmora:
To ensure that columns with different data types are stored accordingly after running the copy function to big query, we have added a mapping line to the python to Big Query data type map `'NoneType': 'String'`.

The other alternative was a set of codes that filtered out NoneType from the get_column_stats function and kept the first type that was available other than NoneType, i.e., string, int, float.  

We landed on the addition below as it will provide more flexibility when multiple data types are present in the column.


SorenSpicknall (approved):

shaunagm:
Does this close #691?

PR #696, opened 2022-06-01
VAN: Fix codes bugs.

jburchard:
Two small bug fixes:

- `get_code` : The method for the underlying VANConnector class was misnamed.
- `apply_code`: The _is_applicable_ parameter was misnamed and thus, not applying.

Furthermore, I might want to rework the _supported_entities parameter. It requests constructing a dict, which is sort of unnecessary and not keeping with the common sense abstraction of Parsons. But, that would come in another PR.

### DIFF
--- a/parsons/ngpvan/codes.py
+++ b/parsons/ngpvan/codes.py
@@ -53,7 +53,7 @@ class Codes(object):
                 See :ref:`parsons-table` for output options.
         """
 
-        c = self.connection.request(f'codes/{code_id}')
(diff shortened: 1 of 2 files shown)


shaunagm (approved):
This looks good to me, but why wasn't it caught in the tests? Do we want to update the tests to cover this?

jburchard:
Test added. Ready for re-review. @shaunagm

shaunagm:
Looks good to me. Thanks for adding the test, @jburchard!

Issue #673, opened 2022-04-27
NGPVAN -- update_person_json function doesn't update person record


harnoorSingh-wtpmi:
The update_person_json() function does not seem to function as the documentation indicates. When I use the function to attempt to update a contact (in the demo EA site using a sandbox the EA team setup for me) in the following way:

`van = VAN(db='MyCampaign')
van.update_person_json(id=101991312, match_json={'codes':None})`

I get a strange JSON output that is completely different from the actual person's record (it is shown below)

```
{'vanId': 101991312,
 'firstName': 'joe',
 'lastName': 'jack',
 'middleName': None,
 'suffix': None,
 'title': None,
 'sourceFileTitle': None,
 'sourceFileFirstName': None,
 'sourceFileMiddleName': None,
 'sourceFileLastName': None,
 'sourceFileSuffix': None,
 'contactMode': None,
 'organizationContactCommonName': None,
 'organizationContactOfficialName': None,
 'salutation': 'joe',
 'formalSalutation': 'joe jack',
 'additionalSalutation': None,
 'preferredPronoun': None,
 'pronouns': None,
 'envelopeName': 'joe jack',
 'formalEnvelopeName': 'joe jack',
 'additionalEnvelopeName': None,
 'contactMethodPreferenceCode': None,
 'nickname': None,
 'website': None,
 'professionalSuffix': None,
 'party': None,
 'employer': None,
 'occupation': None,
 'jobTitle': None,
 'sex': None,
 'dateOfBirth': None,
 'selfReportedRace': None,
 'selfReportedEthnicity': None,
 'selfReportedRaces': None,
 'selfReportedEthnicities': None,
 'selfReportedGenders': None,
 'selfReportedSexualOrientations': None,
 'selfReportedLanguagePreference': None,
 'emails': None,
 'phones': None,
 'addresses': None,
 'recordedAddresses': None,
 'identifiers': None,
 'codes': None,
 'customFields': None,
 'primaryCustomField': None,
 'contributionSummary': None,
 'suppressions': None,
 'caseworkCases': None,
 'caseworkIssues': None,
 'caseworkStories': None,
 'notes': None,
 'scores': None,
 'customProperties': None,
 'electionRecords': None,
 'membershipStatus': None,
 'organizationRoles': None,
 'districts': None,
 'surveyQuestionResponses': None,
 'finderNumber': None,
 'biographyImageUrl': None,
 'primaryContact': None}
```

But if I run  

`van.get_person(id=101991312)`

[...]

PR #700, opened 2022-06-08
Attempt to filter expand_fields before stringifying them

shaunagm:
[Two years ago](https://github.com/move-coop/parsons/commit/fb54f5b849198cdb3dbb2a2e33603ce6e493fa2a), filter logic was added to the `get_person` VAN method to remove fields that didn't exist on particular databases. Unfortunately, it was added after the logic which turns `expand_fields` from an array into a string, leading to weirdness like this:

Screenshot from 2022-06-08 18-13-06

This in turn leads to 404 errors. Has `get_person` just always returned 404 errors for the last two years? Is no one using the `expand_fields` parameter? Have people been silently working around this? I am deeply curious.

### DIFF
--- a/parsons/ngpvan/people.py
+++ b/parsons/ngpvan/people.py
@@ -360,15 +360,15 @@ class People(object):
         id_type = '' if id_type in ('vanid', None) else f"{id_type}:"
         url += id_type + str(id)
 
-        expand_fields = ','.join([json_format.arg_format(f) for f in expand_fields])
-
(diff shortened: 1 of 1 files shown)


shaunagm:
Note: this may be related to #673

SorenSpicknall (approved):
Thanks for catching this. I'm also very curious about what people were doing up to this point, especially since it's such a heavily used connector.

PR #702, opened 2022-06-11
More flexible importing in parsons/__init__.py

cmc333333:
This implements the try-imports method @schuyler1d [described](https://github.com/move-coop/parsons/pull/632#issuecomment-1065494121) in a sustainable way, thereby allowing
```python
from parsons import ...
```
for users without all of the dependencies.

I'm a little torn on the "Table" callout -- it's definitely cooky to
make circular dependencies /easy/, but I figured this is the least
invasive approach.

I won't update the docs until this is on pypi!


SorenSpicknall:
Hey, this looks great! Does it still allow for imports in the existing longer format, so that people won't have to refactor existing code?

cmc333333:
Thanks @SorenSpicknall ! Both (e.g.)
```python
from parsons import VAN
```
or 
```python
from parsons.ngpvan.van import VAN
```
should continue to work -- is that what you were asking?

SorenSpicknall (approved):

PR #704, opened 2022-06-24
Fix error in VAN Target Export method.

cmc333333:
It looks like this method assumes VAN return a *list* of exported files
rather than a single target, but VAN currently only returns individual
targets: https://docs.ngpvan.com/reference/targetexportjobsexportjobid

They might've changed the API since this was last updated? In any event,
these tweaks return correct data in my recent testing!

### DIFF
--- a/parsons/ngpvan/targets.py
+++ b/parsons/ngpvan/targets.py
@@ -2,7 +2,6 @@
 
 from parsons.etl.table import Table
 import logging
-import json
 import requests
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


bzupnick (approved):
Looks good to me! I also like seeing the var names go from `j` and `i` to `response`. 👍

cmc333333:
Found another little optimization: rather than downloading the whole file into memory, we can let petl do its lazy loading thing.

SorenSpicknall (approved):

PR #724, opened 2022-07-28
New helper module with some reusable functions

shaunagm:
I would like for Parsons to include more reusable code which helps users with transformations and analysis. I have placed some examples in a new `etl_helpers` folder. 

I am looking for feedback on:

1) the overall goal of providing more helper code for users
2) the architecture here. is `etl_helpers` a decent name? I wanted to distinguish from the existing `utilities` which is primarily for code that helps Parsons contributors enhance Parsons connectors.
3) the specific helper functions I added

Once the overall approach is approved, and I have feedback on the specific helper functions, I will write some tests for them as well as incorporate them into the documentation.

Note: I copied two datetime functions from `utilities/datetime.py` which I didn't even know were there before. Eventually with enough notice we should remove the `utilities/datetime.py` file.

Checklist:

- [x] initial PR for review
- [ ] confirm architecture
- [ ] confirm specific functions
- [ ] add tests
- [ ] add docs


parsons/etl_helpers/conversions.py
```diff
@@ -0,0 +1,188 @@
...
+
+from dateutil.parser import parse
+import datetime
+import re
+
+
+# Datetime conversion functions
+
+
+def date_to_timestamp(value, tzinfo=datetime.timezone.utc):
```
alxmrs, line 15: Please add type annotations. For example: 

```
import typing as t
def date_to_timestamp(
    value: t.Union[int, str, datetime.datetime],
    tzinfo: datetime.tzinfo = datetime.timezone.utc) -> t.Optional[int]:


```

parsons/etl_helpers/conversions.py
```diff
@@ -0,0 +1,188 @@
...
+        tzinfo: datetime.timezone
+            `Optional`: Timezone for the datetime; defaults to UTC.
+    `Returns:`
+        Unix timestamp (int)
+    """
+
+    parsed_date = parse_date(value)
+
+    if not parsed_date:
+        return None
```
alxmrs, line 30: Either the function is misdocumented, or this is incorrect. What if you used `-1` as a sentinel values instead of `None`?

parsons/etl_helpers/conversions.py
```diff
@@ -0,0 +1,188 @@
...
+
+    if not parsed_date.tzinfo:
+        parsed_date = parsed_date.replace(tzinfo=tzinfo)
+
+    return int(parsed_date.timestamp())
+
+
+def parse_date(value, tzinfo=datetime.timezone.utc):
```
alxmrs, line 38: Add type annotations.

parsons/etl_helpers/conversions.py
```diff
@@ -0,0 +1,188 @@
...
+            'Unable to parse value; must be one of string or int or datetime, but got type '
+            f'{type(value)}')
+
+    if not parsed.tzinfo:
+        parsed = parsed.replace(tzinfo=tzinfo)
+
+    return parsed
+
+
+def timestamp_to_readable(value, tzinfo=datetime.timezone.utc, format_as='%Y-%m-%d %H:%M:%S'):
```
alxmrs, line 74: Why not use the ISO format as the default format?

alxmrs: i.e. the built-in one?

parsons/etl_helpers/conversions.py
```diff
@@ -0,0 +1,188 @@
...
+        str
+    """
+    datetime_obj = parse_date(value, tzinfo)
+    return datetime_obj.strftime(format_as)
+
+
+# Flatten contacts
+
+def get_primary_contact_from_nested(contact_list, get_first=True, selector=None):
+    """Extracts single contact value from list of dictionaries.
```
alxmrs, line 94: I see a shifting in dostring styles between this style (which I prefer) and the description on a new line, e.g.:
```
"""
Extract single contact ...
```

Which is the correct style?

parsons/etl_helpers/conversions.py
```diff
@@ -0,0 +1,188 @@
...
+
+    Multiple connectors, including Mobilize and Action Network, have a format for
+    contact info that looks like:
+
+    [{"primary": True, "number": "111-111-1111"}]
+
+    This helper method helps "flatten" the dictionary by returning the primary number (or,
+    if no primary number is found and get_first is True, the first number found).
```
alxmrs, line 102: Idea: what if we made the keys of the dictionary a tuple with all the numbers? e.g. 
```
result = {
   ('5554444444'): {'foo': 'bar'},
   ('4441234567', '7771234567'): {'bin': 'ban'},
   # ...
}
```

I bring this up, because there would be no exceptional cases here. If users wanted to get the first number, they would simply access the tuple and get the first element.

parsons/etl_helpers/conversions.py
```diff
@@ -0,0 +1,188 @@
...
+
+        for option in ["number", "postal_code", "address"]:  # try common options
+            if option in contact_list[0]:
+                selector = option
+                break
+
+        if not selector:  # if selector still not found, look in dict
+            dict_keys = list(contact_list[0].keys())
+            dict_keys.pop("primary")
+            selector = dict_keys[0]  # NOTE: this will break on dicts that have additional keys
```
alxmrs, line 134: This needs fixing / addressing.

alxmrs: e.g. it could be addressed with a try-catch block, or by raising an error msg.

parsons/etl_helpers/conversions.py
```diff
@@ -0,0 +1,188 @@
...
+    `Returns:`
+        str
+    """
+    datetime_obj = parse_date(value, tzinfo)
+    return datetime_obj.strftime(format_as)
+
+
+# Flatten contacts
+
+def get_primary_contact_from_nested(contact_list, get_first=True, selector=None):
```
alxmrs, line 93: How is this function used? What problem does it solve?

parsons/etl_helpers/conversions.py
```diff
@@ -0,0 +1,188 @@
...
+    `Returns:`
+        str
+    """
+
+    # Parse string and standardize to 10 digit format
+
+    value = value.split("x")[0]  # if there's an x in the old_number, use only numbers before the x
+
+    parsed_value = re.compile(r'\d+(?:\.\d+)?').findall(value)  # extracts digits only
+    value = "".join(parsed_value)
```
alxmrs, line 177: I find this a bit more readable / easier to maintain (regexes, in my view, should be avoided at all costs): 

```suggestion
    value = "".join([v for v in value if v.isdigit()])
```

alxmrs (commented):
Questions + feedback.

alxmrs:
> the overall goal of providing more helper code for users

What problems are you trying to solve?

shaunagm:
Hi Alex, thanks for these reviews,  lots of good catches here. Before you
do any additional pr reviews, let's touch base about the project, our goals
and community norms, etc. I'll email you tomorrow to set up a time.

parsons/etl_helpers/conversions.py
shaunagm: ISO is a little less readable, though certainly more readable than unix timestamps! But it might be worth the tradeoff to use a standardized value.

parsons/etl_helpers/conversions.py
shaunagm: I don't think we have a preference. I was going to say "let's go with what Python prefers" but it looks like Python itself [doesn't have a preference](https://peps.python.org/pep-0257/). I know it's a little inconsistent but I don't think it's a big deal to switch back and forth within parsons, and that's one less thing that PR contributors have to worry about.

shaunagm:
[Adding helper utilities to Parsons](https://trello.com/c/aBOcEQRh/5-adding-helper-utilities-to-parsons)

shaunagm:
Closing this but linking it from #836 so that the code/discussion here isn't lost.

PR #725, opened 2022-08-10
Don't auto-grab session token if it's not passed in

sjwmoveon:
It turns out that trying to grab a session token from the environment variable can break things when used in concert with Zappa asynchronous processing - the access key and secret key passed in may not match the temporary token generated for the asynchronous process. Thus, this change forces the token to be explicitly set instead of automatically pulling from the environment variable.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,3 +1,11 @@
+---
+name: Bug Report
+about: Tell us about a problem with the project's current functionality
+title: "[Bug]"
+labels: bug
+assignees: ''
+---
+
 <!--- Provide a general summary of the bug you're experiencing in the Title above -->
 
 
--- a/.github/ISSUE_TEMPLATE/proposed-feature-addition.md
+++ b/.github/ISSUE_TEMPLATE/proposed-feature-addition.md
@@ -1,3 +1,11 @@
+---
+name: Proposed Feature/Addition
+about: Suggest an addition to Parsons
+title: "[Feature/Addition]"
+labels: enhancement
+assignees: ''
+---
+
 <!--- Provide a general summary of the proposed Parsons addition in the Title above -->
 
 
--- a/docs/auth0.rst
+++ b/docs/auth0.rst
@@ -1,0 +1,38 @@
+Auth0
+=========
+
+********
+Overview
+********
+
+`Auth0 <https://auth0.com/>`_ is an authentication and authorization platform. This Parsons integration with the `Auth0 Management API <https://auth0.com/docs/api/management/v2>`_ supports fetching and deleting user records.
+
+**********
+Quickstart
+**********
+
+To instantiate the Auth0 class, you can either store your Auth0 API client ID, client secret, and domain as environment variables (``AUTH0_CLIENT_ID``, ``AUTH0_CLIENT_SECRET``, and ``AUTH0_DOMAIN``, respectively) or pass in your client ID, client secret, and domain as arguments:
+
+.. code-block:: python
+
+   from parsons import Auth0
+
+   # First approach: Use API credentials via environmental variables
+   auth0 = Auth0()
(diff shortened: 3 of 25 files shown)


shaunagm:
Hi @sjwmoveon! Thanks for submitting this PR. I don't use the S3 connector so apologies if I'm misunderstanding the original code and/or the change you're suggesting. If you can provide a bit more detail about how the current code is breaking for you and how your code fixes that, I'd really appreciate it. :)

In particular, I would like to understand:

The code as currently written always tries to get `aws_session_token` from the environment but only when it hasn't been passed in as a parameter. You write: "this change forces the token to be explicitly set instead of automatically pulling from the environment variable", but how is it being explicitly set? Is this something that Zappa is handling independently? 

I'm also curious about where the `aws_session_token` in your environment is coming from. Is it an old/out of date value? Why isn't it just `None`? And what happens when you run the code - how does it "break"?

sjwmoveon:
>  "this change forces the token to be explicitly set instead of automatically pulling from the environment variable", but how is it being explicitly set? Is this something that Zappa is handling independently
Sorry, I was being imprecise.  The change requires the token to be passed in, even if it's set as an environment variable.  In my use case, I don't want to use the token at all, since it's associated with a different user or role than the credentials I'm passing in. If the token is set as an environment variable, I have no way to _not_ use it, even though I don't want to use it at all. I tried unsetting `aws_session_token` directly with `del os.environ['AWS_SESSION_TOKEN']`, but it didn't appear to work, so I needed some other way to not use the token..

As for the error:
As described [here](https://github.com/zappa/Zappa#asynchronous-task-execution), with asynchronous task execution, Zappa spins up a completely separate Lambda instance.  The session token is generated as part of that process, and (as far as I can tell) is based on the execution role of the Lambda function.
So when the asynchronous task is running, it has the `aws_session_token` set even though there wasn't one originally.  I'm not sure why it isn't just "None".
In any case, the token doesn't match the access key ID and secret access key that I was being passed in, resulting in an InvalidTokenError: `An error occurred (InvalidToken) when calling the PutObject operation: The provided token is malformed or otherwise invalid`.  It's not that the token is actually invalid, but it doesn't match the credentials being passed in.

I hope that helps! Let me know if you have any more questions.

shaunagm:
Gotcha. Is there a known value for `aws_session_token`? You originally wrote "this change forces the token to be explicitly set instead of automatically pulling from the environment variable" but how is it being explicitly set? Is it being passed in on initialization as `aws_session_token` or through some other method or not at all?

sjwmoveon:
Yes, passed in on initialization.  If someone wants to set up an AWSConnection object using an access token, they will need to pass it in.

shaunagm:
Thanks for clarifying - I just realized I missed you saying all this in your previous comment.

If you look at [the current code](https://github.com/move-coop/parsons/blob/main/parsons/aws/s3.py#L27), we only check the environment for `AWS_SESSION_TOKEN` when `aws_session_token` is not passed in on initialization (when it is `None`). So passing that variable in should already work to bypass the environment check. If that's not what's happening, we should debug. I'm happy to pair with you to figure out what's going wrong.

shaunagm:
Hi @sjwmoveon, want to do a 1:1 to talk through this PR? It looks like the commit history has gotten tangled.

sjwmoveon:
Whoops, you're completely right!  I can start a new PR with just the S3 changes, also happy to do a 1:1 if it'd be helpful.

sjwmoveon:
"If you look at [the current code](https://github.com/move-coop/parsons/blob/main/parsons/aws/s3.py#L27), we only check the environment for AWS_SESSION_TOKEN when aws_session_token is not passed in on initialization (when it is None). So passing that variable in should already work to bypass the environment check."

Right, in my use case I don't want the token to be used at all (I want it to be None).  Right now there's no way to force the token to None if the environment variable is set - I can override to a different value, but I can't just erase it entirely.

PR #728, opened 2022-08-12
Convert None to "" in convert_columns_to_str

Jason94:
The function `parsons.Table.convert_columns_to_str()` converts `None` to `"None"`. I think that converting it to the empty string is more intuitive and more helpful. It does make it harder to deserialize the result of `convert_columns_to_str`, but the `str` function is not generally deserializable anyway.

### DIFF
--- a/parsons/etl/etl.py
+++ b/parsons/etl/etl.py
@@ -181,11 +181,16 @@ class ETL(object):
 
         cols = self.get_columns_type_stats()
 
+        def str_or_empty(x):
+            if x is None:
+                return ""
+            return str(x)
+
         for col in cols:
             # If there's more than one type (or no types), convert to str
             # Also if there is one type and it's not str, convert to str
             if len(col['type']) != 1 or col['type'][0] != 'str':
-                self.convert_column(col['name'], str)
+                self.convert_column(col['name'], str_or_empty)
 
         return self
 


shaunagm:
This seems straightforward. I wish there was a way to check with users, though! Can we think of any case or workflow in which someone might be relying on getting "None" vs an empty string?

Jason94:
> 

In theory I could imagine a case where someone does something like this:
``` python
strs = table.convert_columns_to_str()
for row in strs.to_dict():
    if row['col'] == "None":
        do_something()
```

But I think that is somewhat unlikely. The purpose of this function seems to be for presentation, not for logic. And there is no reason you couldn't check for `None` values before calling `convert_columns_to_str()`. So you could say that relying on this behavior doesn't allow you to do anything you couldn't do before; just do it in a more confusing fashion. (In fact the above code has the problem that if the original table had "None" values in 'col' they would call the function, which is probably not what you would want.)

You're right to be cautious because it is an API breaking change. But the existing behavior is IMO niche and confusing enough that I think it's warranted. Certainly worth noting on release though, at the very least.

shaunagm:
I've created a 'breaking change' label to help us flag these kinds of things in releases. With that handled, this looks good.

shaunagm (approved):

PR #731, opened 2022-08-23
Update from 2.8.5 to 2.9.3

shaunagm:
Opening this PR on behalf of @Jason94. We've had tons of trouble with psycopg2-binary, maybe updating to the most recent version will fix some of them.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ boto3==1.17.98
 boxsdk==2.10.0
 civis==1.14.2
 slackclient==1.3.0
-psycopg2-binary==2.8.5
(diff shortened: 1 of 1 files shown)


SorenSpicknall (approved):
I'm generally in favor of keeping deps updated, but has this been tested locally to see if it resolves any of the installation issues we're going after?

shaunagm:
I think we'd have to run another install fest to see that - most of the people who had problems fixed them.

Alternatively, I could investigate what it would take (time and money-wise) to test the install process on 3-5 setups every X days (or every time we merge to master) and then see whether this helps us successfully execute a parsons install on a setup that isn't currently working.

PR #732, opened 2022-08-24
Always use utf-8 encoding writing to RS s3

Jason94:
I think I figured out why Redshift.copy gives an encoding error very frequently, like this:
```
Traceback (most recent call last):
  File "C:\Users\JasonWalker\Programming\Workspace\one-off-scripts\mobilize\events.py", line 28, in <module>
    main()
  File "C:\Users\JasonWalker\Programming\Workspace\one-off-scripts\mobilize\events.py", line 24, in main
    rs.copy(events, 'indivisible_test.org_events', if_exists='drop', compupdate=False)
  File "c:\users\jasonwalker\programming\lib\parsons\parsons\databases\redshift\redshift.py", line 539, in copy
    key = self.temp_s3_copy(tbl, aws_access_key_id=aws_access_key_id,
  File "c:\users\jasonwalker\programming\lib\parsons\parsons\databases\redshift\rs_copy_table.py", line 144, in temp_s3_copy
    local_path = tbl.to_csv(temp_file_compression='gzip')
  File "c:\users\jasonwalker\programming\lib\parsons\parsons\etl\tofrom.py", line 137, in to_csv
    petl.tocsv(self.table,
  File "C:\Users\JasonWalker\Programming\lib\petl\petl\io\csv.py", line 108, in tocsv
    tocsv_impl(table, source=source, encoding=encoding, errors=errors,
  File "C:\Users\JasonWalker\Programming\lib\petl\petl\io\csv_py3.py", line 44, in tocsv_impl
    _writecsv(table, source=source, mode='wb', **kwargs)
  File "C:\Users\JasonWalker\Programming\lib\petl\petl\io\csv_py3.py", line 62, in _writecsv
    writer.writerow(row)
  File "C:\Python310\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u2028' in position 757: character maps to <undefined>
```
Redshift.copy winds up calling petl.to_csv , the result of which is copied to s3 and then transferred to Redshift. The acceptinvchars argument to Redshift.copy (@Soren Spicknall has pointed this out past times I've had this problem) does insert the acceptinvchars statement to the SQL to load the CSV into Redshift. However, the encoding for writing the temp .csv is not specified, so it fails if it encounters non-ascii characters.

### DIFF
--- a/parsons/databases/redshift/redshift.py
+++ b/parsons/databases/redshift/redshift.py
@@ -390,7 +390,7 @@ class Redshift(RedshiftCreateTable, RedshiftCopyTable, RedshiftTableUtilities, R
              columntypes=None, specifycols=None, alter_table=False, alter_table_cascade=False,
              aws_access_key_id=None, aws_secret_access_key=None, iam_role=None,
              cleanup_s3_file=True, template_table=None, temp_bucket_region=None,
-             strict_length=True):
+             strict_length=True, csv_encoding='utf-8'):
         """
         Copy a :ref:`parsons-table` to Redshift.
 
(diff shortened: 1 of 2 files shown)


Jason94:
@SorenSpicknall mentioned we should add a flag to allow overriding the encoding.

Jason94:
I've added a `csv_econding` parameter to both the Redshift.copy method and the to_s3_csv method. Both default to 'utf-8'.

SorenSpicknall (approved):

Issue #706, opened 2022-07-01
ActionNetwork.update_person() does not update the person


Maggiedp:
When I run update_person(), it returns the person but does not actually update the value for that person.

## To Reproduce
```
from parsons import ActionNetwork
an = ActionNetwork()
person_id='d52040b6-0168-401f-8e7a-b49b2b4e15a1'
person_update=an.update_person(person_id, custom_fields={'2022Letter': 'test'})
person_update=an.update_person(person_id, family_name='test')
person_update['family_name']
person_update['custom_fields']['2022Letter']

```
person_update['family_name'] returns 'Charles' and person_update['custom_fields']['2022Letter'] returns a keyerror because 2022Letter was not added to the custom fields. It does successfully return the json for the correct person, it just doesn't update the person.

## Your Environment
* Version of Parsons used (if you know): parsons==0.19.0
* Operating System and version (desktop or mobile): Ubuntu 21.10
* Is there information about my Action Network group that would be helpful here? I'm just using the key from the API page

## Additional Context
I'm using the lower level function api.get_request() with the Action Network connector and it's working fine. 

## Priority
I think this is a high priority, since it's a bug that makes this function not work but I'm not sure how popular this function is. I'm working on workarounds that don't use parsons for updating people in AN.

Maggiedp:
I think the json for PUT needs to go into the data parameter of the request. Right now it's going in the [json parameter](https://github.com/move-coop/parsons/blob/683d4b02985bcaa8b3a0d7f5d00648e375582444/parsons/action_network/action_network.py#L241)

`person_update = an.api.put_request(url=f"{an.api_url}/people/{person_id}", data=json.dumps(data), success_codes=[204, 201, 200])` seems to work correctly

SorenSpicknall:
Thanks for reporting this, @Maggiedp - we're trying to get some eyes on it within the contributor community. Feel free to put in a PR for a fix yourself, too, if you want!

ydamit:
Chiming in to report that I was just told by support that the "right" way to update a record is to use the same "helper" endpoint we're already using in `add_person()`. This supposedly only works with POST requests, but if you supply an email or mobile number of an existing record, it'll update that record instead of creating a new one.

In my investigation, it did seem that it was possible to update _some_ data with a PUT request to the record-specific endpoint. Specifically, I was able to update the value of a custom field (that already had a value on the record) this way. And, interestingly, I was able to do that without nesting `custom_field` inside of a `person` item.

[...]

PR #734, opened 2022-08-25
Add upsert_person method to ActionNetwork

shaunagm:
Also adds deprecation warnings to update_person and add_person.

Addresses #706 

cc @maggiedp

### DIFF
--- a/parsons/action_network/action_network.py
+++ b/parsons/action_network/action_network.py
@@ -90,10 +90,14 @@ class ActionNetwork(object):
         """
         return self.api.get_request(url=f"people/{person_id}")
 
-    def add_person(self, email_address=None, given_name=None, family_name=None, tags=None,
-                   languages_spoken=None, postal_addresses=None, mobile_number=None,
-                   mobile_status='subscribed', **kwargs):
+    def upsert_person(self, email_address=None, given_name=None, family_name=None, tags=None,
+                      languages_spoken=None, postal_addresses=None, mobile_number=None,
+                      mobile_status='subscribed', **kwargs):
         """
+        Creates or updates a person record. In order to update an existing record instead of
+        creating a new one, you must supply an email or mobile number which matches a record
+        in the database.
+
         `Args:`
             email_address:
                 Either email_address or mobile_number are required. Can be any of the following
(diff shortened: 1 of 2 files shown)


parsons/action_network/action_network.py
```diff
@@ -200,8 +204,25 @@ def add_person(self, email_address=None, given_name=None, family_name=None, tags
         logger.info(f"Entry {person_id} successfully added to people.")
```
ydamit, line 204: Not a huge deal, but should probably change this language to reflect the upsert functionality.

parsons/action_network/action_network.py
```diff
@@ -200,8 +204,25 @@ def add_person(self, email_address=None, given_name=None, family_name=None, tags
...
+        self.upsert_person(email_address=email_address, given_name=given_name,
+                           family_name=family_name, languages_spoken=languages_spoken,
+                           postal_addresses=postal_addresses, mobile_number=mobile_number,
+                           mobile_status=mobile_status, **kwargs)
+
     def update_person(self, entry_id, **kwargs):
         """
+        Updates a person's data in Action Network. WARNING: this endpoint has been deprecated in
+        favor of upsert_person and may not work properly.
+
```
ydamit, line 225: Now that we know that the person-specific endpoint _does_ work with `PUT` requests, do we still think that we want to deprecate this? As an alternative, we could just remove the tagging functionality. For instance, NGPVAN has both upsert and update methods. On the other hand, maybe keeping things simpler is a better approach?

ydamit (approved):
Just a couple of thoughts: an open-ended question and a suggestion for updating logging messages, neither of which should prevent this from getting approved.

parsons/action_network/action_network.py
shaunagm: Do you think the person-specific endpoint is easier to use? If so, we should probably keep it.

ydamit: I have no reason to believe it is, but I'm afraid I also don't have any reason to believe it _isn't_. The only thing I know for sure is that the method we're changing to upsert (i.e. the one that uses the helper endpoint) is _definitely_ used, and I have no notion of whether `update_person()` is used at all. Honestly, I think it matters very little, so I was hoping you'd have a structural, conventional, or even esthetic reason to help make the call.

shaunagm: Hmm, yeah. I think conventionally/aesthetically you minimize duplicate calls, but otoh we don't want to break things for people.

PR #703, opened 2022-06-17
Added tests for SendMail notification connector

AndrewRook:
Hi!

I added tests for the `SendMail` class, as part of addressing #635. I installed pytest-cov and got the following coverage results (with the tests added):

```bash
$ pytest test/ --cov=parsons/notifications/ --cov-report=term-missing
<full output snipped>
Name                                Stmts   Miss  Cover   Missing
-----------------------------------------------------------------
parsons/notifications/__init__.py       0      0   100%
parsons/notifications/gmail.py         34     15    56%   29, 32, 41-42, 67-85
parsons/notifications/sendmail.py     116      0   100%
parsons/notifications/slack.py         73     11    85%   28, 114, 145-147, 195-197, 220-224, 229
parsons/notifications/smtp.py          34      8    76%   37-41, 59-62
-----------------------------------------------------------------
```

I was going to add documentation for `SendMail` as discussed in #635, but when I looked at it closely it seems like this class is more like an abstract base class than something that users should use directly. You have to inherit from it and override `_send_message` in order for it to work, for one thing. So I figured I'd just PR the tests and then ask — should there be user-facing documentation for this class?

PR #735, opened 2022-08-27
Simple docs for the SMTP connector

### LINKED ISSUE #635: Add docs for SMTP and Sendmail connectors
In working on PR #634 I noticed that there aren't any docs for our SMTP and Sendmail notifications connectors, which means they're pretty undiscoverable. We should add docs here: https://github.com/move-coop/parsons/blob/master/docs/notifications.rst

Note that SMTP has a test file but Sendmail doesn't; we may want to also check that both connectors have good test coverage before adding their docs and making them discoverable.

AndrewRook:
I added some really simple documentation for the STMP class. It's based on the docs for the Gmail connector, with tweaks where appropriate. I have fairly limited knowledge of both SMTP and the class itself, so if I'm missing anything obvious please let me know!

This, along with #703 and #717, closes #635.

### DIFF
--- a/docs/notifications.rst
+++ b/docs/notifications.rst
@@ -184,10 +184,10 @@ To initialize the SMTP class you will need to tell it how to connect to the SMTP
 
 .. note::
   Environment Variables
-  * Instead of passing in values to initialize an instance of the SMTP class, you can set environment
-    variables to hold the values. The names of the environment variables are the names of the arguments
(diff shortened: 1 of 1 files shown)


shaunagm (approved):
This looks great! My apologies that it took so long for me to review.

AndrewRook:
No worries, I'm not in a rush 😄. Thanks!

PR #737, opened 2022-08-30
feat: scytl connector

agreenspan24:
[Scytl](https://www.scytl.com/resources-and-references/customers/u-s-elections/), or Clarity Elections, is a company that creates a tool for publishing election results in real-time. It's used in the U.S. by several states and over 800 counties, including Georgia, Colorado, Arkansas, and Dallas County, Texas.

For each participating election administrator, they publish a site with results that can be published on election night. Unfortunately, while that site contains downloadable data, that data is formatted in a complex way, making it difficult for developers to fetch election results. In general, their results either come zipped in either an unformatted text file or a complex XML document. Summary results come as a zipped CSV, but just contain top-line results. The JSON that powers the site results is even more complicated.

This connector provides methods to download the latest election results from their site and formats them into readable lists of dictionaries, which can easily be converted into a Parsons Table or Pandas dataframe.

Because this connector is can be useful for live reporting, it also contains a polling feature. As long as the class is instantiated, it will only fetch results that are new since the previous fetch of that method. To skip this feature, set force_update to true on any of the fetch methods.

Note: please do not confuse with [Clarity Campaign Labs](https://www.claritycampaigns.com/)

Example microsite: https://results.enr.clarityelections.com/CO/105975/web.276603/


shaunagm:
Hi @agreenspan24! Thanks so much for this contribution. I haven't had a chance to go through the whole PR yet but I'm hoping to get to it and leave a review later this week.

I appreciate you having thorough tests for you connector! I am a little hesitant about the 25-ish test files though. Are they all necessary? Might they cause issues for long-term maintainability if there are formatting changes?

agreenspan24:
> Hi @agreenspan24! Thanks so much for this contribution. I haven't had a chance to go through the whole PR yet but I'm hoping to get to it and leave a review later this week.
> 
> I appreciate you having thorough tests for you connector! I am a little hesitant about the 25-ish test files though. Are they all necessary? Might they cause issues for long-term maintainability if there are formatting changes?

Hi Shauna! Makes total sense - was trying to keep the 18 county files to match the real results, but I think you're right to reduce them. I cut out all but two and adjusted the tests - run a good bit faster now too!

For maintainability, those files are just direct copies of real election results as Scytl holds them. If they change their format, I'm hoping we can just drop newer versions of those files into the tests and adjust the parser accordingly. But let me know your thoughts!

docs/scytl.rst
```diff
@@ -0,0 +1,40 @@
...
+
+For each participating election administrator, they publish a site with results that can be published on election night. Unfortunately, while that site contains downloadable data, that data is formatted in a complex way, making it difficult for developers to fetch election results. In general, their results either come zipped in either an unformatted text file or a complex XML document. Summary results come as a zipped CSV, but just contain top-line results. The JSON that powers the site results is even more complicated.
+
+This connector provides methods to download the latest election results from their site and formats them into readable lists of dictionaries, which can easily be converted into a Parsons Table or Pandas dataframe.
+
+Because this connector is can be useful for live reporting, it also contains a polling feature. As long as the class is instantiated, it will only fetch results that are new since the previous fetch of that method. To skip this feature, set force_update to true on any of the fetch methods.
+
```
shaunagm, line 15: The vast majority of our connectors (maybe all of them, other than this one?) require authentication. It seems like this doesn't because the data source is public? It would be useful to call that out so that a reader doesn't think that authentication is required but not documented. You can do something like
```

.. note::
  Authentication
    All endpoints for Scytl are public, and do not need authentication.
```

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,561 @@
...
+            ex: GA
+        election_id: str
+            The numeric identifier for the election found in the url of the election's website.
+            ex: "114729"
+        county: str (optional)
+            The name of the county publishing the results.
+            ex: Clarke
+    """
+
+    def __init__(self, state: str, election_id: str, county=""):
```
shaunagm, line 74: It's unusual for anything other than auth info to get passed in to a Parsons connector, although it certainly does happen, and it makes sense to pass in `state` and `election_id` here so they don't have to be passed in with `get_summary_results`, etc. 

That said, `county` in particular confuses me. How does the county parameter here fit with the counties passed in with `get_detailed_results_for_participating_counties`? Does this county parameter get used at any other point? It seems like the other two are always state level but maybe I'm misunderstanding. Does passing in the county on init mean that `get_summary_results` and `get_county_results` apply only to the county?

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,561 @@
...
+        self.administrator = f"{self.state}/{self.county}" if self.county else self.state
+        self.election_id = election_id
+
+        self.previous_summary_version_num = None
+        self.previous_details_version_num = None
+        self.previous_county_details_version_num = None
+        self.previous_county_details_list = None
+        self.previously_fetched_counties = set([])
+
+    def _parse_date_to_utc(self, input_dt):
```
shaunagm, line 87: Can you document the private methods? This is not for the users, since they won't display in our docs, but very helpful for future maintainability/readability.

shaunagm (changes requested):
Overall looks good. Mostly docs issues here.

shaunagm:
@agreenspan24 I think your test fixes look good!

parsons/scytl/scytl.py
agreenspan24: Updated!

parsons/scytl/scytl.py
agreenspan24: Yeah, the setup is kinda confusing, would love your help navigating it. 

Essentially, they have a concept of an election (at the state or county level), which can have summary results and detailed results. The detailed results show results by county for state level elections, while they show precinct results for a county-level election.

State-level elections can also have sub-elections in each county, and link to each of those county elections, which each have summary and detailed results. So, a county-level election won't have participating counties and the method won't make sense.

The main reason I wanted to add the state and election_id at the class level instead of the method is so the class could track the previous version- I have a feeling people would use this for polling and I want to make that as easy as possible for them to see if results have updated.

One idea might be to split this up into a `StateScytlElection` and `CountyScytlElection` class, which would have some of the same implementation, but not all of it. What do you think?

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+import requests
+import urllib.error
+import urllib.request
+import xml.etree.ElementTree as ET
+from datetime import datetime
+from dateutil.parser import parse as parsedate
+from pytz import timezone
+from io import BytesIO, StringIO
+
+clarity_url = 'https://results.enr.clarityelections.com/'
```
alxmrs, line 13: constants should be IN_APP_CAPS.

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+
+get_version_url = \
+    clarity_url + '{administrator}/{election_id}/current_ver.txt'
+summary_csv_url_template = \
+    clarity_url + '{administrator}//{election_id}/{version_num}/reports/summary.zip'
```
alxmrs, line 18: Is the double slash (`//`) intentional?

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+    "PST": "UTC-8",
+    "PDT": "UTC-7",
+    "AKST": "UTC-9",
+    "AKDT": "UTC-8",
+    "HST": "UTC-10",
+    "HDT": "UTC-9",
+}
+
+
+class CountyDetails:
```
alxmrs, line 42: This could be a dataclass.

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+        county_update_date: datetime = None
+    ):
+        self.state = state
+        self.county_name = county_name
+        self.county_election_id = county_election_id
+        self.county_version_num = county_version_num
+        self.county_update_date = county_update_date
+
+
+class Scytl(object):
```
alxmrs, line 58: Why the `object` is correct (and the convention of the codebase), it also is unnecessary.

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+        self.administrator = f"{self.state}/{self.county}" if self.county else self.state
+        self.election_id = election_id
+
+        self.previous_summary_version_num = None
+        self.previous_details_version_num = None
+        self.previous_county_details_version_num = None
+        self.previous_county_details_list = None
+        self.previously_fetched_counties = set([])
+
+    def _parse_date_to_utc(self, input_dt: str):
```
alxmrs, line 87:
```suggestion
    def _parse_date_to_utc(self, input_dt: str) -> datetime.datetime:
```

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+                data = remote.read()
+                zipdata.write(data)
+
+            zf = zipfile.ZipFile(zipdata)
+
+            with zf.open(file_name) as input:
+                return input.read()
+
+        except urllib.error.HTTPError as e:
+            print('HTTPError: {} {}'.format(e.code, zipfile_url))
```
alxmrs, line 156: Better to use a logger for the class. -- Are you sure you don't want to raise an error? Should the pipeline continue if it can't read the data?

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+                The url where the zip file can be found
+            election_id: str
+                The expected name of the file in the zipfile to read
+        `Returns`:
+            bytes
+            The unzipped file as bytes
+        """
+
+        try:
+            zipdata = BytesIO()
```
alxmrs, line 144: Consider wrapping this in a context manager to automatically handle closing and flushing.

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+
+            with urllib.request.urlopen(zipfile_url) as remote:
+                data = remote.read()
+                zipdata.write(data)
```
alxmrs, line 148: Is a flush prudent here?

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+
+            with zf.open(file_name) as input:
+                return input.read()
+
+        except urllib.error.HTTPError as e:
+            print('HTTPError: {} {}'.format(e.code, zipfile_url))
+
+    def _get_latest_counties_scytl_info(
+        self, state: str, election_id: str, version_num: str
+    ) -> dict[str, CountyDetails]:
```
alxmrs, line 160:
```
import typing as t

t.Dict[str, CountryDetails]
```

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+                county_update_date
+            )
+
+            county_dict[source_county_name] = county_details
+
+        return county_dict
+
+    def _parse_county_xml_data_to_precincts(
+        self, county_data: bytes, county_details: CountyDetails
+    ) -> list[object]:
```
alxmrs, line 210:
```
import typing as t
-> t.List[t.Dict]:
```

alxmrs: `object` != `dict`

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+
+        root = tree
+
+        for child in root:
+
+            if child.tag == 'VoterTurnout':
+                precincts = child[0]
+
+                for precinct in precincts:
+                    data = (precinct.attrib)
```
alxmrs, line 239: parens are redundant.

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+                    cand_party = source_cand_data.get('party')
+
+                    for vote_type in choice:
+                        vote_type_label = vote_type.attrib['name']
+
+                        for precinct in vote_type:
+                            precinct_name = precinct.attrib['name']
+                            cand_votes[precinct_name] = int(precinct.attrib['votes'])
+
+                            precinct_turnout = precinct_dict.get(precinct_name) or {}
```
alxmrs, line 273:
```suggestion
                            precinct_turnout = precinct_dict.get(precinct_name, {})
```

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+        for child in root:
+
+            if child.tag == 'VoterTurnout':
+                precincts = child[0]
+
+                for precinct in precincts:
+                    data = (precinct.attrib)
+                    name = data.get('name')
+
+                    precinct_info = {
```
alxmrs, line 242: It looks like a lot of precinct info doesn't make it to the final function return. Am I missing something?

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+                    cand_party = source_cand_data.get('party')
+
+                    for vote_type in choice:
+                        vote_type_label = vote_type.attrib['name']
+
+                        for county in vote_type:
+                            county_name = county.attrib["name"]
+                            cand_votes[county_name] = int(county.attrib["votes"])
+
+                            county_turnout = county_dict.get(county_name) or {}
```
alxmrs, line 353:
```suggestion
                            county_turnout = county_dict.get(county_name, {})
```

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+                                'timestamp_last_updated': timestamp
+                            }
+
+                            county_votes.append(result)
+
+        return county_votes
+
+    def _fetch_and_parse_summary_results(
+        self, administrator: str, election_id: str, version_num: str, county=''
+    ) -> list[object]:
```
alxmrs, line 376: See typing notes above.

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+
+        zip_bytes = self._parse_file_from_zip_url(summary_csv_zip_url, 'summary.csv')
+
+        if not zip_bytes:
+            return []
+
+        string_buffer = StringIO(zip_bytes.decode('latin-1'))
+        csv_data = csv.DictReader(string_buffer, delimiter=",")
+
+        data = list(
```
alxmrs, line 408: I think a list comprehension would be more readable here.

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+        summary_csv_zip_url = summary_csv_url_template.format(
+            administrator=administrator, election_id=election_id, version_num=version_num
+        )
+
+        zip_bytes = self._parse_file_from_zip_url(summary_csv_zip_url, 'summary.csv')
+
+        if not zip_bytes:
+            return []
+
+        string_buffer = StringIO(zip_bytes.decode('latin-1'))
```
alxmrs, line 405: recommended: use context managers to ensure that these get cleaned up on function return.

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+                    'total_precincts': x.get('num Precinct total'),
+                    'candidate_name': x.get('choice name'),
+                    'candidate_party': x.get('party name'),
+                    'recorded_votes': x.get('total votes')
+                },
+                csv_data))
+
+        return data
+
+    def get_summary_results(self, force_update=False):
```
alxmrs, line 427:
```
import typing as t
-> t.List[Dict]:
```

parsons/scytl/scytl.py
```diff
@@ -0,0 +1,664 @@
...
+
+        self.previous_details_version_num = version_num
+
+        return parsed_data
+
+    def get_detailed_results_for_participating_counties(
+        self,
+        county_names: list[str] = None,
+        force_update=False
+    ):
```
alxmrs, line 559:
```
-> t.List[t.Union[t.Dict, str]]:
```

test/test_scytl/test_scytl.py
```diff
@@ -0,0 +1,253 @@
...
+import os
+import requests_mock
+import csv
+from parsons.scytl import Scytl, scytl
+
+TEST_STATE = 'GA'
+TEST_ELECTION_ID = '114729'
+TEST_VERSION_NUM = '296262'
+
+_dir = os.path.dirname(__file__)
```
alxmrs, line 11: `_DIR`

test/test_scytl/test_scytl.py
```diff
@@ -0,0 +1,253 @@
...
+            administrator=TEST_STATE, election_id=TEST_ELECTION_ID, version_num=TEST_VERSION_NUM
+        )
+
+        with open(f'{_dir}/114729_summary.zip', 'r') as summary:
+            m.get(mock_url, body=summary)
+
+        scy = Scytl(TEST_STATE, TEST_ELECTION_ID)
+        result = scy.get_summary_results()
+
+        with open(f'{_dir}/114729_summary_expected.csv', 'r') as expected:
```
alxmrs, line 30: I find it hard to understand if the data is correct since it's hard to compare the source and destination data files in this diff.

test/test_scytl/test_scytl.py
```diff
@@ -0,0 +1,253 @@
...
+
+                expectedResultRow['counties_reporting'] = \
+                    expectedResultRow['counties_reporting'] or None
+                expectedResultRow['total_counties'] = expectedResultRow['total_counties'] or None
+
+                self.assertDictEqual(row, expectedResultRow)
+
+    @requests_mock.Mocker()
+    def test_get_summary_results_skips_if_no_version_update(self, m: requests_mock.Mocker):
+        self._mock_responses(m)
```
alxmrs, line 44: Is it possible to move this to the `setUp` method?

test/test_scytl/test_scytl.py
```diff
@@ -0,0 +1,253 @@
...
+
+        scy = Scytl(TEST_STATE, TEST_ELECTION_ID)
+
+        result = scy.get_summary_results()
+
+        assert result is not None
```
alxmrs, line 50: prefer unittest asserts over the `assert` operation since they give more specific error messages.

test/test_scytl/test_scytl.py
```diff
@@ -0,0 +1,253 @@
...
+        self._mock_responses(m)
+
+        mock_url = scytl.detail_xml_url_template.format(
+            administrator=TEST_STATE, election_id=TEST_ELECTION_ID, version_num=TEST_VERSION_NUM
+        )
+
+        with open(f'{_dir}/114729_detailxml.zip', 'r') as detailxml:
+            m.get(mock_url, body=detailxml)
+
+        scy = Scytl(TEST_STATE, TEST_ELECTION_ID)
```
alxmrs, line 71: I bet a lot of this could go into either the `setUp` method or a class test utility method.

alxmrs (commented):
A few notes.

parsons/scytl/scytl.py
agreenspan24: Another interesting idea. If @shaunagm wants to start using the typing package, that works for me, but I've just added normal type annotations for all the functions.

parsons/scytl/scytl.py
agreenspan24: Good catch, might as well add the rest to the dictionary.

test/test_scytl/test_scytl.py
agreenspan24: Yeah, looks like it, with a bit of hackyness.

test/test_scytl/test_scytl.py
agreenspan24: Yeah that's a reasonable point. I think it's important to use the zip files since unzipping is a big part of what we're testing.

I guess it's somewhat helpful to just read the expected CSV and make sure all the data looks good and reasonable, but not ideal.

parsons/scytl/scytl.py
agreenspan24: I'm not really sure how context managers work, how would they help here?

parsons/scytl/scytl.py
agreenspan24: I'm not actually too familiar here with the underlying implementation. How would a flush work and be useful here? Do I just need to wrap the zipdata in a `with` statement?

parsons/scytl/scytl.py
agreenspan24: Letting it throw, catching in certain cases.

agreenspan24 (commented):
@alxmrs Thanks for the review! Still pretty new to proper python conventions, so I appreciate the deeper look! A couple of ideas here sound interesting to me, like typing and dataclasses, but I don't want to introduce new patterns without approval from maintainers, so @shaunagm let us know your thoughts!

parsons/scytl/scytl.py
alxmrs: Sorry, I am using a really wonky name. By contextmanager, I mean a `with` expression. 

All IO in python inherits from the IOBase, which supports automatically closing files / freeing memory if you wrap things in `with` statements. 

https://docs.python.org/3/library/io.html#io.IOBase

Here, I think you have a memory leak, since `close()` is never called on `zipdata`.

parsons/scytl/scytl.py
alxmrs: While I agree that wrapping zipdata with `with` is needed (see above), you may also want to call flush. 

Flush tells the file object to actually write the data from the buffers, which usually happens when the file is closed. 
Usually, if you want to access the data sooner, it's prudent to explicitly call flush.

parsons/scytl/scytl.py
shaunagm: I'm of two minds about this. Turning it into a dataclass will improve readability for some, but may make the code harder to read and reason about for many who are newer to Python and unfamiliar with dataclasses. Perhaps we can switch to using a dataclass if you're up for it @agreenspan24 and put a code comment for readers linking to an explanation of dataclasses ie https://docs.python.org/3/library/dataclasses.html

shaunagm:
> Still pretty new to proper python conventions, so I appreciate the deeper look! A couple of ideas here sound interesting to me, like typing and dataclasses, but I don't want to introduce new patterns without approval from maintainers, so @shaunagm let us know your thoughts!

Responded re: dataclasses inline. The question of typing is a bigger one. We currently don't require any sort of type annotations and certainly not any strict enforcement, but perhaps that's something we could at least ask of contributors going forward. We certainly don't have capacity to retroactively add typing to the existing codebase but incremental improvements are better than nothing.

Thanks for raising these questions @agreenspan24 & @alxmrs

parsons/scytl/scytl.py
agreenspan24: Added as a dataclass with comment

agreenspan24:
@shaunagm any idea what's causing `docs-build` to fail? No clear answer from CircleCI

shaunagm:
I don't think it's an issue with your PR. @SorenSpicknall? This and a couple other open PRs are failing with this same error `Error: [Errno 2] No such file or directory: '/home/circleci/project/venv/bin/python3'`

SorenSpicknall:
> I don't think it's an issue with your PR. @SorenSpicknall? This and a couple other open PRs are failing with this same error `Error: [Errno 2] No such file or directory: '/home/circleci/project/venv/bin/python3'`

I'm working on this now; definitely looks like a CircleCI config breakage. If we can't figure it out in the next day or so, we may just merge some of these newer PRs without the additional testing step in order to not pause ongoing work, since the PRs generally test green aside from the CI issue.

parsons/scytl/scytl.py
agreenspan24: Sorry, that was a bit of a word salad. Does it make sense to you to split up the state-wide election and county-wide election into two separate connectors, since they have slightly different functions and outputs?

Separately, if we're using this for polling, to prevent fetching unnecessary results, does it make sense to you to store the previous version in the class or return it with the results and have them pass it back in the function?

shaunagm:
@SorenSpicknall I know you submitted a PR to fix the CI issue (https://github.com/move-coop/parsons/pull/752), what do we need to do to get this PR passing? Do we need to rebase it onto the updated main branch?

SorenSpicknall:
> @SorenSpicknall I know you submitted a PR to fix the CI issue (#752), what do we need to do to get this PR passing? Do we need to rebase it onto the updated main branch?

Yes, changes from `main` will need to be merged in or the PR will need to be recreated from scratch based on that newly updated `main` branch. Sorry about the annoyance of that, @agreenspan24 - you just got unlucky with timing of a weird cache issue with our CI.

shaunagm:
Let me know if you need help rebasing, @agreenspan24. It can be tricky.

shaunagm:
Hi @agreenspan24! Just checking in. I know a lot of folks are busy with the election, so I wanted to offer a few different options here:

(a) I can rebase your PR for you (I should be able to as a maintainer)
(b) we can pair and I can walk you through the process (don't know if you've rebased before, but it can be intimidating)
(c) we can just wait for you to have a chance to do it yourself

Totally up to you!

agreenspan24:
> Hi @agreenspan24! Just checking in. I know a lot of folks are busy with the election, so I wanted to offer a few different options here:
> 
> (a) I can rebase your PR for you (I should be able to as a maintainer) (b) we can pair and I can walk you through the process (don't know if you've rebased before, but it can be intimidating) (c) we can just wait for you to have a chance to do it yourself
> 
> Totally up to you!

@shaunagm so sorry I missed this request, I can find some time in the next few days to rebase

agreenspan24:
@shaunagm all done!

shaunagm (approved):

shaunagm:
Thank you so much for the effort you put into this PR! Excited to finally get to merge it. :)

PR #745, opened 2022-09-16
Add use_env_token option to S3 object

sjwmoveon:
Previously this was only settable in AWSConnection, but you can't pass an AWSConnection to the S3 object.


sjwmoveon:
@shaunagm here's the new pull request, following up on https://github.com/move-coop/parsons/pull/740

shaunagm (approved):
Looks good to me!

PR #753, opened 2022-09-26
Fixed GitHub connector API docs

### LINKED ISSUE #742: Github autodocs not displaying
The Github connector has plenty of public methods, but not are appearing in the docs. 

## Detailed Description

If you look at the Github connector there's [plenty of public methods](https://github.com/move-coop/parsons/blob/main/parsons/github/github.py#L112) but they don't appear in the docs, which just shows the class itself:

Screenshot from 2022-09-13 11-26-49

## Additional Context

The [wrapgithub404 decorator](https://github.com/move-coop/parsons/blob/main/parsons/github/github.py#L36) may be interfering with autodoc? That's my best guess. May be interacting with the specific directives given [here](https://github.com/move-coop/parsons/blob/main/docs/github.rst#api). I found [this StackOverflow question](https://stackoverflow.com/questions/3687046/python-sphinx-autodoc-and-decorated-members) but don't have the time to chase down further.

## Priority

Medium priority. This affects usability of the Github connector, but OTOH I think it's a sparsely used connector. But maybe it's sparsely used because the docs don't work.  ¯\_(ツ)_/¯

AndrewRook:
Even though `_wrap_method` is not _explicitly_ used as a decorator with the @ syntax, `decorate_methods` uses it in [functionally the same way](https://github.com/move-coop/parsons/blob/4775b5dfbcf5560a7dc2a617ed7847e41abdf450/parsons/github/github.py#L31). So adding `@wraps` to the inner function of `_wrap_method` seems to do the trick to propagate the docstrings.

I built the docs locally and it seems to work:
image

I also ran all the unit tests and nothing seems to have broken. 

Closes #742.

### DIFF
--- a/parsons/github/github.py
+++ b/parsons/github/github.py
@@ -13,6 +13,7 @@ logger = logging.getLogger(__name__)
 
 
 def _wrap_method(decorator, method):
+    @wraps(method)
     def _wrapper(self, *args, **kwargs):
         bound_method = partial(method.__get__(self, type(self)))
         return decorator(bound_method)(*args, **kwargs)


shaunagm:
@SorenSpicknall do you know if this is failing due to the CI issue from a few weeks ago? Do we need to rebase?

AndrewRook:
@shaunagm I just pulled in updates from `main` (or at least I think I did 😅) but it looks like the problem persists. Let me know if there's anything else I can try from my end, otherwise I'll hang tight. (Alternatively if you have a working setup I would not be offended if you just opened up a PR with the change yourself, the modification is tiny 😄.)

SorenSpicknall:
@AndrewRook, the CircleCI config can be a bit tricky at times with branch histories. Can you close this PR, recreate your changes on a new branch that's up to date with the main branch, and then open a PR from that new branch? I think that'll fix the issue, but if not I'll launch a deeper investigation.

AndrewRook:
Ok I just synced my fork, made a new branch off of `main`, remade the change, and PRed in #761. It doesn't look like CircleCI was able to find the config.yml in that one either, but regardless I'll close this PR now since it's superfluous.

PR #756, opened 2022-10-03
Update Code of Conduct contact email

shaunagm:
This looks like a bigger change than it is because I moved some linebreaks around to preserve matching line lengths. I just replaced "the project team at justin@movementcooperative.org" with "hr@movementcooperative.org".


elyse-weiss (approved):
LGTM

PR #759, opened 2022-10-18
Fix assert_matching_tables

crayolakat:
In `utils.py`, when comparing `assert list(r1) == list(r2)`, this only compares if the headers of `r1` and `r2` are equal, when it should be comparing if the entire contents of both tables are equal.

This is why the `test_google_admin.py` test was passing even though it had a typo and should have failed. This typo is also fixed in this PR


crayolakat:
After this change, a few CircleCI tests failed. This is because those tests either had typos or flaws which weren't being caught in the assert_matching_tables test since the function previously only checked if the headers of the dict tables matched.

I went ahead and fixed those tests.

shaunagm (approved):
We already talked about this in the Parsons Party, but (a) this looks good to me and (b) thank you so much for catching and fixing this!

PR #760, opened 2022-10-20
use requests library

agreenspan24:
@shaunagm 

When we merged to main, the outgoing requests started failing with a "Forbidden" response, even though it worked both locally and in the pull request.

Let's see if migrating to the requests library improves the situation.


agreenspan24:
Closing in favor of: https://github.com/move-coop/parsons/pull/762

PR #762, opened 2022-10-21
Scytl fix browser headers and requests lib w mock

agreenspan24:
@shaunagm 

Three issues were occurring that caused the tests to fail:
- First, the zip file request mocks weren't being set up correctly. This sets them as such.
- Second, the outgoing requests using urllib.request tooling was not respecting the mocks, instead going directly to the Scytl server. We didn't notice because the urls and files that were mocked and saved were identical to the ones on the server.
- Third, Scytl started enforcing logic to prevent requests that aren't coming from a browser, causing the 403 Forbidden errors. By adding a UserAgent header, Scytl will now think we are making a request from a browser.

Not sure why it didn't fail when we re-ran the tests last night in CircleCI, but this should be the fix.


SorenSpicknall (approved):
Thanks for the quick turnaround on this!

PR #769, opened 2022-10-22
Add get_aliases() function to Google Admin connector

crayolakat:
Also fix existing tests for them to make more sense

### DIFF
--- a/parsons/google/google_admin.py
+++ b/parsons/google/google_admin.py
@@ -63,6 +63,21 @@ class GoogleAdmin(object):
 
         return Table(ret)
 
+    def get_aliases(self, group_key, params=None):
+        """
+        Get aliases for a group. `Google Admin API Documentation <https://developers.google.com/\
+        admin-sdk/directory/reference/rest/v1/groups.aliases/list>`_
+
+        `Args:`
+            group_key: str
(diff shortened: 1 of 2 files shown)


shaunagm:
Overall this looks good to me - just had one question. Sorry to take so long to review!

crayolakat:
@shaunagm No problem, what is your question?

shaunagm:
I left a comment on line 50 of `test_google_admin.py`. In case it's not showing up for you, it's "Can you walk me through why we need to change the tests for these other methods? Are you just doing some cleanup, or did something actually change?"

crayolakat:
The `get_all_group_members` method is supposed to take a group key/ID as a parameter and return all the emails in that group. I noticed in the original `test_all_group_members` test that it was taking the string `group_key` as a parameter instead of the key/ID of one of the test groups defined in `mock_all_groups`. It makes more sense to pass in a group key, such as `1`, to test the method, so that's why I changed that test. Nothing in the code regarding the `get_all_group_members` function changed though, or any other Google Admin function

shaunagm (approved):

shaunagm:
Gotcha @crayolakat - thanks for the explanation! Merging now.

PR #770, opened 2022-10-26
NGPVAN: Saved List Overwrite Bug Fix

jburchard:
There was a bug in the code that would not allow you to overwrite a list. Also, cleaned up a copy/paste typo in the logging.

### DIFF
--- a/parsons/ngpvan/saved_lists.py
+++ b/parsons/ngpvan/saved_lists.py
@@ -122,7 +122,7 @@ class SavedLists(object):
         if folder_id not in [x['folderId'] for x in self.get_folders()]:
             raise ValueError("Folder does not exist or is not shared with API user.")
 
-        if list_name in [x['name'] for x in self.get_saved_lists(folder_id)]:
+        if list_name in [x['name'] for x in self.get_saved_lists(folder_id)] and not overwrite:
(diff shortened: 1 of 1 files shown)


parsons/ngpvan/saved_lists.py
```diff
@@ -102,7 +102,8 @@ def upload_saved_list_rest(self, tbl, url_type, folder_id, list_name,
                  Whether or not fields are enclosed in quotation marks within each
                  column of the file.
             overwrite: int
-                Replace saved list if already exists.
+                Replace saved list if already exists. Pass in the list id of the
```
SorenSpicknall, line 105: You've got a trailing space on this line that's creating a linting failure during testing - this looks good to me otherwise, so just patch that up and we will hopefully be able to get it moving.

SorenSpicknall (changes requested):
See comments on line 105 for one small change!

SorenSpicknall:
Hey @jburchard, do you happen to have any time coming up to fix these small linting issues? If not, I can do it myself - just didn't want to mess with the branch history if you were planning on coming back to it.

jburchard:
I can pick this up after Thanksgiving.

SorenSpicknall (approved):

Issue #718, opened 2022-07-21
[Feature/Addition] Create Donorbox Class


elyse-weiss:
A TMC member asked about syncing Donorbox data. I figured it would therefore be good to put a low priority issue in Parsons for class creation.

## Detailed Description
https://github.com/donorbox/donorbox-api

## Priority
Low priority

PR #773, opened 2022-11-15
Donorbox connector

shaunagm:
Adds a Donorbox connector implementing the endpoints listed [here](https://github.com/donorbox/donorbox-api). Addresses #718 

Contains tests and docs, although the live tests are currently skipped, as they require a paid account and we don't have one set up with our CI. I've tested them locally, though.


SorenSpicknall (approved):
Tentatively approving with two small suggestions that you can take or leave:

- Consider linking the API docs somewhere in the connector README (I know we don't do this universally right now, but it seems like a good habit to get in)
- Consider calling the test data file `donorbox_test_data` rather than `donorbox_fake_data`

PR #795, opened 2023-03-17
Use black formatting rather than flake8

### LINKED ISSUE #791: Switch to using Black instead of/in addition to PyLint
Manual linting is time-consuming for contributors. Switching to Black will allow contributors to lint with a single command.

Implementing this is going to involve a massive PR, since Black is very opinionated and will touch a ton of files. But it seems worth it in the long run.

austinweisgrau:
Resolves #791


austinweisgrau:
Probably should use black in addition to flake8, rather than replacing flake8. Closing this PR and will open another one.

PR #796, opened 2023-03-17
Use black formatting in addition to flake8

austinweisgrau:
Resolves [#791](https://github.com/move-coop/parsons/issues/791)


austinweisgrau:
Having an issue restoring the flake8 `noqa` functionality on F401 in two modules. Seems to be a bug in flake8.

A few options here:
1. run flake8 with this option `--per-file-ignores='test/test_sftp.py:F401 test/test_sftp_ssh.py:F401'` 
2. add F401 to `--extend-ignore` option
3. Remove the unused imports from these two modules

SorenSpicknall (approved):

shaunagm (approved):
haven't actually looked through everything, but Soren has so this lets us merge

PR #797, opened 2023-03-17
Redshift.copy() can use AWS Session Token as part of auth variables

austinweisgrau:
A session token is sometimes a necessary component of AWS auth
credentials alongside an access key and secret key. This commit
enables a session token to be passed in the same ways that the other
credentials are passed:

- directly as part of copy_args
- through environmental variables

A test is added at `test/test_redshift.py::TestRedshift::test_get_creds_kwargs_with_token` to validate this behavior.

### DIFF
--- a/parsons/databases/redshift/redshift.py
+++ b/parsons/databases/redshift/redshift.py
@@ -488,6 +488,8 @@ class Redshift(
         alter_table_cascade=False,
         aws_access_key_id=None,
         aws_secret_access_key=None,
+        aws_session_token=None,
+        use_env_token=False,
         iam_role=None,
         cleanup_s3_file=True,
         template_table=None,
@@ -586,6 +588,9 @@ class Redshift(
             aws_secret_access_key:
                 An AWS secret access key granted to the bucket where the file is located. Not
                 required if keys are stored as environmental variables.
+            aws_session_token:
+                An AWS session token granted to the bucket where the file is located. Not
+                required if keys are stored as environmental variables.
             iam_role: str
                 An AWS IAM Role ARN string; an alternative credential for the COPY command
                 from Redshift to S3. The IAM role must have been assigned to the Redshift
@@ -653,6 +658,8 @@ class Redshift(
                 tbl,
                 aws_access_key_id=aws_access_key_id,
                 aws_secret_access_key=aws_secret_access_key,
+                aws_session_token=aws_session_token,
+                use_env_token=use_env_token,
                 csv_encoding=csv_encoding,
             )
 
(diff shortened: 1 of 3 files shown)


austinweisgrau:
Resolves #798

parsons/databases/redshift/redshift.py
```diff
@@ -488,6 +488,8 @@ def copy(
         alter_table_cascade=False,
         aws_access_key_id=None,
         aws_secret_access_key=None,
+        aws_session_token=None,
+        use_env_token=False,
```
Jason94, line 492: I looked through the code, and I don't _think_ that this is being used. It gets passed around a bit, but ultimately `temp_s3_copy` is passing the field from the Redshift instance, not the parameters added in this PR, into the `S3` object. See rs_copy_table.py:L190.

And I agree with the way that it's technically implemented; the better design here is to defer to the `use_env_token` from the Redshift constructor. I think you can just remove the `use_env_token` parameter from the `copy` method, and the other methods where it's just being passed around.

parsons/databases/redshift/redshift.py
```diff
@@ -586,6 +588,9 @@ def copy(
             aws_secret_access_key:
                 An AWS secret access key granted to the bucket where the file is located. Not
                 required if keys are stored as environmental variables.
+            aws_session_token:
+                An AWS session token granted to the bucket where the file is located. Not
+                required if keys are stored as environmental variables.
```
Jason94, line 593: Can you give the exact environmental variable the session token needs to be stored in as part of the documentation?

parsons/databases/redshift/rs_copy_table.py
```diff
@@ -125,30 +132,41 @@ def get_creds(self, aws_access_key_id, aws_secret_access_key):
...
+
+        return result
 
     def temp_s3_copy(
         self,
         tbl,
         aws_access_key_id=None,
         aws_secret_access_key=None,
+        aws_session_token=None,
+        use_env_token=False,
```
Jason94, line 169: Same comment as above - is this `use_env_token` being used?

Jason94 (commented):
Thanks for submitting the PR! I think this would be great to have.

sharinetmc:
Hi @austinweisgrau, just wondering if you were able to get to follow-up on some of the changes Justin suggested?

sharinetmc (changes requested):
Hi @austinweisgrau! Just changing the status to 'request changes'. Please let me know if you have any questions!

austinweisgrau:
Closing this

Issue #784, opened 2023-01-09
[Bug] S3 utilities in Parsons do not accommodate different subfolder permissions


elyse-weiss:
Many of the S3 utilities in Parsons aren't able to accommodate setups like TMC's where subfolders of S3 buckets have different permissions from each other - we need to make some adjustments to avoid future breakages, likely by detecting when subfolders are present in a bucket and adding them to a "prefix" portion of an AWS API call rather than calling directly into the AWS API with the bucket + subfolder path.

## Your Environment

* Version of Parsons used (if you know):
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):

## Priority
Please indicate whether fixing this bug is high, medium, or low priority for you. If the issue is time-sensitive for you, please let us know when you need it addressed by.

elyse-weiss:
Just pinging on this, as it came up for me once again. ControlShift gives clients access to sub-folders, so you can't use `s3.list_keys()` to see what's in your sub-folder.

PR #803, opened 2023-04-10
Standardize S3 Prefix Conventions

IanRFerguson:
# Included Here

This PR addresses Issue #784 (albeit incompletely). We included logic to confirm the `prefix/` conventions in `s3.list_keys()` and alert the user via logs and a raised Exception that their API call was rejected due to missing prefix input.

# Future Directions

This is a band-aid and doesn't address the root issue. `boto3` is not built to handle API calls to buckets where permission allowances are incomplete (i.e., we have access to a few prefixes but NOT the whole bucket). The API is somewhat opaque in this way, and attempting to do this gets inefficient very fast. We should revisit but this is sufficient for now.

### DIFF
--- a/parsons/aws/s3.py
+++ b/parsons/aws/s3.py
@@ -1,5 +1,6 @@
 import re
 import boto3
+from botocore.client import ClientError
 from parsons.utilities import files
 import logging
 import os
@@ -15,7 +16,6 @@ class AWSConnection(object):
         aws_session_token=None,
         use_env_token=True,
     ):
-
         # Order of operations for searching for keys:
         #   1. Look for keys passed as kwargs
         #   2. Look for env variables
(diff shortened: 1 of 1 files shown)


parsons/aws/s3.py
```diff
@@ -159,18 +158,33 @@ def list_keys(
...
                 args["ContinuationToken"] = continuation_token
+
             args.update(kwargs)
 
-            resp = self.client.list_objects_v2(**args)
+            try:
+                resp = self.client.list_objects_v2(**args)
 
-            for key in resp.get("Contents", []):
+            except ClientError as e:
```
willyraedy, line 179: Generally prefer to catch the smallest error. Is this as granular as it gets or is there a more specific auth error?

willyraedy (approved):
Overall looks good! One small question around error granualarity

parsons/aws/s3.py
IanRFerguson: I haven't been able to throw a more granular exception here... seems as soon as the `resp = self.client.list_objects_v2(**args)` request processes and *any* keys are out of the scope of our access it'll raise this error

My initial plan was to isolate the auth component and build a list of files that we have access to, but in a bucket where we have partial access I think we're going to keep bumping up against the same issue. Does that make sense?

PR #807, opened 2023-04-13
Add Default Parameter Flexibility

IanRFerguson:
My last update did not account for the fact that `prefix = None` is default ... rookie mistake!

This hotfix first confirms that a prefix was supplied, and THEN it checks to see if a trailing `/` character is included

### DIFF
--- a/parsons/aws/s3.py
+++ b/parsons/aws/s3.py
@@ -158,8 +158,8 @@ class S3(object):
 
         continuation_token = None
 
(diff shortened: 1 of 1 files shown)


IanRFerguson:
Before this PR...

image

IanRFerguson:
With new logic included...

image

KasiaHinkson (approved):
LGTM, my kind of error!

PR #809, opened 2023-04-17
Catch File Extensions in S3 Prefix

IanRFerguson:
Adding in logic to skip the `prefix -> prefix/` logic here if a file extension is passed in.

Will defer to @KasiaHinkson on what the best solution is here (not sure I totally understand where this is breaking in STW sync)

### DIFF
--- a/parsons/aws/s3.py
+++ b/parsons/aws/s3.py
@@ -158,9 +158,21 @@ class S3(object):
 
         continuation_token = None
 
-        # Confirm that prefix has trailing / if it's provided
-        if prefix and not prefix.endswith("/"):
-            prefix = f"{prefix}/"
+        """
+        Two conditions here...
+
+        * User supplied a prefix
+        * The prefix is a route OR file type
+        * We assume a '.' here specifies a file extension
+
+        If all of these conditions are met then a trailing slash is implicitly
+        added to the prefix
+        """
+
+        if prefix and not prefix.endswith("/") and "." not in prefix:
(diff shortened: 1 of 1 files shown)


IanRFerguson:
@willyraedy what do you think about this logic? instead of explicitly adding a slash this PR will try the API call first and then if that raises an exception it will add a trailing slash

parsons/aws/s3.py
```diff
@@ -177,10 +173,12 @@ def list_keys(
...
 
             except ClientError as e:
-                logger.error("FAILED TO RETURN OBJECTS!")
-                logger.error(
-                    "Check your permissions in this bucket, and consider providing a prefix"
-                )
+                error_message = """Unable to list bucket objects!
+                This may be due to a lack of permission on the requested
+                bucket. Double-check that you have sufficient READ permissions
+                on the bucket you've requested"""
```
willyraedy, line 179:
```suggestion
                error_message = """Unable to list bucket objects!
                This may be due to a lack of permission on the requested
                bucket. Double-check that you have sufficient READ permissions
                on the bucket you've requested. If you only have permissions for
                keys within a specific prefix, make sure you include a trailing '/' in
                in prefix."""
```

willyraedy (approved):
Looks good just have to figure out the circle ci thing

parsons/aws/s3.py
willyraedy: I think I have a typo but you get the idea

PR #815, opened 2023-05-09
DatabaseConnector Interface to Major Release

Jason94:
[Here is a spreadsheet with the analysis I did of current feature parity between the four database connectors](https://docs.google.com/spreadsheets/d/1_jbhYk7lRCGwbwA00WKjJK9Ne0Zku_g_C3SuPx0m5-k/edit#gid=0).

Currently there are three methods they share in common that I think are worth representing in the interface:

- table_exists
- query
- copy

Of those, the main difference between the functionality is copy , where Redshift has a ton of extra functionality that the others do not have.

This PR creates the interface ABC, DatabaseConnector, and add the nifty discover_database function. I also do a little cleanup on the existing connectors, mostly adding type hints. These are very useful to make sure that the connector classes align properly with the interface and each other.

The only breaking change is switching the default value of strict_length in the Postgres connector from False to True, to align with the Redshift & MySQL connectors.

This PR only attempts to create the interface "as is", documenting the existing behavior shared by the database connectors (apart from the one Postgres default value). Once DatabaseConnector is merged in, we can plan out how we want to address the feature parity.

### DIFF
--- a/parsons/databases/database_connector.py
+++ b/parsons/databases/database_connector.py
@@ -65,7 +65,7 @@ class DatabaseConnector(ABC):
             if_exists (str):
                 If the table already exists, either ``fail``, ``append``, ``drop``
                 or ``truncate`` the table.
-            strict_length (bool, optional): 
+            strict_length (bool, optional):
                 Whether or not to tightly fit the length of the table columns to the length
                 of the data in ``tbl``; if ``padding`` is specified, this argument is ignored.
                 Defaults to True.
--- a/parsons/databases/database_connector.py
+++ b/parsons/databases/database_connector.py
@@ -9,8 +9,63 @@ class DatabaseConnector(ABC):
     This class should be used in functions instead of the specific database connector classes
     when the functions don't rely on database-specific functionality.
 
-    It ensures that any class that inherits from it implements the `table_exists`, `copy`, and
-    `query` methods, which are common operations when working with databases.
+    It ensures that any class that inherits from it implements the methods that are uniform
+    operations when working with databases.
+
+    Should you use `DatabaseConnector` instead of `Redshift`/`BigQuery`/etc?
+
+    Overall this class is mostly useful for code in the Parsons library, not code using it.
+    There could be some exceptions. In general though, if you are writing a script to do a task
+    like moving data out of an API service and into a data warehouse, you probably do not need
+    to use DatabaseConnector. You can probably just use the Parsons class that directly corresponds
+    with the database that you use.
+
+    Here are more examples of situations where you may or may not need to use DatabaseConnector:
+
+        1. You do not use type annotations, or you don't know what "type annotations" are - No
+
+            If you do not use type annotations for your code, then you do not need to think about
+            `DatabaseConnector` when writing your code. This is the most common case. If none
+            of the cases below apply to you, then you probably don't need it.
+
+            In this simple example, we are not using type annotations in our code. We don't need
+            to think about exactly what class is being passed in. Python will figure it out.
+
+            ```python
+            def my_database_function(db):
+                    some_data = get_some_data()
+                    db.copy("some_table", some_data)
+
+            # These will all just work:
+            my_database_function(Redshift())
+            my_database_function(MySQL())
+            my_database_functon(BigQuery())
+            ```
+
+        2. You only use one database in your work - No
+
+            This is where most people will fall. Usually code is not intended to run on
+            multiple databases without modification. For example, if you are working for
+            an organization that uses Amazon Redshift as your data warehouse, you do not
+            need to use `DatabaseConnector` to write ETL scripts to load data into your
+            Redshift. It is rare that organizations switch databases. In the cases where
+            that does occur, usually more work is required to migrate your environment and
+            your vendor-specific SQL than would be saved by using `DatabaseConnector`.
+
+        3. You are writing a sample script or a tutorial - Yes
+
+            If you are using Parsons to write a sample script or tutorial, you should use
+            `DatabaseConnector`! If you use `DatabaseConnector` type annotations and the
+            `discover_database` function, then your sample code will run on any system.
+            This makes it much easier for new programmers to get your code working on
+            their system.
+
+        4. Utility code inside Parsons or other libraries - Yes
+
+            If you are writing a utility script inside Parsons or another library meant
+            for broad distribution, you should probably use `DatabaseConnector` type
+            annotations. This will ensure that your library code will be usable by the
+            widest possible set of users, not just users on one specific database.
 
     Note:
         The Python type system (as of 3.10.6) will not stop you from breaking the type contract
--- a/parsons/databases/database_connector.py
+++ b/parsons/databases/database_connector.py
@@ -1,0 +1,112 @@
+from abc import ABC, abstractmethod
+from typing import Optional
+from parsons.etl.table import Table
+
+
+class DatabaseConnector(ABC):
+    """
+    An abstract base class that provides a uniform interface for all Parsons database connectors.
+    This class should be used in functions instead of the specific database connector classes
+    when the functions don't rely on database-specific functionality.
+
+    It ensures that any class that inherits from it implements the `table_exists`, `copy`, and
+    `query` methods, which are common operations when working with databases.
+
+    Note:
+        The Python type system (as of 3.10.6) will not stop you from breaking the type contract
+        of method signatures when implementing a subclass. It is up to the author of a database
+        connector to ensure that it satisfies this interface. Be careful to, for example, not
+        change the types of the parameters or leave out optional parameters that are specified
+        in the interface.
+
+        Any such inconsistencies can cause unexpected runtime errors that will not be caught by
+        the type checker.
+
+        It is possible to safely add additional features, such as new methods or extra **optional**
+        parameters to specified methods. In general adding new methods is safe, but adding optional
+        parameters to methods specified in the interface should be considered bad practice, because
+        it could result in unexpected behavior.
+
+    Example usage:
+
+    .. code-block:: python
+
+        def my_function(db: DatabaseConnector, data: Table):
+            # Your code here, using the db object
+
+        # Pass an instance of a class that inherits from DatabaseConnector, e.g. Redshift
+        my_function(some_db_instance, some_data)
+
+    """
+
+    @abstractmethod
+    def table_exists(self, table_name: str) -> bool:
+        """ Check if a table or view exists in the database.
+
+        `Args:`
+            table_name: str
+                The table name and schema (e.g. ``myschema.mytable``).
+
+        `Returns:`
+            boolean
+                ``True`` if the table exists and ``False`` if it does not.
+        """
+        pass
+
+    @abstractmethod
+    def copy(self, tbl: Table, table_name: str, if_exists: str, strict_length=True):
+        """ Copy a :ref:`parsons-table` to the database.
+
+        `Args`:
+            tbl (Table):
+                Table containing the data to save.
+            table_name (str):
+                The destination table name (ex. ``my_schema.my_table``).
+            if_exists (str):
+                If the table already exists, either ``fail``, ``append``, ``drop``
+                or ``truncate`` the table.
+            strict_length (bool, optional): 
+                Whether or not to tightly fit the length of the table columns to the length
+                of the data in ``tbl``; if ``padding`` is specified, this argument is ignored.
+                Defaults to True.
+        """
+        pass
+
+    @abstractmethod
+    def query(self, sql: str, parameters: Optional[list] = None) -> Optional[Table]:
+        """ Execute a query against the database. Will return ``None`` if the query returns empty.
+
+        To include python variables in your query, it is recommended to pass them as parameters,
+        following the `psycopg style
+          <http://initd.org/psycopg/docs/usage.html#passing-parameters-to-sql-queries>`.
+        Using the ``parameters`` argument ensures that values are escaped properly, and avoids SQL
+        injection attacks.
+
+        **Parameter Examples**
+
+        .. code-block:: python
+
+            # Note that the name contains a quote, which could break your query if not escaped
+            # properly.
+            name = "Beatrice O'Brady"
+            sql = "SELECT * FROM my_table WHERE name = %s"
+            db.query(sql, parameters=[name])
+
+        .. code-block:: python
+
+            names = ["Allen Smith", "Beatrice O'Brady", "Cathy Thompson"]
+            placeholders = ', '.join('%s' for item in names)
+            sql = f"SELECT * FROM my_table WHERE name IN ({placeholders})"
+            db.query(sql, parameters=names)
+
+        `Args:`
+            sql: str
+                A valid SQL statement
+            parameters: Optional[list]
+                A list of python variables to be converted into SQL values in your query
+
+        `Returns:`
+            Parsons Table
+                See :ref:`parsons-table` for output options.
+        """
+        pass
--- a/parsons/databases/database_connector.py
+++ b/parsons/databases/database_connector.py
@@ -54,7 +54,7 @@ class DatabaseConnector(ABC):
         pass
 
     @abstractmethod
-    def copy(self, tbl: Table, table_name: str, if_exists: str, strict_length=True):
+    def copy(self, tbl: Table, table_name: str, if_exists: str):
         """ Copy a :ref:`parsons-table` to the database.
 
         `Args`:
@@ -65,9 +65,6 @@ class DatabaseConnector(ABC):
             if_exists (str):
                 If the table already exists, either ``fail``, ``append``, ``drop``
                 or ``truncate`` the table.
-            strict_length (bool, optional):
-                Whether or not to tightly fit the length of the table columns to the length
-                of the data in ``tbl``. Defaults to True.
         """
         pass
 
--- a/parsons/databases/database_connector.py
+++ b/parsons/databases/database_connector.py
@@ -80,9 +80,10 @@ class DatabaseConnector(ABC):
 
         If you go the second route, you can add a stub method like this:
 
-            ```python
-            def new_method(self, arg1, arg2):
-                raise NotImplementedError("Method not implemented for this database connector.")
+            .. code-block:: python
+
+                def new_method(self, arg1, arg2):
+                    raise NotImplementedError("Method not implemented for this database connector.")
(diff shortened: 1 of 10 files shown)


elyse-weiss:
@willyraedy can catch you up during our check in, but this is a pretty big architectural change so would love your eyes on this (as well as Shauna's review). Jason is also going to add some additional documentation that we talked about on our call.

shaunagm:
Hi Jason, here are some documentation notes from our call:

- The impact to users should be relatively minimal, so we don't need too many user-facing doc updates. Possibly a style recommendation to ie `db = Redshift()` instead of `rs = Redshift()` in their code if they think there's any chance, however slim, of them switching database connectors some day.
- For developer docs, we should add a brief explanation of how ABCs work and how our intention is to keep the interfaces between the different DB connectors as similar as possible. They should be told about the three options (add a method to the ABC and implement for all databases; add a method to ABC and implement for some databases while adding stubs for others; and simply implementing the method on a database connector itself instead of the ABC.
- Developers should be told how to add stub methods if they go that route. We should have a template or example we point devs to that clearly conveys to the user that _this method does not exist yet_
- If they choose the "just add to an existing connector" route, they need to manually check the other connectors and conform to their interface. (So if you are adding method X that is already on Redshift to BigQuery but not the other databases, make sure BigQuery's new method X is the same interface as Redshift's method X, or covariant with it (I think I'm using covariant correctly?))

parsons/databases/database_connector.py
```diff
@@ -0,0 +1,108 @@
+from abc import ABC, abstractmethod
+from typing import Optional
+from parsons.etl.table import Table
+
+
+class DatabaseConnector(ABC):
+    """
+    An abstract base class that provides a uniform interface for all Parsons database connectors.
+    This class should be used in functions instead of the specific database connector classes
```
shaunagm, line 9: I think readers are going to need an example of when they should use DatabaseConnector vs a specific connector class. I need an example and I understand ABCs pretty well!

Also worth thinking about whether this documentation is primarily meant for Parsons contributors or Parsons users. I assume Parsons contributors, but I could see a user reading this and getting confused.

parsons/databases/discover_database.py
```diff
@@ -0,0 +1,35 @@
...
+    If no suitable configuration is found, will raise an error.
+
+    Note that the variables to be searched for are hard-coded in this function,
+    since they are unlikely to change. If that is done, for some reason, or a
+    new database connector is added, ``discover_database`` should be updated
+
+    Returns:
+        DatabaseConnector: The database connector configured in the environment.
+    """
+    if os.getenv("REDSHIFT_PASSWORD"):
```
shaunagm, line 26: I am slightly concerned about the logic here, where if someone happens to have two different db environmental variables set and wants to use the DB connector that is later in this list, they'd get an unexpected result. But I imagine "someone happens to have two different db environmental variables set" is probably pretty unlikely in practice.

shaunagm (approved):

shaunagm (changes requested):
Whoops, meant to submit review as "request changes" since we need additional docs at least.

parsons/databases/database_connector.py
```diff
@@ -0,0 +1,108 @@
...
+
+    .. code-block:: python
+
+        def my_function(db: DatabaseConnector, data: Table):
+            # Your code here, using the db object
+
+        # Pass an instance of a class that inherits from DatabaseConnector, e.g. Redshift
+        my_function(some_db_instance, some_data)
+
+    """
```
willyraedy, line 40: I think we should also specify the arguments for the init method. I glanced at a few of the database connector classes and seems the all have username, password, db, host, port. Specifying that in the ABC will also make the consistency easier for the instantiation of the generic db connector, particularly like in the toy example I put in another comment.

I think the goal of being able to point the same pipeline or script at a different warehouse without a deployment (ie without changing the code) and just adjusting env variables is easier if we keep this consistent as well

parsons/databases/discover_database.py
willyraedy: Love the idea of this function! Would be great to parameterize what database the connection points to.

However, I think Shauna has a point and don't think we should do this globally. For example, a pipeline from a Postgres application database to a Redshift data warehouse is not all that uncommon in which case, the same env would need both Postgres and Redshift credentials.

I think in that type of situation, you'd want something like
```
app_db = fetch_application_database_connector()
data_warehouse = fetch_datawarehouse_connector()
```

Perhaps we just do a relatively simple `fetch_database_connector` function that accepts a database type argument (preferably an enum) and then forwards all the arguments. So my toy example above for a postgres-redshift pipeline would look something like:

```
app_db = fetch_database_connector(database_type=os.getenv('APPLICATION_DATABASE_TYPE'))
data_warehouse = fetch_database_connector(database_type=os.getenv('DATA_WAREHOUSE_TYPE'))
```

And by passing through the arguments, we could support multiple database connections of the same type. For instance, perhaps there's one _main_ Postgres database but someone has a second software application. They'd want to define the default Postgres credentials in their env but then have the ability to connect to the second db when necessary.

Then that would look like (pretending the second software application was called widget):
```
app_db = fetch_database_connector(
        database_type=os.getenv('WIDGET_DATABASE_TYPE'),
        username = os.environ.get("WIDGET_USERNAME"),
        password = os.environ.get("WIDGET_PASSWORD"),
        ...
)
data_warehousse = fetch_database_connector(database_type=os.getenv('DATA_WAREHOUSE_TYPE'))
```

parsons/databases/postgres/postgres.py
```diff
@@ -52,7 +54,13 @@ def __init__(
         self.timeout = timeout
         self.dialect = "postgres"
 
-    def copy(self, tbl, table_name, if_exists="fail", strict_length=False):
+    def copy(
+        self,
+        tbl: Table,
+        table_name: str,
+        if_exists: str = "fail",
+        strict_length: bool = True,
```
willyraedy, line 62: This could theoretically be a breaking change for some pipelines. To @elyse-weiss comment about improving padding defaults, what if we added `padding` to the ABC? I think a sensible default padding makes sense to me and would mean this would be less likely to break things

parsons/databases/database_connector.py
```diff
@@ -0,0 +1,108 @@
...
+
+    It ensures that any class that inherits from it implements the `table_exists`, `copy`, and
+    `query` methods, which are common operations when working with databases.
+
+    Note:
+        The Python type system (as of 3.10.6) will not stop you from breaking the type contract
+        of method signatures when implementing a subclass. It is up to the author of a database
+        connector to ensure that it satisfies this interface. Be careful to, for example, not
+        change the types of the parameters or leave out optional parameters that are specified
+        in the interface.
```
willyraedy, line 20: I'm gonna take a gander and read some more about options here. Would love to see if there's a package or some other way to enforce things. I would also love to learn more about the extra optional parameters. That seems like something we'll want to support and then have descriptive errors for.

willyraedy (commented):
Super exciting and very much appreciate the care and thought here!

I want to poke around and see what options might exist for enforcing the ABC and then I think we should think a little bit more about the `discover_database` functionality and how we want to parameterize database connections. But this is a great start!

sharinetmc:
Hi @Jason94! Wondering if you had any updates on this (definitely no rush though).

parsons/databases/database_connector.py
Jason94: I have added some (possibly too) extensive documentation that hopefully makes this clearer!

parsons/databases/database_connector.py
```diff
@@ -0,0 +1,108 @@
...
+
+        Any such inconsistencies can cause unexpected runtime errors that will not be caught by
+        the type checker.
+
+        It is possible to safely add additional features, such as new methods or extra **optional**
+        parameters to specified methods. In general adding new methods is safe, but adding optional
+        parameters to methods specified in the interface should be considered bad practice, because
```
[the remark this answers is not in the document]
Jason94: Done!

parsons/databases/discover_database.py
Jason94: Those are valid concerns. I wanted to keep it simple from a user-standpoint, so I kept all of the environmental variable checking inside the function. However I did make a few changes:

- If multiple databases are detected, it raises an error _unless_
- You can specify a default database by simply passing in the class: `discover_database(default_connector=MySQL)`
- If multiple databases are detected but the default is not among them, it errors
- You can pass in an ordered list of preference like this, and it will default to the first one given in the list if it finds multiple, otherwise return the one that is found: `discover_database(default_connector=[MySQL, Redshift])`

The logic is a bit intricate so I added a test file `tests\test_databases\test_discover_database.py` with nine test cases.

parsons/databases/database_connector.py
Jason94: I don't think this is a good idea. Constructors in class hierarchies are special cases. That would limit the possible definition of what could be a database connector to things that only have a username, password, db, host, and port. For example, that would rule out creating a SQLite connector, because SQLite does not have any of those things.

I'm not even sure that Python's ABC's support this.

parsons/databases/discover_database.py
Jason94: I also think that for some of these more specialized cases, like where you want to control exactly what username gets passed into what, the recommendation should be to simply instantiate the database directly. I'm not sure if something like this:

```
app_db = fetch_database_connector(
        database_type=os.getenv('WIDGET_DATABASE_TYPE'),
        username = os.environ.get("WIDGET_USERNAME"),
        password = os.environ.get("WIDGET_PASSWORD"),
        ...
)
data_warehousse = fetch_database_connector(database_type=os.getenv('DATA_WAREHOUSE_TYPE'))
```

is better than:

```
app_db = Postgres(...)
warehouse_db = Redshift(...)
```

To expound, I don't think that discover_database should be a recommendation for all-purpose usage. If someone is an engineer at a TMC member org and they have a PostgreSQL data warehouse, they should probably just use 

```
db: DatabaseConnector = Postgres()
# or
db: Postgres = Postgres()
```

instead of `discover_database`. I think the main benefit of having `discover_database` is:

1. Sample scripts and tutorials
2. Utility functions inside the Parsons library like this one https://github.com/move-coop/parsons/pull/841
3. Other libraries that might want to use Parsons to have portable data-warehouse routines.

parsons/databases/database_connector.py
Jason94: Please let me know if you found anything. As far as I could tell, the ecosystem of Python type-checking tools isn't to this level of sophistication at static-time analysis.

The ABC module does raise a runtime error if you instantiate classes that inherit from an ABC and haven't defined all of the methods. But from what I was able to determine, it doesn't check the type signature of those methods. I'm not surprised, because the Python runtime isn't aware of the static time type annotations at all.

I agree this would be great to have.

parsons/databases/postgres/postgres.py
Jason94: Yes, that's true. I think at some point we're probably going to want to make a breaking change. For the meantime, I've reverted this change.

Regarding padding, I think that's a good idea. We should probably add that to the issue tracking the features to add to database parity, and we can work on that once the ABC is merged into `major-release`.

parsons/databases/database_connector.py
shaunagm: This documentation is great! Super helpful. And yes it's a lot of words but I'd rather err on the side of giving people too much information than too little.

parsons/databases/database_connector.py
shaunagm: Yeah we've talked a bit about trying to add more type hints/type checking to Parsons. I'd love to see that too but it seems like a separate issue that should not block this PR.

parsons/databases/database_connector.py
shaunagm: Python's ABCs do support this (see [this StackOverflow question](https://stackoverflow.com/questions/72643177/how-to-create-a-class-with-an-abstract-init-method)) but I agree with Jason that the benefits of standardizing the `__init__` method are outweighed by the fact doing so would exclude potential edge cases like SQLite. 

However - and this may be what Willy was suggesting in the first place - we _can_ specify a default `__init__` implementation that subclasses can inherit or choose to override, without making it an abstract method that they would have to signature-match (see [this other StackOverflow question](https://stackoverflow.com/a/71159174)).

parsons/databases/discover_database.py
shaunagm: Thanks for the clarification, Jason. I was initially imagining this might get used in user scripts but I see your point now about it mostly being helpful for situations where you don't know what database connector might be found, like a sample script or an internal utility function. If you know what databases you have you can just instantiate them directly. (Although I don't think Parsons currently lets you instantiate two connectors of the same type to two different databases, because the environmental variables would override each other, which is it's own separate issue.)

@willyraedy does this resolve your question or are we misunderstanding the use cases you're talking about?

shaunagm (commented):
Overall this looks good to me. I think it might be nice to implement a non-abstract default `__init__` method for DatabaseConnector which will enhance (but not require) uniformity, but that's the only change I'd make at this point, and I defer to you on whether you think that's worth doing.

I want to make sure Willy gets a chance to re-review so not setting this to Approved yet.

parsons/databases/database_connector.py
Jason94: Ah, that's good to know. Thanks for finding that, Shauna. I'm still not sure what the purpose would be, though. If the goal is just to have a documented preferred signature for a typical connector (while not ruling out an atypical case) and to save some typing, that seems fine to me.

I think that @willyraedy's vision for the `discover_database` function might be informing his comment. In his example he posted in that thread, the discover function supported passing arguments through to the different constructors. If that's the starting point, then you would _need_ the constructor signatures to have consistent parameter names and types.

parsons/databases/discover_database.py
Jason94: I've thought about this more, and I think that what @willyraedy is suggesting is a different use case. The idea behind my `discover_database` function is that it's a super easy to use, no-thought function you can just drop into some code and it should do its best to just make the database connection part of the script work.

Willy, if I'm understanding you correctly, then you want to have a function that will automatically load the database connection from specific configuration for that purpose in the environment. This is kind of the opposite purpose of `discover_database`. It's a very advanced use case, where you want the ability to deploy the same piece of software in different environments and to be able to change the database that it's running against without any code changes. I think that's another really interesting utility function around `DatabaseConnector` and I can see why TMC would be interested in that.

I think it would be best to implement that as a separate function, because it has a different job.  Perhaps you could pass in a prefix for a set of environmental variables. The function gets the database type from PREFIX_DATABASE_TYPE. Then it gets all of the other env vars that start with that prefix and passes them (with the prefix stripped out) as keyword arguments to the connector class that it finds.

```python
app_db: DatabaseConnector = fetch_environment_connection(prefix="APP")
    # env: APP_DATABASE_TYPE = "Redshift"
    # env: APP_REDSHIFT_USERNAME = "..."
    # etc.

warehouse_db: DatabaseConnector = fetch_environment_connection(prefix="WAREHOUSE")
    # env: WAREHOUSE_DATABASE_TYPE = "GoogleBigQuery"
    # env: WAREHOUSE_GOOGLE_CREDENTIALS = "..."
    # etc.
```

This would allow you to hot-swap multiple databases being used in the code by just changing the environmental variables, which is definitely something `discover_database` does _not_ support.

Jason94: Something like this would work (although it would require that DatabaseConnectors only take string values, or at least can handle strings as any of their arguments):

```python
def fetch_environment_connection(prefix: str) -> DatabaseConnector:
    database_type = os.getenv(f'{prefix}_DATABASE_TYPE')

    if database_type is None:
        raise EnvironmentError(f"Missing environment variable: {prefix}_DATABASE_TYPE")

    try:
        database_class = globals()[database_type]
    except KeyError:
        raise EnvironmentError(f"Unsupported database type: {database_type}")

    if not issubclass(database_class, DatabaseConnector):
        raise EnvironmentError(f"{database_type} is not a subclass of DatabaseConnector")

    kwargs = {}
    for key, value in os.environ.items():
        if key.startswith(f'{prefix}_') and key != f'{prefix}_DATABASE_TYPE':
            key_without_prefix = key[len(f'{prefix}_'):]
            # Lower is required because env. vars. are always (often?) uppercased. If anyone
            # broke Python convention and used a capital letter in their constructor parameters
            # then this won't work. But don't do that >_>
            kwargs[key_without_prefix.lower()] = value

    instance = database_class(**kwargs)

    return instance
```

parsons/databases/database_connector.py
shaunagm: I have a modest preference for including the default/overridable implementation in `__init__`, totally okay with being overridden though.

shaunagm: @willyraedy can you weigh in here and on the other comment threads in this PR? I feel like we're getting close to the finish line.

parsons/databases/postgres/postgres.py
Jason94: @shaunagm We're tracking these changes to do in an issue, right? Could we add padding there?

parsons/databases/discover_database.py
```diff
@@ -0,0 +1,79 @@
...
+from typing import Optional, Union, Type, List
+
+from parsons.databases.database_connector import DatabaseConnector
+from parsons.databases.redshift import Redshift
+from parsons.databases.mysql import MySQL
+from parsons.databases.postgres import Postgres
+from parsons.google.google_bigquery import GoogleBigQuery
+
+
+def discover_database(
```
willyraedy, line 11: I think this new implementation avoids the unexpected behaviors so I don't think this needs to be a blocker here.

parsons/databases/database_connector.py
willyraedy: Yeah I think this is about how implicit / explicit the multi-database support is.

My instinct is to keep it simple and as-is until we start working some of those use cases in a more concrete way.

willyraedy (approved):
Sorry for the delay here. I think we can go ahead and merge this.

Longer term, TMC has a PR on expanding out the functionality of the BigQuery class that should be up this week. 

My suggestion would be to wait for a major release to include that.

parsons/databases/postgres/postgres.py
shaunagm: Added it here: https://github.com/move-coop/parsons/issues/838#issuecomment-1682849628

shaunagm (approved):
Looks good to me! Thanks for all your work on this, Jason.

PR #819, opened 2023-05-18
Drop dependencies for cols needs a CASCADE 

### LINKED ISSUE #818: [Bug] Drop dependencies for cols needs a CASCADE
We are using `alter_table_cascade=True` in a `rs.upsert()`. It is failing because some of the dependent views have their own dependent views. We need to add CASCADE to this line in the code: https://github.com/move-coop/parsons/blob/main/parsons/databases/redshift/redshift.py#L1090

elyse-weiss:
Closes #818

### DIFF
--- a/parsons/databases/redshift/redshift.py
+++ b/parsons/databases/redshift/redshift.py
@@ -1087,7 +1087,7 @@ class Redshift(
             tbl = self.query_with_connection(sql_depend, connection)
             dropped_views = [row["table_name"] for row in tbl]
             if dropped_views:
-                sql_drop = "\n".join([f"drop view {view};" for view in dropped_views])
+                sql_drop = "\n".join([f"drop view {view} CASCADE;" for view in dropped_views])
                 tbl = self.query_with_connection(sql_drop, connection)
                 logger.info(f"Dropped the following views: {dropped_views}")
 


Jason94 (approved):

shaunagm:
Looks like there's an issue with linting. You can now run a single black command to fix linting issues. Unfortunately the docs are not up-to-date with our process because we haven't cut a release lately, but you can see the commands in the [most recent docs](https://move-coop.github.io/parsons/html/latest/contributing.html#linting):

> flake8 --max-line-length=100 --extend-ignore=E203,W503 parsons
> black parsons

elyse-weiss:
@Jason94 let me know how I can point this toward the major release

shaunagm:
Let's get this landed. I think per our discussion on Slack, this doesn't actually need to target the major release. @elyse-weiss, can you confirm that the method you're altering, `drop_dependencies_for_cols`, is not exposed in the documentation, and that the methods that _do_ call it say that the drop cascades, even though it currently doesn't?

elyse-weiss:
Yes, this change is in line with current documentation.

elyse-weiss:
done! should I go ahead and merge?

shaunagm:
Yup, we're good to go!

PR #822, opened 2023-05-22
Pass use_env_token to to_s3_csv

codygordon:
This should extend the fix to functions using `to_s3_csv` (when False is passed) for the error which only appears in Lambda prod (using the same keys as when running locally):

```
S3UploadFailedError: Failed to upload /tmp/[...].zip to ephemeral-jobs-internal/[...].zip: An error occurred (InvalidToken) when calling the PutObject operation: The provided token is malformed or otherwise invalid.
```


codygordon:
@crayolakat flagging for review

crayolakat (approved):

PR #823, opened 2023-05-22
Add NGPVAN bulk apply suppressions function

codygordon:
Simply adds the `Contacts` -> `Suppressions` bulk import mapping type. 

Note, the `Suppressions` mapping type will need to be enabled for your API key via support request for this to work.

QA'd and working on our end.


codygordon:
@crayolakat flagging for review (I can't seem to add reviewers)

crayolakat (approved):

PR #825, opened 2023-05-25
Specify version of slackclient in setup.py

crayolakat:
When running `pip install parsons[slack]`, the latest version of slackclient is installed (2.9.4). The version of slackclient specified in `requirements.txt` is 1.3.0: https://github.com/move-coop/parsons/blob/853b27cb14394daaf113b88718facfde1190a572/requirements.txt#L6

In slackclient 2.9.4, the `slackclient` module has been renamed, causing all references to `parsons/notifications/slack.py` to return the error `ModuleNotFoundError: No module named 'slackclient'`. https://github.com/move-coop/parsons/blob/853b27cb14394daaf113b88718facfde1190a572/parsons/notifications/slack.py#L7

This PR specifies the slackclient version in `setup.py` to be the same version specified in `requirements.txt`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ def main():
             "s3": ["boto3"],
             "salesforce": ["simple-salesforce"],
             "sftp": ["paramiko"],
-            "slack": ["slackclient"],
+            "slack": ["slackclient==1.3.0"],
             "smtp": ["validate-email"],
             "targetsmart": ["xmltodict"],
             "twilio": ["twilio"],
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ def main():
             "s3": ["boto3"],
             "salesforce": ["simple-salesforce"],
             "sftp": ["paramiko"],
-            "slack": ["slackclient==1.3.0"],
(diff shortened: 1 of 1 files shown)


shaunagm:
Thanks, @crayolakat! Tagging in @cmc333333 to help review this one.

setup.py
```diff
@@ -41,7 +41,7 @@ def main():
             "s3": ["boto3"],
             "salesforce": ["simple-salesforce"],
             "sftp": ["paramiko"],
-            "slack": ["slackclient"],
+            "slack": ["slackclient==1.3.0"],
```
cmc333333, line 44: Thanks @crayolakat! `setup.py` files should be as loose with versions as possible (to increase compatibility), so does this look right? Should work with anything >= 1 but less than 2
```suggestion
            "slack": ["slackclient~=1"],
```

crayolakat: Yes, thank you! I committed the suggestion.

shaunagm (approved):
Sorry for the delay, @crayolakat, I meant to do this after the meeting last week and forgot.

PR #826, opened 2023-05-26
Action Builder connector

ydamit:
A very initial submission of a class for an Action Builder connection. Woefully bare bones, this is the minimum viable product necessary for a working Action Builder template sync in the TMC environment.

### DIFF
--- a/parsons/action_builder/action_builder.py
+++ b/parsons/action_builder/action_builder.py
@@ -4,6 +4,8 @@ from parsons.utilities import check_env
 from parsons.utilities.api_connector import APIConnector
 import logging
 
+logger = logging.getLogger(__name__)
+
 API_URL = "https://{subdomain}.actionbuilder.org/api/rest/v1"
 
 
@@ -183,10 +185,10 @@ class ActionBuilder(object):
                 retrieved or edited. Not necessary if supplied when instantiating the class.
         `Returns:`
             Dict containing Action Builder entity data.
-        """
+        """ # noqa: E501
 
         # Check that we have appropriate entity type/identifier, name, and campaign first
-        if entity_type == identifiers == None:
+        if entity_type == identifiers is None:
             error_msg = "Must provide either entity_type (to insert a new record) "
             error_msg += "or identifiers (to update an existing record)"
             raise ValueError(error_msg)
@@ -317,7 +319,7 @@ class ActionBuilder(object):
                 retrieved or edited. Not necessary if supplied when instantiating the class.
         `Returns:`
             Dict containing Action Builder connection data.
-        """
+        """ # noqa: E501
 
         # Check that there are exactly two identifiers and that campaign is provided first
         if not isinstance(identifiers, list):
--- a/parsons/action_builder/action_builder.py
+++ b/parsons/action_builder/action_builder.py
@@ -7,6 +7,17 @@ import logging
 API_URL = 'https://{subdomain}.actionbuilder.org/api/rest/v1'
 
 class ActionBuilder(object):
+    """
+    `Args:`
+        api_token: str
+            The OSDI API token
+        subdomain: str
+            The part of the UI URL preceding '.actionbuilder.org'
+        campaign: str
+            Optional. The 36-character "interact ID" of the campaign whose data is to be retrieved
+            or edited. Can also be supplied in individual methods in case multiple campaigns need
+            to be referenced.
+    """
     
     def __init__(self, api_token=None, subdomain=None, campaign=None):
         self.api_token = check_env.check('ACTION_BUILDER_API_TOKEN', api_token)
@@ -64,16 +75,59 @@ class ActionBuilder(object):
                     return Table(return_list[0:limit])
     
     def get_campaign_tags(self, campaign=None, limit=None, per_page=25, filter=None):
+        """
+        Retrieve all tags (i.e. custom field values) within provided limit and filters
+        `Args:`
+            campaign: str
+                Optional. The 36-character "interact ID" of the campaign whose data is to be
+                retrieved or edited. Not necessary if supplied when instantiating the class.
+            limit: int
+                The number of entries to return. When None, returns all entries.
+            per_page: int
+                The number of entries per page to return. 25 maximum and default.
+            filter
+                The OData query for filtering results. E.g. "modified_date gt '2014-03-25'".
+                When None, no filter is applied.
+        `Returns:`
+            Parsons Table of full set of tags available in Action Builder.
+        """
         
         return self._get_entry_list(campaign, 'tags', limit=limit, per_page=per_page, filter=filter)
     
     def get_tag_by_name(self, tag_name, campaign=None):
+        """
+        Convenience method to retrieve data on a single tag by its name/value
+        `Args:`
+            tag_name: str
+                The value of the tag to search for.
+            campaign: str
+                Optional. The 36-character "interact ID" of the campaign whose data is to be
+                retrieved or edited. Not necessary if supplied when instantiating the class.
+        `Returns:`
+            Parsons Table of data found on tag in Action Builder from searching by name.
+        """
         
         filter = f"name eq '{tag_name}'"
         
         return self.get_campaign_tags(campaign=campaign, filter=filter)
     
     def insert_new_tag(self, tag_name, tag_field, tag_section, campaign=None):
+        """
+        Load a new tag value into Action Builder. Required before applying the value to any entity
+        records.
+        `Args:`
+            tag_name: str
+                The name of the new tag, i.e. the custom field value.
+            tag_field: str
+                The name of the tag category, i.e. the custom field name.
+            tag_section: str
+                The name of the tag section, i.e. the custom field group name.
+            campaign: str
+                Optional. The 36-character "interact ID" of the campaign whose data is to be
+                retrieved or edited. Not necessary if supplied when instantiating the class.
+        `Returns:`
+            Dict containing Action Builder tag data.
+        """
         
         campaign = self._campaign_check(campaign)
         url = f'campaigns/{campaign}/tags'
@@ -87,6 +141,30 @@ class ActionBuilder(object):
         return self.api.post_request(url=url, data=json.dumps(data))
     
     def upsert_entity(self, entity_type=None, identifiers=None, data=None, campaign=None):
+        """
+        Load or update an entity record in Action Builder based on whether any identifiers are
+        passed.
+        `Args:`
+            entity_type: str
+                The name of the record type being inserted. Required if identifiers are not
+                provided.
+            identifiers: list
+                The list of strings of unique identifiers for a record being updated. ID strings
+                will need to begin with the origin system, followed by a colon, e.g.
+                `action_builder:abc123-...`.
+            data: dict
+                The details to include on the record being upserted, to be included as the value
+                of the `person` key. See
+                [documentation for the Person Signup Helper](https://www.actionbuilder.org/docs/v1/person_signup_helper.html#post)
+                for examples, and
+                [the Person endpoint](https://www.actionbuilder.org/docs/v1/people.html#field-names)
+                for full entity object composition.
+            campaign: str
+                Optional. The 36-character "interact ID" of the campaign whose data is to be
+                retrieved or edited. Not necessary if supplied when instantiating the class.
+        `Returns:`
+            Dict containing Action Builder entity data.
+        """ # noqa: E501
         
         if {entity_type, identifiers} == {None}:
             error_msg = 'Must provide either entity_type (to insert a new record) '
@@ -119,6 +197,27 @@ class ActionBuilder(object):
         return self.api.post_request(url=url, data=json.dumps(data))
     
     def add_tags_to_record(self, identifiers, tag_name, tag_field, tag_section, campaign=None):
+        """
+        Add a tag (i.e. custom field value) to an existing entity record in Action Builder. The
+        tag, along with its category/field and section/group, must already exist unless it is a
+        date field.
+        `Args:`
+            identifiers: list
+                The list of strings of unique identifiers for a record being updated. ID strings
+                will need to begin with the origin system, followed by a colon, e.g.
+                `action_builder:abc123-...`.
+            tag_name: str
+                The name of the new tag, i.e. the custom field value.
+            tag_field: str
+                The name of the tag category, i.e. the custom field name.
+            tag_section: str
+                The name of the tag section, i.e. the custom field group name.
+            campaign: str
+                Optional. The 36-character "interact ID" of the campaign whose data is to be
+                retrieved or edited. Not necessary if supplied when instantiating the class.
+        `Returns:`
+            Dict containing Action Builder entity data of the entity being tagged.
+        """
         
         # Ensure all tag args are lists
         tag_name = tag_name if isinstance(tag_name, list) else [tag_name]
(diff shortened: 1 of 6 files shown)


parsons/action_builder/action_builder.py
```diff
@@ -0,0 +1,351 @@
...
+
+        data = {
+            "name": tag_name,
+            "action_builder:field": tag_field,
+            "action_builder:section": tag_section,
+        }
+
+        return self.api.post_request(url=url, data=json.dumps(data))
+
+    def upsert_entity(
```
willyraedy, line 162: My instinct is to make this an internal function and have the public api specify which entity is being updated. Might make the function validation simpler as well

parsons/action_builder/action_builder.py
```diff
@@ -0,0 +1,351 @@
...
+            data["person"]["action_builder:entity_type"] = entity_type
+
+        return self.api.post_request(url=url, data=json.dumps(data))
+
+    def add_tags_to_record(
+        self, identifiers, tag_name, tag_field, tag_section, campaign=None
+    ):
+        """
+        Add a tag (i.e. custom field value) to an existing entity record in Action Builder. The
+        tags, along with their category/field and section/group, must already exist (except for
```
willyraedy, line 237: is there a reason not to create one now if they don't already exist?

parsons/action_builder/action_builder.py
```diff
@@ -0,0 +1,351 @@
...
+                Optional. The 36-character "interact ID" of the campaign whose data is to be
+                retrieved or edited. Not necessary if supplied when instantiating the class.
+        `Returns:`
+            Dict containing Action Builder entity data of the entity being tagged.
+        """
+
+        # Ensure all tag args are lists
+        tag_name = tag_name if isinstance(tag_name, list) else [tag_name]
+        tag_field = tag_field if isinstance(tag_field, list) else [tag_field]
+        tag_section = tag_section if isinstance(tag_section, list) else [tag_section]
```
willyraedy, line 264: what would happen if someone provided data as the wrong type. I'm thinking like a tuple which would then become a list of tuples. If we passed that to the api, would we get really weird tags or would it fail?

parsons/action_builder/action_builder.py
```diff
@@ -0,0 +1,351 @@
...
+        # Raise an error if there are fewer specific items provided than generic
+        if sorted_keys[0] != "name":
+            raise ValueError(
+                "Not enough tag_names provided for tag_fields or tag_sections"
+            )
+
+        if sorted_keys[1] != "field":
+            raise ValueError("Not enough tag_fields provided for tag_sections")
+
+        # Construct tag data, applying final generic taxonomy to any excess specific items
```
willyraedy, line 289: I'm not sure I'm following but this behavior doesn't seem intuitive to me and could lead to some surprising data.

parsons/action_builder/action_builder.py
```diff
@@ -0,0 +1,351 @@
...
+                will need to begin with the origin system, followed by a colon, e.g.
+                `action_builder:abc123-...`.
+            tag_name: str or list
+                The name(s) of the new tag(s), i.e. the custom field value(s).
+            tag_field: str or list
+                The name(s) of the tag category(ies), i.e. the custom field name(s). Must have
+                equal or fewer items as `tag_name`.
+            tag_section: str or list
+                The name(s) of the tag section(s), i.e. the custom field group name(s). Must have
+                equal or fewer items as `tag_section`.
```
willyraedy, line 253: I may not be understanding the data structure here, but I wonder if it would be easier to accept a list of dictionaries:

[
   { tag_name: 'foo', tag_fields: ['bar', 'baz'], tag_sections: ['candy', 'cane'] },
   ...
]

willyraedy: seems like that might make the parsing trivial and would force the user to be explicit about what relationships they wanted between names, fields, sections

parsons/action_builder/action_builder.py
```diff
@@ -0,0 +1,351 @@
...
+                "action_builder:section": tag_section[min(i, len(tag_section) - 1)],
+            }
+            for i, x in enumerate(tag_name)
+        ]
+
+        data = {"add_tags": tag_data}
+
+        return self.upsert_entity(identifiers=identifiers, data=data, campaign=campaign)
+
+    def upsert_connection(self, identifiers, tag_data=None, campaign=None):
```
willyraedy, line 303: unless there's an insane number of entities and therefore types of connections, would also lean towards having a more explicit public api of like `upsert_person_campaign_connection` and using this as a private method to keep things dry

test/test_action_builder/test_action_builder.py
```diff
@@ -0,0 +1,345 @@
+import unittest
+import requests_mock
+import json
+from parsons import Table, ActionBuilder
+from test.utils import assert_matching_tables
+
+
+class TestActionBuilder(unittest.TestCase):
```
willyraedy, line 8: Tests!

test/test_action_builder/test_action_builder.py
```diff
@@ -0,0 +1,345 @@
...
+
+        self.fake_tag_1 = "Fake Tag 1"
+        self.fake_tag_2 = "Fake Tag 2"
+        self.fake_tag_3 = "Fake Tag 3"
+        self.fake_tag_4 = "Fake Tag 3"
+
+        self.fake_field_1 = "Fake Field 1"
+        self.fake_section = "Fake Section 1"
+
+        self.fake_tags_list_1 = {
```
willyraedy, line 33: this setup gets pretty involved. you might consider some generator functions that accept args and spit out these more complicated objects

test/test_action_builder/test_action_builder.py
```diff
@@ -0,0 +1,345 @@
...
+                for i in range(len(val)):
+                    print(i)
+                    insub, compsub = self.compare_to_incoming(compare[key][i], val[i])
+                    incoming[key][i] = insub
+                    compare[key][i] = compsub
+
+        return incoming, compare
+
+    @requests_mock.Mocker()
+    def test_upsert_entity(self, m):
```
willyraedy, line 289: there's a lot going on in this test along with the `compare_to_incoming`. I'm not sure I really follow what's happening so might be easier to grab 10 min and talk through it

test/test_action_builder/test_action_builder.py
```diff
@@ -0,0 +1,345 @@
...
+    def tagging_callback(self, request, context):
+        # Internal method for returning the constructed tag data to test
+
+        post_data = request.json()
+        tagging_data = post_data["add_tags"]
+        return tagging_data
+
+    @requests_mock.Mocker()
+    def test_add_tags_to_record(self, m):
+        m.post(f"{self.api_url}/people", json=self.tagging_callback)
```
willyraedy, line 320: interesting! I've not used requests_mock. is this mocking the json method? what is the `request` param that the `tagging_callback` is receiving?

willyraedy (commented):
Exciting! Seems like a killer connector. 

A few places where I think we could make the API a little more user friendly and want to make sure the structuring of tags doesn't lead to unexpected results. And then would love to see if there's a way to simplify one of the tests.

parsons/action_builder/action_builder.py
```diff
@@ -0,0 +1,351 @@
...
+        subdomain: str
+            The part of the web app URL preceding '.actionbuilder.org'
+        campaign: str
+            Optional. The 36-character "interact ID" of the campaign whose data is to be retrieved
+            or edited. Can also be supplied in individual methods in case multiple campaigns need
+            to be referenced.
+    """
+
+    def __init__(self, api_token=None, subdomain=None, campaign=None):
+        self.api_token = check_env.check("ACTION_BUILDER_API_TOKEN", api_token)
```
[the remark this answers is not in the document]
ydamit: I might be reading it wrong, but I _think_ that's the default behavior of the `check_env()` [utility](https://github.com/move-coop/parsons/blob/main/parsons/utilities/check_env.py)?

parsons/action_builder/action_builder.py
ydamit: I agree, though perhaps a little grudgingly. Both Action Network and Action Builder use a common endpoint for inserting and updating: the signup helper. Action Network has much simpler upsert logic, because records are unique on additional data: email and/or mobile number. Action Builder has no such uniqueness, so you're right that these should be separate. I'm just grumbly because I need to make sure this change is reflected in the sync that's live, but that's on me for running a sync on branch like that, before the connector is even merged.

test/test_action_builder/test_action_builder.py
ydamit: I'm also just wrapping my head around it. The idea as I understand it is to intercept specific HTTP requests and force a desired response. This is done by setting any of the properties available in a `requests` response object. So you'll notice that most of the tests above just return a `text`, and to fit that expectation, it's `json.dumps()` applied to whatever Python object was prepared for the purpose of that test.

In more complicated cases such as this test, we can pass `request` (the details of the request) and `context` (the details of the response) to a callback function, to return something that requires more processing. Specifically here, we're extracting just the list of dicts that should appear under the `add_tags` key of the top-most dict in the request data, and returning that to compare to what we expect that tagging data to look like.

test/test_action_builder/test_action_builder.py
ydamit: Happy to do that. In a nutshell, though, `compare_to_incoming()` is just trying to make sure to exclude out any keys (including in nested dicts) from an incoming dict that don't also appear on the existing dict to which the incoming dict is to be compared. In other words, this if an incoming dict has extra keys anywhere, don't include them in the comparison.

test/test_action_builder/test_action_builder.py
ydamit: You're probably right, but I'm hoping it's ok to save that for the next version of updates to this connector.

parsons/action_builder/action_builder.py
ydamit: I'm not sure I completely understand the suggestion here. Just in case there's any misunderstanding about connections and how they're applied, I'll try to summarize here.

Connections can exist between any two entity records, regardless of whether the entity type is person or not. The system is strict about there only ever being one connection record between two specific entity records, and the connection helper endpoint is pretty much the only way to create or update connections (other than the signup helper, which doesn't really work differently in this regard). So you don't actually ever know for sure (unless you make wasteful additional API calls) whether you're inserting or updating.

Connection records can accept tags, much the same way as entity records. These can help illuminate important information about the relationship between entities. For instance, between two people records, you could have a tag indicate the human relationship between them, or on a connection between a person and an organization, you could have one that indicates the person's role at that organization.

Given all this, I have a hard time imagining what public method we could write that would be more intuitive than this one, and use this one as internal. But very open to anything I'm just not seeing!

parsons/action_builder/action_builder.py
ydamit: I'm willing to be swayed here, but I wanted to name a bit more of the intent first. This is meant as a convenience method, because someone who is more comfortable with the JSON needed to apply tags could just include that in the `data` provided to `_upsert_entity()`. It's a list of dicts to be added under a key called `add_tags` at the same level as `person`.

So my reasoning is that someone would only use this method if they want a more intuitive approach to adding tags than constructing that JSON. It's also annoying because every tag submitted must have a field and a section in this JSON data as well, and I wouldn't be surprised if it's often the case that all tags applied this way are in the same section.

The current logic I have is that if the lists provided are not the same length, the last field/section continues to be applied for any remaining tags. That way, you could just provide one section, and it would apply to all the tags. If that's _not_ intuitive, then I definitely haven't achieved my goal here, but I'm worried that asking for a list of dicts would essentially make this method no longer any more convenient than expecting that to be added to the `data` in any of the upsert family of methods...

parsons/action_builder/action_builder.py
ydamit: It really should fail, because the resulting data would essentially be attempting to apply a tag that doesn't exist.

parsons/action_builder/action_builder.py
ydamit: As above, if there isn't a way to make this more intuitive than requiring a list of dicts, I'm not sure this method provides any added convenience.

parsons/action_builder/action_builder.py
ydamit: I can imagine use-cases where someone would want to prevent that in case of bad data, so I think we should leave leave that for the logic used by folks writing to this connector, personally.

parsons/action_builder/action_builder.py
```diff
@@ -0,0 +1,351 @@
...
+    ):
+        # Returns a list of entries for a given object, such as people, tags, or connections.
+        # See Action Builder API docs for more: https://www.actionbuilder.org/docs/v1/index.html
+
+        count = 0
+        page = 1
+        return_list = []
+
+        # Keep getting the next page until record limit is exceeded or an empty result returns
+        while True:
```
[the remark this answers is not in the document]
ydamit: No objection from me, just so long as they can handle OSDI weirdness!

parsons/action_builder/action_builder.py
```diff
@@ -0,0 +1,351 @@
...
+            )
+
+        params = {"page": page, "per_page": per_page, "filter": filter}
+
+        campaign = self._campaign_check(campaign)
+        url = f"campaigns/{campaign}/{object_name}"
+
+        return self.api.get_request(url=url, params=params)
+
+    def _get_entry_list(
```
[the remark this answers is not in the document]
ydamit: Good point. I copied it pretty much wholesale [from the Action Network connector](https://github.com/move-coop/parsons/blob/main/parsons/action_network/action_network.py#L40), so we may eventually want to change that one, too.

parsons/action_builder/action_builder.py
```diff
@@ -0,0 +1,351 @@
...
+        if per_page > 25:
+            per_page = 25
+            logger.info(
+                "Action Builder's API will not return more than 25 entries per page. \
+            Changing per_page parameter to 25."
+            )
+
+        params = {"page": page, "per_page": per_page, "filter": filter}
+
+        campaign = self._campaign_check(campaign)
```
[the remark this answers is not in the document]
ydamit: Completely agree!

parsons/action_builder/action_builder.py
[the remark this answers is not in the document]
willyraedy: you are correct! the order of the arguments confuses me but that's for another time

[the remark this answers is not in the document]
ydamit: Yeah, good point.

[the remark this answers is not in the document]
willyraedy: you are correct! the order of the arguments confuses me but that's for another time

parsons/action_builder/action_builder.py
willyraedy: I could also see an argument for both. Having convenience methods like `upsert_person` (or whatever it would be) that are just syntactic sugar on the more general `upsert_entity`. But I do think having the more semantically clear parts of the API is important

parsons/action_builder/action_builder.py
willyraedy: Interesting. I see what you're going for but I definitely did not intuit the last value continues to be applied to the remaining tags.

ydamit: Then I should definitely see what I can do about making that clearer in the comments/docstrings!

parsons/action_builder/action_builder.py
willyraedy: oh interesting. we may have missed each other. I think upsert is a better public api to use than an insert/update split. I was suggesting the _person, _campaign versions. 

we might be reaching the limits of asyncronous review and may need to grab some time

parsons/action_builder/action_builder.py
ydamit: Ok, just pushed some updates to comments and docstrings, in case they help

test/test_action_builder/test_action_builder.py
shaunagm: It's worth thinking about creating some general guidance for implementing factory functions to create test data (and maybe even make some default factory functions that create common objects that people can subclass/build off of). Right now most connectors rely on creating json or python dicts, in the tests like Yotam's done here or imported from separate files, and either way it's messy and tedious to write and generally not ideal. 

But factory methods are a relatively advanced python concept so we haven't wanted to say "just do this" without giving contributors support. If we can make some good guidance/templates and some default classes, we could slowly transition over connectors (especially highly used connectors, and new connectors) to use factory functions and that would probably be better for the project/community overall.

ydamit:
@willyraedy , I think I've incorporated everything we discussed today. Let me know, though, if I left anything out or if you notice anything new that requires attention!

sharinetmc:
Hi @willyraedy, it looks like yotam made some changes par your comments. Wondering if you had any other suggestions for him for review!

sharinetmc:
Hi @willyraedy! Just wondering if there are too many things on your plate at the moment, if I should just go over this PR with Kasia tomorrow during our pairing! Thanks!

willyraedy:
@sharinetmc Looking at it now!

willyraedy (approved):
Looks good! Digging the new APIs and tests are much simpler.

Apologies I lost track of this. Thought you had the green light from our convo

shaunagm:
Looks like this is ready to merge, so I'm going to merge it. Thank you for the contribution @ydamit!

PR #827, opened 2023-05-27
added encoding to s3_copy function in redshift class

sharinetmc:


shaunagm (changes requested):
Need to be linted by running Black but otherwise should be ready to go (relying on Willy's review in the comments)

shaunagm:
So the instructions for how to use Black in the docs are wrong, whoops. Can you try `black --check parsons/ test/ useful_resources/`

shaunagm:
@sharinetmc can you double-check that I resolved the merge conflict correctly and that the function does what it should? https://github.com/move-coop/parsons/pull/827/commits/0fb5ab3ee3d21e9e72a59a81e6b9ed93c1888e87

If you can confirm that, we can merge this

sharinetmc:
@shaunagm, tests are all passing now and everything looks good!

shaunagm (approved):
LGTM

PR #829, opened 2023-06-02
 Fix Salesforce query output

sjwmoveon:
The existing code did not match the method description and did not produce a useful result. Salesforce returns a set of nested dicts, similar to JSON; the new method reformats the nested ordered dicts using the json library, similar to describe_fields.

### DIFF
--- a/parsons/salesforce/salesforce.py
+++ b/parsons/salesforce/salesforce.py
@@ -80,9 +80,9 @@ class Salesforce:
             list of dicts with Salesforce data
         """  # noqa: E501,E261
 
-        q = Table(self.client.query_all(soql))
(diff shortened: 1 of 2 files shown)


crayolakat:
@sjwmoveon Can you modify the tests to pass with this new change?

crayolakat (changes requested):
@sjwmoveon Can you modify the tests to pass with this new change?

sjwmoveon:
Yes, thank you! Since the tests mocked everything out I assumed they would pass; clearly a bad assumption on my part.

sjwmoveon:
It passed locally, I'm not sure what's wrong here.

crayolakat (approved):

PR #830, opened 2023-06-05
Enable __len__ dunder method for parsons Table

austinweisgrau:
This enables the pythonic use of len(Table) rather than needing to use the idiosyncratic Table.num_rows. len(Table) calls Table.num_rows


shaunagm (approved):
Thanks for this - great catch

PR #832, opened 2023-06-06
Revert "Drop dependencies for cols needs a CASCADE "

elyse-weiss:
Reverts move-coop/parsons#819


shaunagm (approved):
LGTM, let's revert this and fix it

PR #842, opened 2023-06-16
Implement, test, and document Parsons Table deduplicate method

### LINKED ISSUE #820: Add "deduplicate" helper method
It would be good to expose basic functionality on the Parsons table to allow people to deduplicate their data. Currently a lot of folks convert the Parsons table to a Pandas dataframe to accomplish this which seems wasteful both in terms of developer times and also could potentially impact performance:

I'm imagining something that looks like this:

`parsons_table.deduplicate(keys, sort)`

Where you can give no keys, one key, or several to deduplicate by, as well as indicating whether the data is already sorted or should be sorted (basically a wrapper on petl's `presorted` parameter.

Essentially it would be a wrapper around the petl function that accomplish this, ie:

```
from petl import etl
from parsons import table

people = people.to_petl()
result = etl.transform.dedup.distinct(people)
Table(result)
```
Medium priority - this seems like it would be very helpful but also it's not blocking anything.

jafayer:
This PR resolves #820.

# Summary of changes:

`parsons/etl/etl.py`: adds a `deduplicate` method which optionally accepts a key or list of keys to deduplicate on, and an optional flag to indicate whether the method should first sort the table.
`docs/table.rst`: adds deduplicate method to Table documentation.
`test/test_etl.py`: adds several deduplicate unit tests for various scenarios involving zero, one, or multiple keys on both sorted and unsorted tables.

This method is essentially a wrapper around the built-in [petl.transform.dedup.distinct](https://petl.readthedocs.io/en/stable/transform.html#petl.transform.dedup.distinct) method and extends the functionality natively to the Parsons Table class. This solution also heavily references the implementation provided in the original issue by @shaunagm!

## One sticky point on terminology
As suggested in the original issue, this method accepts an optional `sort` parameter to instruct petl (imperatively) whether to first sort the table. However, this is more or less the opposite definition used internally by the petl library. In petl, the same argument is called `presorted` which (declaratively) describes a table as either presorted or not presorted &mdash; if it's not presorted, petl sorts it.

It's a small detail, but anyone who is used to the petl definition may be confused by this. A `sort=True` value (meaning, "this table is not yet sorted, sort it for me") is the exact opposite of a `presorted=True` value in petl (meaning, "this table is sorted, don't sort it"). In the code, this is represented by an explicit inversion (`presorted=not sort`) on line 1186 when it's passed into petl's method. I've also included a comment explaining this for anyone who might be confused by the definitions.

For one, I totally agree that "sort" makes more sense here to instruct the deduplicate method to sort the array, but I'm newer to the community and wanted to note here that it might run against what someone might expect if they're used to interacting with petl and Parsons Tables directly.

Happy to adjust any of the above!

### DIFF
--- a/docs/table.rst
+++ b/docs/table.rst
@@ -196,6 +196,8 @@ of commonly used methods. The full list can be found in the API section.
       - Divide tables into smaller tables based on row count
     * - :py:meth:`~parsons.etl.etl.ETL.remove_null_rows`
       - Removes rows with null values in specified columns
+    * - :py:meth:`~parsons.etl.etl.ETL.deduplicate`
+      - Removes duplicate rows based on optional key(s), and optionally sorts
 
 
 **Extraction and Reshaping**
--- a/parsons/etl/etl.py
+++ b/parsons/etl/etl.py
@@ -1168,30 +1168,91 @@ class ETL(object):
 
         return Table(getattr(petl, petl_method)(self.table, *args, **kwargs))
 
-    def deduplicate(self, keys=None, sort=False):
+    def deduplicate(self, keys=None, presorted=False):
         """
         Deduplicates table based on an optional ``keys`` argument,
         which can contain any number of keys or None.
 
+        Method considers all keys specified in the ``keys`` argument
+        when deduplicating, not each key individually. For example,
+        if ``keys=['a', 'b']``, the method will not remove a record
+        unless it's identical to another record in both columns ``a`` and ``b``.
+
+        .. code-block:: python
+
+            >>> tbl = Table([['a', 'b'], [1, 3], [1, 2], [1, 2], [2, 3]])
+            >>> tbl.table
+            +---+---+
+            | a | b |
+            +===+===+
+            | 1 | 3 |
+            +---+---+
+            | 1 | 2 |
+            +---+---+
+            | 1 | 2 |
+            +---+---+
+            | 2 | 3 |
+            +---+---+
+
+            >>> tbl.deduplicate('a')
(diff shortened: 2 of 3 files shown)


shaunagm:
Amazing, thank you Josh! I'm just about done for the day but will test this out early next week.

FYI, the build fail seems to be a linting error:

Screenshot from 2023-06-16 16-48-01

jafayer:
Ahh apologies, it looks like I only ran the linter on the parsons directory. I've corrected the error locally and run the linters again but before I add it to the PR, just want to make sure I'm following whatever conventions work best for the community.

Would you prefer I rebase/squash the change into the initial commit or make another commit on this PR to preserve the history?

Thank you!!

shaunagm (changes requested):
This looks great. Thanks for flagging the terminology issue - while my instinct was that `sort` should be the keyword, it's probably better to stay consisted with petl and with the one other time we use this functionality within Parsons, in [reduce rows](https://github.com/move-coop/parsons/blob/b8f41b5b706182349bb9e5745f738fa68b6afd70/parsons/etl/etl.py#L999). So please do make that switch. 

I think we can also use a bit more information in the docstring on how exactly the keys argument works. Like, just being super clear that if you put in a key, it will search for duplicates only in that key, if you put in multiple it will search for duplicates of the set of keys, not each key individually, and if you put in none it will search for duplicates across all columns. But less clumsy than I just said it. :)

shaunagm (approved):
This looks great! I'm so excited to see this land. :)  Thank you again!

PR #845, opened 2023-06-22
add script

sharinetmc:


shaunagm (changes requested):
This looks good - the only issue is the linting. Did you run Black?

shaunagm (changes requested):
Looks like linting is still failing?

sharinetmc:
Hmm...interesting. I just ran it again and it said that the file was left unchanged.

shaunagm:
Can you share the full output of the Black command, as well as the location you're running the command from?

sharinetmc:
(parsons) ➜  parsons git:(sample_ngpvan_printed_list) black useful_resources/sample_code/ngpvan_sample_printedlist.py
All done! ✨ 🍰 ✨
1 file left unchanged.

shaunagm:
So the instructions for how to use Black in the docs are wrong, whoops. Can you try `black --check parsons/ test/ useful_resources/`

shaunagm:
Thanks for the help @bmos but this PR is now including 3 extra files which it shouldn't need to touch. Any idea why/how to fix?

bmos:
Afaik that is because it was needed to get the linting check to pass. You can see the checks on the two commits in my PR.

bmos:
Now the issue is probably the longer length target.

shaunagm:
It's probably easier for Sharine to just redo with a new PR off the most recent main branch. It's easy enough since it's a single file, and then she can run Black which will use whatever the current standards are.

PR #848, opened 2023-06-26
feat: add nation builder connector

gcollazo:
**NOTE:** This is the second attempt at merging this feature. The first attempt had issues when running under **Python 3.7.17**. This new PR makes sure that the class types are compatible with that version of Python.

---

This PR adds a new [Nation Builder](https://nationbuilder.com/developers) connector.

The connector implements the following public methods:

- `get_people`: make all necessary HTTP request to get all people and return a `Table`. If a request fails, it will wait 30 seconds and keep trying until it fetches everything.
- `update_person`: update a person by their Nation Builder `id`.
- `upsert_person`: update or create a person.

## Examples

The following examples help illustrate the API usage.

```python
from dotenv import load_dotenv

from parsons.nation_builder.nation_builder import NationBuilder

load_dotenv()

nb = NationBuilder()

# get all people and return a Table
people_table = nb.get_people()
people_table.to_csv("nbexport.csv")

# update person by id
data = {"email": "foo@example.com", "tags": ["foo", "bar"]}
updated_person_1 = nb.update_person("123456789", data)

# update or create a person
data = {"email": "foo@example.com", "tags": ["dogs", "cats"]}
created, person = nb.upsert_person(data)

if created:
    print("Person created:", person)
else:
    print("Person updated:", person)
```


shaunagm (approved):
This looks good to me. I'm going to try and set up testing on all Parsons-supported versions of Python as part of our CI process sometime in the next few weeks, so this won't happen again.

PR #856, opened 2023-07-17
Update setup.py with new release number

KasiaHinkson:


shaunagm (changes requested):
This is perfectly fine but the tests aren't passing. After some investigation, it's because Yotam's ActionBuilder PR was so old that it predated us added Black to the tests, thus the tests didn't include running Black. Which is a silly way for the CI system to work, it really ought to have used the most recent tests, not the tests at the time the PR was first opened. 

Anyway, if you can add a commit to fix those issues and stick 'em in this PR, or quickly do a separate PR and rebase, that'd be great. It feels messy to cut a release where the tests don't pass.

shaunagm (approved):
Linting issue fixed! :)

PR #857, opened 2023-07-17
NewMode - Add bulk `targets` endpoint

IanRFerguson:
This PR adds an additional `GET` request to the NewMode client that allows ALL targets to be queried and returned as a table. The addition of this function will give Parsons users flexibility in their API requests to NewMode, as they may be interested in exploring target data broadly and across outreaches.

---
  - https://app.asana.com/0/0/1204935310402358

### DIFF
--- a/parsons/newmode/newmode.py
+++ b/parsons/newmode/newmode.py
@@ -165,6 +165,27 @@ class Newmode:
             logging.warning("Empty target returned")
             return None
 
+    def get_targets(self, params={}):
+        """
+        Get all targets
+
+        Args:
+            params dict:
+                Extra paramaters sent to New/Mode library
+
+        Returns:
+            Target information
+        """
+
+        targets = self.client.getTargets(params=params)
+
+        if targets:
+            return self.convert_to_table(targets)
+
+        else:
+            logging.warning("No targets returned")
+            return None
+
     def get_campaigns(self, params={}):
         """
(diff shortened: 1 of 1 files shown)


willyraedy:
@ydamit looks like there might be some linting / formatting errors from the action builder connector. Would you mind taking a look at that so we can get the build to run here?

ydamit:
@willyraedy , I'll do my best! Am I using criteria from black or flake8?

willyraedy:
> @willyraedy , I'll do my best! Am I using criteria from black or flake8?

If you click the "Details" link in the build step above, it should show you the errors. 

image

shaunagm:
If you update your branch, it should get rid of those linting errors since they've been fixed on `main` as of this morning.

ydamit:
@willyraedy , gotcha. So it sounds like we're just moving over to black, since it looks like part of the issue is that some of the docstrings that go long are decorated with `noqa` for flake8 purposes, but black doesn't seem to like that?

ydamit:
Oh nevermind, then!

willyraedy:
@IanRFerguson did you see that? Sounds like you just need to re-merge latest from main here

IanRFerguson:
@shaunagm @willyraedy Checks passing! ready for another look when y'all have time

willyraedy (approved):
LGTM

Issue #790, opened 2023-02-21
Add CI installation + import tests for all supported versions and all major environments


shaunagm:
Given that Parsons can be tricky to install, I'd like to add to our tests a step where Parsons is installed on a variety of systems - covering all major OSs and versions. We have also had some issues recently with not testing out all supported versions.

## Possible Implementation

In addition to the work itself, the main question is which versions to test. We can probably poll the community to find the most popular OSs and versions of things.

## Priority

High - this will likely help reliability a lot.

shaunagm:
This is mostly done, although we're waiting on a couple of bugfixes brought to light by adding Windows testing in order to actually add the windows tests (see #860)

PR #858, opened 2023-07-26
Add multiple python versions to CI tests

shaunagm:
We had a pair programming session where we attempted to address #790 within CircleCI, that was unsuccessful so we switched to Github Actions.

- [x] get tests working in github actions
- [x] add caching
- [x] move Windows error to new issue #860 
  - see [this windows error](https://github.com/move-coop/parsons/actions/runs/5718559743/job/15494634900?pr=858)
- [x] check usage requirements to make sure we won't run out of minutes/storage
   - a single run of the tests (both Linux and Mac) is about 5 minutes, and we seem to have 2000 free minutes total, so we have about 400 runs per month or an average of 12-13 test runs per day average. that may not be enough but it may be fine? Buying additional minute seems to cost about $10 per thousand minutes so it's not a huge deal if we overrun occasionally
- [x] check that the docs build stuff is still working on circle CI


.github/workflows/test-linux-windows.yml
```diff
@@ -0,0 +1,50 @@
+name: tests
+
+on:
+  pull_request:
```
gcollazo, line 4: Consider adding "on push" so tests run once more once PRs are merged into main

```yml
on:
  push:
    branches: ["main"]
```

.github/workflows/tests-mac.yml
```diff
@@ -0,0 +1,42 @@
+name: tests for mac
+# test mac on single python version as mac tests use 10x minutes/storage
+
+on:
```
gcollazo, line 4: Consider adding "on push"

gcollazo (commented):
Thank you for moving this to GH actions 👍

A few comments:

- Tests not running after PR is merged into `main`. Consider adding "on push".
- For Mac it's only running on Python 3.8. Does that satisfy the requirements?
- Not testing withPython 3.7. Consider updating README, docs and setup.py to deprecate support of 3.7.

shaunagm:
Thanks for the feedback @gcollazo! I updated to address suggestions 1 & 3. Regarding question 2 - we have a set number of minutes we can use per month, and tests on Mac are given a 10x multiplier, so something that takes 3 minutes costs 30 minutes (as opposed to 3 minutes on Linux, and 6 minutes on Mac). I don't think we have enough free minutes to actually run that many Mac tests, so I'm hoping that just running on one version of Python will catch any Mac-specific errors.

PR #861, opened 2023-08-02
Enable passing `identifiers` to ActionNetwork `upsert_person()`

austinweisgrau:
`identifiers` is a useful method when upserting a person, for either matching to an existing contact or creating a new contact.

The `**kwargs` keyword arguments on upsert_person() are passed as custom fields and so excludes the ability to add identifiers as additional keyword arguments without adding this functionality.

Small code readability improvements are also made to the `get_tags()` method in this PR by removing code that didn't do anything.

### DIFF
--- a/parsons/action_network/action_network.py
+++ b/parsons/action_network/action_network.py
@@ -156,7 +156,11 @@ class ActionNetwork(object):
                 background queue for eventual processing.
                 https://actionnetwork.org/docs/v2/#background-processing
             identifiers:
-                list of strings to be used as globally unique identifiers
+                list of strings to be used as globally unique identifiers.
+                e.g.: [
+                   "action_network:d3e27e15-e5b5-4707-be96-8bc359462133",
+                   "foreign_system:1"
+                ]
             **kwargs:
                 Any additional fields to store about the person. Action Network allows
                 any custom field.
--- a/parsons/action_network/action_network.py
+++ b/parsons/action_network/action_network.py
@@ -100,6 +100,7 @@ class ActionNetwork(object):
         mobile_number=None,
         mobile_status="subscribed",
         background_processing=False,
+        identifiers=None,
         **kwargs,
     ):
         """
@@ -153,6 +154,8 @@ class ActionNetwork(object):
                 an immediate success, with an empty JSON body, and send your request to the
                 background queue for eventual processing.
                 https://actionnetwork.org/docs/v2/#background-processing
+            identifiers:
+                list of strings to be used as globally unique identifiers
             **kwargs:
                 Any additional fields to store about the person. Action Network allows
                 any custom field.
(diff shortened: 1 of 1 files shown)


shaunagm (changes requested):
Hi Austin! We've reviewed this as part of our group pairing session. If you can address the comments other made, that would be great.

I'm also not sure how passing identifiers changes the response (if at all) and whether it's the sort of change that might need to be tested. Can you provide a bit more explanation of what this change does? Thank you!

parsons/action_network/action_network.py
```diff
@@ -321,21 +325,15 @@ def update_person(self, entry_id, background_processing=False, **kwargs):
         logger.info(f"Person {entry_id} successfully updated")
         return response
 
-    def get_tags(self, limit=None, per_page=25, page=None):
```
jafayer, line 324: We may want to add `**kwargs` to the method signature to cover backwards compatibility if anyone had previously declared `per_page` or `page` on this method. Removing them otherwise might cause an "unexpected keyword argument" TypeError.

austinweisgrau: What do you think about this implementation? 22952c73aed29a729da3f3067654a9d9dfe6ab7b

If the page argument was used previously, it would have resulted in an Exception because it would have called a non-existent method. So I think we can be confident that getting rid of that argument will not break any code that wasn't already broken.

austinweisgrau: To be explicit, I re-include the `per_page` argument and issue a DeprecationWarning if it is used.

I think this is more concise and clear than adding `**kwargs`, which is more open-ended and I think would introduce more ambiguity.

parsons/action_network/action_network.py
```diff
@@ -153,6 +154,8 @@ def upsert_person(
                 an immediate success, with an empty JSON body, and send your request to the
                 background queue for eventual processing.
                 https://actionnetwork.org/docs/v2/#background-processing
+            identifiers:
+                list of strings to be used as globally unique identifiers
```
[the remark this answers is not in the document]
austinweisgrau: Good idea! Added with 55ecd18064c164371bfcc65c9e5a554d2d514f99

austinweisgrau:
> Hi Austin! We've reviewed this as part of our group pairing session. If you can address the comments other made, that would be great.
> 
> I'm also not sure how passing identifiers changes the response (if at all) and whether it's the sort of change that might need to be tested. Can you provide a bit more explanation of what this change does? Thank you!

Yeah definitely! After a more careful read of the docs, it looks like these cannot be used to match with existing contacts in ActionNetwork. Matching works exclusively on email and cell. 

When creating or updating a person, any provided identifiers will be included on the contact record in ActionNetwork.

These must be globally unique in ActionNetwork, so if an identifier is provided that is already associated with a different ActionNetwork contact, ... I will test this and report back.

austinweisgrau:
If the identifier provided is not globally unique in ActionNetwork, it will simply be ignored and not added to the object.

shaunagm (approved):
Looks good!

PR #868, opened 2023-08-14
add pypika to requirements

elyse-weiss:


elyse-weiss:
@willyraedy @shaunagm I seem to have an overactive formatter -- let me know if you want me to edit this to really just be pypika or if we're ok with this

shaunagm:
Yes, if you could edit this to be just the pypika change, that would be great.

PR #871, opened 2023-08-17
Enable use of "extension" argument to Redshift unload

austinweisgrau:
Extension is a helpful argument. This PR makes it possible to use the extension argument on Redshift.unload()

I have a workflow where I unload  SQL query to S3, and then make the resulting file available to download. It's great if that file already has a proper extension and I don't need to rename the file in S3 first. Currently all Redshift.unload() files have an extension of .gz if the file is compressed or simply ends with `000` if not.


jafayer (approved):
This change lgtm, can pass off to @shaunagm for a double-check!

shaunagm (approved):
This looks good to me too!

PR #873, opened 2023-08-17
Zoom Authentication + Polling API

IanRFerguson:
This PR adds two significant changes to the Zoom connector

## Server to Server OAuth
Zoom is deprecating JWT authentication on September 1st, which necessitates an update to the authentication strategy used by this connector. The `Zoom()` constructor now requires three values to connect to the API:

* `account_id`
* `client_id`
* `client_secret`

These values are passed to the Zoom OAuth callback URL and an access token is assigned to a class attribute.

**Please note** that these values replace the `api_key` and `api_secret` previously passed into the Zoom connector

### Setting Up a Server-to-Server OAuth App

Follow the steps below to generate the required env variables and scopes to authenticate your Zoom connector:

* Go to [Zoom's API Marketplace](https://developers.zoom.us/docs/api/)
* Click  "Create an App" -> "Develop" -> "Build an App" 
* Select **Server-to-Server OAuth**
* Save the client ID, client secret, and account ID somewhere safe that your connector can access
* Under "scopes" select the relevant endpoints that you want your connector to access (meetings, users, etc.)

## Zoom Poll API

Per request from @Jason94 and co, we've added functions to query Zoom polls from a given meeting. There are two relevant functions here, one to get a `Table` of all polls for a given meeting (just requires `meeting_id`) and one to get data about one specific poll (requires `meeting_id` AND `poll_id`)

---
  - https://app.asana.com/0/0/1205295200749537


IanRFerguson:
we're rockin 😎

parsons/zoom/zoom.py
```diff
@@ -43,17 +74,36 @@ def refresh_header_token(self):
...
+            Parsons Table of API responses
+        """
+
+        self.access_token = self.__generate_access_token()
+        headers = {
+            "Authorization": f"Bearer {self.access_token}",
+            "Content-type": "application/json",
+        }
+
+        self.client = APIConnector(ZOOM_URI, headers=headers)
```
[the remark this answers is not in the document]
Jason94: I think having it be an instance variable that's instantiated in the constructor is the usual pattern.

Jason94 (changes requested):
Thanks for working on this in such quick time, Ian! A few small comments.

parsons/zoom/zoom.py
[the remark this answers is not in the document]
IanRFerguson: that's cool, we were running into issues with requests_mock in the unit test so I moved it. embedding it in the constructor makes more sense to me too

parsons/zoom/zoom.py
```diff
@@ -23,15 +26,49 @@ class Zoom:
...
+        `Returns`:
+            String representation of access token
+        """
 
-    def refresh_header_token(self):
-        # Generate a token that is valid for 30 seconds and update header. Full documentation
-        # on JWT generation using Zoom API: https://marketplace.zoom.us/docs/guides/auth/jwt
+        temp_client = APIConnector(
+            uri=ZOOM_URI, auth=(self.client_id, self.__client_secret)
+        )
```
willyraedy, line 53: I assume the reason for the second client is that this one needs the `auth` param passeed in and the other one needs the bearer token?

test/test_zoom.py
```diff
@@ -4,14 +4,19 @@
...
-API_SECRET = "fake_api_secret"
+ACCOUNT_ID = "fakeAccountID"
+CLIENT_ID = "fakeClientID"
+CLIENT_SECRET = "fakeClientSecret"
+
 ZOOM_URI = "https://api.zoom.us/v2/"
+ZOOM_AUTH_CALLBACK = "https://zoom.us/oauth/token"
 
 
 class TestZoom(unittest.TestCase):
```
willyraedy, line 15: Let's add some unit tests for the new functions we added. Should be able to just copy the existing patterns for the new functions

willyraedy (approved):
Overall looks good! If we're under the gun could merge as is I think and could add tests in a fast follow PR. So will go ahead and approve but if there's time, we should add unit tests for the new functions.

parsons/zoom/zoom.py
```diff
@@ -225,3 +272,103 @@ def get_webinar_registrants(self, webinar_id):
...
+                Unique identifier for poll
+
+        `Returns`:
+            Parsons Table of all polling responses
+        """
+
+        endpoint = f"meetings/{meeting_id}/polls/{poll_id}"
+        tbl = self._get_request(endpoint=endpoint, data_key="polls")
+
+        if type(tbl) == dict:
```
Jason94, line 293: The return type is a parsons.Table, but here we're returning a dict. Should this just return `Table()` here?

Same comment applies for the other poll functions.

parsons/zoom/zoom.py
```diff
@@ -23,15 +26,49 @@ class Zoom:
             variable set.
     """
 
-    def __init__(self, api_key=None, api_secret=None):
+    def __init__(self, account_id=None, client_id=None, client_secret=None):
+        self.account_id = check_env.check("ZOOM_ACCOUNT_ID", account_id)
```
Jason94, line 30: Could you:
1. Rebase your branch on the major-release branch
2. Change the PR to target major-release
3. Add details in the PR description for all of the changes that would break existing code? In this case, I think it's just the changes to the env vars. This is really important so users know what they need to migrate when they upgrade.

Jason94: #3 is kind of mentioned, but it would be nice to list what the ENV variables were before and what the exact ENV variables need to be now.

parsons/zoom/zoom.py
```diff
@@ -43,17 +80,27 @@ def refresh_header_token(self):
...
+        `Returns`:
+            Parsons Table of API responses
+        """
 
-        self.refresh_header_token()
         r = self.client.get_request(endpoint, params=params, **kwargs)
         self.client.data_key = data_key
         data = self.client.data_parse(r)
 
-        if not params:
```
Jason94, line 54: I think this change to `params` could cause some problems. Currently the public facing functions that call `_get_request` have it default to `None`, not `{}`. That would still result in `params=None` getting passed in, and it won't be reset to an empty dict.

Could that cause issues?

Jason94 (changes requested):

IanRFerguson:
@Jason94 + @willyraedy I'm OOO tomorrow and Friday, will prioritize merging this on Monday with additional unit tests to stay ahead of the 9/1 deadline

IanRFerguson:
## Done
* Rebased and repointed PR
* Added walkthrough to PR description for creating an OAuth app in Zoom's API marketplace

## To Do
Need to finish unit tests for new endpoints

parsons/zoom/zoom.py
IanRFerguson: yep exactly, happy to workshop this if you think it's a little clunky

IanRFerguson:
@Jason94 I need some more time to work through the unit testing, but the authentication works as is so I think this is ready to merge with the caveat that the poll endpoints might not be 100% functional just yet (happy to dep those temporarily so they're not in the live version)

Jason94 (approved):

austinweisgrau:
Something went wrong with this PR, a lot of unrelated code got included, e.g. on the ActionNetwork connector 🤔

PR #874, opened 2023-08-20
Feature/more auth0 functions

talevy42:
Add functions for retrieving all users, blocking users, and retrieve connection ids


sharinetmc (changes requested):
Hi @talevy42! The code looks good, just need to add the docstrings!

shaunagm:
Looks like this is good to merge, sorry for the delay @talevy42 - I think I kept seeing "changes requested" in the status and assumed we were waiting on you instead of vice versa.

shaunagm (approved):
Sorry for the delay, approving these changes in Sharine's stead

PR #875, opened 2023-08-20
Wraedy/bigquery db connector

willyraedy:
Three main changes:
- Extends the BigQuery database connector to mimic more of the functionality of the Redshift connector.
- Extends the DB Connector class slightly.
- Moves BigQuery into the databases folder.

Main goal is to prepare the TMC network to migrate scripts from Redshift to BigQuery.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7
+FROM --platform=linux/amd64 python:3.7
```
willyraedy, line 1: need to remove this before merge

.circleci/config.yml
```diff
@@ -2,65 +2,22 @@
 #
 # Check https://circleci.com/docs/2.0/language-python/ for more details
 #
+
+version: 2
```
willyraedy, line 6: I think a lot of these changes are because I merged "main" into my branch. Can we merge main into the "major-release" branch?

And apologies if that's not standard practice

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7
```
willyraedy, line 1: Did the python support version get bumped to 3.8? If so, I'm confused why we didn't bump the image too

parsons/databases/bigquery/bigquery.py
```diff
@@ -0,0 +1,968 @@
+import pickle
```
willyraedy, line 1: blarg. I messed up the file renaming the first time.

here's a commit where I tried to (unsuccessfully) restore the file diff if it's helpful: https://github.com/move-coop/parsons/commit/fc2c9fc915eae9e70595a67da4e4ba3369ef8c7c

parsons/databases/bigquery/__init__.py
```diff
@@ -0,0 +1,3 @@
+from parsons.databases.bigquery.bigquery import BigQuery
```
willyraedy, line 1: Went back and forth on which organizational convention to follow. Decided to move to the databases folder but no strong opinion here if ppl want to leave in the google folder

parsons/google/google_cloud_storage.py
```diff
@@ -130,17 +134,21 @@ def delete_bucket(self, bucket_name, delete_blobs=False):
...
                 The name of the bucket
             max_results: int
+                TODO - What do we think about this?
                 TBD
-            prefix_filter: str
+            prefix: str
                 A prefix to filter files
+            pattern: str
+                TODO - Thinking this can offset the recursive behavior here
+                Filters files based on a pattern
```
willyraedy, line 151: TODO: cut this

parsons/databases/postgres/postgres.py
```diff
@@ -62,21 +62,25 @@ def copy(
...
+                        or ``truncate`` the table.
+                    strict_length: bool
+                        If the database table needs to be created, strict_length determines whether
+                        the created table's column sizes will be sized to exactly fit the current data,
+                        or if their size will be rounded up to account for future values being larger
+        <<<<<<< HEAD
+                        then the current dataset. Defaults to ``True``.
+        =======
+                        then the current dataset. Defaults to ``False``.
+        >>>>>>> major-release
```
willyraedy, line 83: woops

parsons/databases/bigquery/bigquery.py
```diff
@@ -0,0 +1,968 @@
...
+        return self._client
+
+    @contextmanager
+    def connection(self):  # TODO: Is this worth doing given the transaction thing?
+        """
+        Generate a BigQuery connection.
+        The connection is set up as a python "context manager", so it will be closed
+        automatically when the connection goes out of scope. Note that the BigQuery
+        API uses jobs to run database operations and, as such, simply has a no-op for
+        a "commit" function. If you would like to manage transactions, please use
```
willyraedy, line 177: Is this connection even worth having on the BigQuery connector because of the state issues?

parsons/databases/bigquery/bigquery.py
```diff
@@ -0,0 +1,968 @@
...
+        """
+
+        if not commit:
+            raise ValueError(
+                """
+                BigQuery implementation uses an API client which always auto-commits. If you wish to wrap
+                multiple queries in a transaction, use Mulit-Statement transactions within a single query
+                as outlined here: https://cloud.google.com/bigquery/docs/transactions
+            """
+            )
```
willyraedy, line 275: Kept this param to match with Redshift and then throw with a descriptive error. Could just omit the param altogether

willyraedy:
@shaunagm @Jason94 I'm gonna add some unit tests before I mark this ready to merge but if y'all want to give it a once over should be ready for that

sharinetmc:
@shaunagm Hiya! Just wondering if you had the time to take a look at this!

.circleci/config.yml
Jason94: Yeah, it's not super well defined at the moment. But it should be safe for anyone to merge main into major-release whenever they need to.

willyraedy: Sure

willyraedy: Ok should be g2g

parsons/databases/bigquery/__init__.py
Jason94: I prefer this way. I think it makes more sense to be in databases.

parsons/databases/bigquery/bigquery.py
```diff
@@ -0,0 +1,1123 @@
...
+                        header = list(row.keys())
+                        pickle.dump(header, temp_file)
+
+                    row_data = list(row.values())
+                    pickle.dump(row_data, temp_file)
+
+        if not got_results:
+            return None
+
+        ptable = petl.frompickle(temp_filename)
```
Jason94, line 1084: What is the point of doing the pickling 100,000 at a time if we're just going to load them all back into memory at the end anyway?

parsons/databases/bigquery/bigquery.py
```diff
@@ -0,0 +1,1123 @@
...
+        table_name: str,
+        if_exists: str = "fail",
+        max_errors: int = 0,
+        data_type: str = "csv",
+        csv_delimiter: str = ",",
+        ignoreheader: int = 1,
+        nullas: str = None,
+        allow_quoted_newlines: bool = True,
+        allow_jagged_rows: bool = True,
+        quote: str = None,
```
Jason94, line 327: This type annotation should be `Optional[str]`

parsons/databases/bigquery/bigquery.py
```diff
@@ -0,0 +1,1123 @@
...
+        schema: list = None,
```
Jason94, line 328: This type annotation should be `Optional[list]`, or better yet `Optional[List[dict]]`

parsons/databases/bigquery/bigquery.py
```diff
@@ -0,0 +1,1123 @@
...
+            ROLLBACK TRANSACTION;
+        END;
+        """
+        self.query(sql=queries_wrapped, parameters=parameters, return_values=False)
+
+    def copy_from_gcs(
+        self,
+        gcs_blob_uri: str,
+        table_name: str,
+        if_exists: str = "fail",
```
Jason94, line 319: Feel free to not do this (don't want to scope creep your PR too much) but it would be better to use an enum for this instead of a string.

parsons/databases/bigquery/bigquery.py
```diff
@@ -0,0 +1,1123 @@
...
+            destination=table_ref,
+            job_config=job_config,
+            **load_kwargs,
+        )
+
+        try:
+            load_job.result()
+        except exceptions.BadRequest as e:
+            if "one of the files is larger than the maximum allowed size." in str(e):
+                logger.info(
```
Jason94, line 423: Maybe these could be debug log level?

parsons/databases/bigquery/bigquery.py
```diff
@@ -0,0 +1,1123 @@
...
+        table_name: str,
+        if_exists: str = "fail",
+        max_errors: int = 0,
+        data_type: str = "csv",
+        csv_delimiter: str = ",",
+        ignoreheader: int = 1,
+        nullas: str = None,
+        allow_quoted_newlines: bool = True,
+        allow_jagged_rows: bool = True,
+        quote: str = None,
```
Jason94, line 468: Same comments about file type

parsons/databases/bigquery/bigquery.py
```diff
@@ -0,0 +1,1123 @@
...
+            distinct_check: boolean
+                Check if the primary key column is distinct. Raise error if not.
+            cleanup_temp_table: boolean
+                A temp table is dropped by default on cleanup. You can set to False for debugging.
+            from_s3: boolean
+                Instead of specifying a table_obj (set the first argument to None),
+                set this to True and include :func:`~parsons.databases.Redshift.copy_s3` arguments
+                to upsert a pre-existing s3 file into the target_table
+            \**copy_args: kwargs
+                See :func:`~parsons.databases.Redshift.copy` for options.
```
Jason94, line 806: Could we link to the BQ copy method instead?

parsons/databases/bigquery/bigquery.py
```diff
@@ -0,0 +1,1123 @@
...
+        Drop the table.
+        """
+
+        self.db.delete_table(self.table)
+
+    def truncate(self):
+        """
+        Truncate the table.
+        """
+        # BigQuery does not support truncate natively, so we will "load" an empty dataset
```
Jason94, line 1111: I don't know if this is out of date or what, but it looks like it's a simple command now: https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#truncate_table_statement

Jason94 (changes requested):
I left comments about the log level in a few places, but didn't touch it every place it could've been changed.

In general I think Parsons over-uses the info log level. I think that the info log level should be used if it's information that I might care about if I need to look back on what was happening in my script. For example, if I have a script that is uploading some new contacts, I probably want to have a log of the event ID's that were generated. But I probably don't care that it was compressing the table data as part of the GCS upload, or whatnot.

Something like that could be justified at the debug log level, maybe. The reason being that if something fails then it might be helpful to change the log level to DEBUG to see under the hood of what's happening. But even then, I'm not sure if giving a "Yep, this function is running." style of log message (that gives no information) is helpful.

parsons/databases/bigquery/bigquery.py
willyraedy: This was inherited from the previous version of BigQuery. Honestly didn't interrogate it much for improvements and I'm inclined to leave it for now

willyraedy:
Appreciate it Jason! I'm a little sick so @IanRFerguson is gonna take a run at making those changes. They mostly seem pretty minor so I suspect he'll be able get through those today or tomorrow

parsons/databases/bigquery/bigquery.py
IanRFerguson: @willyraedy didn't address this one in my review, let's connect on it

IanRFerguson:
@Jason94 I used some discretion to move many of the info-level logs to debug ... there's one outstanding comment that I'm talking through with Willy now but then we should be all set for another review

shaunagm (changes requested):
Overall this looks good to me - some minor nitpicks and flags. The main question I have is around the deprecation of the old connector. 

Also note that I have no personal experience with the BigQuery connector so my review is largely for style and readability, not for substance.

parsons/__init__.py
```diff
@@ -53,6 +53,7 @@
     ("parsons.databases.mysql.mysql", "MySQL"),
     ("parsons.databases.postgres.postgres", "Postgres"),
     ("parsons.databases.redshift.redshift", "Redshift"),
+    ("parsons.databases.bigquery.bigquery", "BigQuery"),
```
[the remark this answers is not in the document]
willyraedy: I guess I was thinking it would be part of a major release so breaking changes would be expected. But now that I think about that, I don't think there's any reason we can't import it from both places.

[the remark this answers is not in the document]
willyraedy: Moved the file back to the original location and modified the database init file to also import it from there. So both should work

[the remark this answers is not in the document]
shaunagm: Thanks for making this adjustment. Where possible, we want to give folks time to make the change, so in 3-6 we can circle back and remove this entirely.

The docs should signal this deprecation. I missed that this lacks a docs update in my first review, but this should update the docs so that the BigQuery docs are in the Databases folder instead of in the Google folder, both in the git repository and in the docs lefthand sidebar. It probably makes sense to leave a stub in the Google folder, though, that says something like "GoogleBigQuery has been moved to the Databases folder and is available under BigQuery, please change this in your code if you don't want it to break later".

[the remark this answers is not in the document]
willyraedy: I think the big query docs can still live in the same place since I moved the file back to its original place. Unless I'm missing something.

All we've basically done is take the original class and export it in the database package as well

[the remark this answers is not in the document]
shaunagm: I double-checked the docs and I was misremembering how they're structured. There's actually no subfolders at all, including no databases folder. Whoops. We can keep the databases .rst file as is - like how Redshift points to the entry under AWS, BigQuery's spot in the databases listing can keep pointing to Google.

That said, we do still want a note about the rename and the planned deprecation in the docs, so that anyone who checks the docs over the next 3-6 months will be able to proactively rename, and then when we finally do make the breaking change of removing GoogleBigQuery as an option, fewer people will be affected.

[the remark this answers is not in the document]
willyraedy: The more I think about it, I don't think renaming it from GoogleBigQuery to BigQuery which is functionally what it would be to users would be worth a deprecation process. So I just removed the second import path and will change our code.

Hoping to get this merged this week so we can onboard a contractor on Monday!

Jason94 (approved):

parsons/databases/bigquery/bigquery.py
coastlines: Bumping this because I'm not sure if it was ever resolved.

parsons/databases/bigquery/bigquery.py
coastlines: @Jason94 can you say more on why this would be beneficial? Purely out of curiosity, no rush.

parsons/databases/bigquery/bigquery.py
```diff
@@ -0,0 +1,1123 @@
...
+        """
+        Execute a query against the BigQuery database, with an existing connection.
+        Useful for batching queries together. Will return ``None`` if the query
+        returns zero rows.
+
+        `Args:`
+            sql: str
+                A valid SQL statement
+            connection: obj
+                A connection object obtained from ``redshift.connection()``
```
coastlines: Update to`bigquery.connection()`

parsons/databases/bigquery/bigquery.py
coastlines: Should this be referencing the BigQuery method?

coastlines: Sorry, just saw Jason's comment. It seems this has been resolved.

parsons/etl/tofrom.py
```diff
@@ -619,8 +619,35 @@ def to_postgres(
...
-    def to_petl(self):
+    def to_bigquery(
+        self, table_name: str, app_creds: str = None, project: str = None, **kwargs
+    ):
+        """
+        Write a table to BigQuery
 
+        `Args`:
+            table_name: str
+                TODO - Does this need project.dataset.table name?
```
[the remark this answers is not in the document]
coastlines: This confuses me too so I'm unresolving this conversation. It would be super helpful to have the documentation for this updated before merging.

parsons/etl/tofrom.py
```diff
@@ -898,6 +925,91 @@ def from_s3_csv(
...
+        `Args`:
+            sql: str
+                A valid SQL statement
+            app_creds: str
+                A credentials json string or a path to a json file. Not required
+                if ``GOOGLE_APPLICATION_CREDENTIALS`` env variable set.
+            project: str
+                The project which the client is acting on behalf of. If not passed
+                then will use the default inferred environment.
+            TODO - Should users be able to pass in kwargs here? For parameters?
```
coastlines, line 949: Yes but not urgent. Passing in kwargs is often very helpful for strategists.

coastlines (approved):
Impressive work y'all!

parsons/databases/bigquery/bigquery.py
willyraedy: good catch

parsons/databases/bigquery/bigquery.py
willyraedy: I think given the error checks we've put in, it's probably not worth deprecating since this was already on the class before

parsons/etl/tofrom.py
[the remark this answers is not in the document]
willyraedy: this was updated! looks like the Github diff is cached somehow

PR #877, opened 2023-08-26
fix: handle 429 error code

talevy42:
If you retrieve too many records, auth0 throws a 429 too many requests error, which gets handled naively. this pr throws an exception instead


sharinetmc (approved):
Looks good!

PR #879, opened 2023-08-29
Generate release notes

sharinetmc:
Added a header to the yml config file to see if it works now. The autogeneraion of release notes doesn't seem to be working, so seeing if this works!


sharinetmc:
@shaunagm Wondering if we can possibly pair on this during our meeting today, if there's time. I tried to create a config file to auto generate the release notes, but I'm not 100% why this isn't working since it doesn't throw any errors...

sharinetmc:
@shaunagm, just wondering if github got back to you about this? Thanks!

sharinetmc:
@KasiaHinkson, mind approving this? Hopefully this will be the fix that will allow us to autogenerate the release notes on Monday! If not, we can just do it manually. Github support apparently has a poor record of responding to support tickets :D

PR #884, opened 2023-09-05
Update example VAN apply_response JSON to reflect accurate keys

ydamit:
Current example shows JSON with `SurveyResponse` as the value for the key `action`, when that key should actually be `type` instead.


sharinetmc (approved):
Looks good!

PR #886, opened 2023-09-06
Zoom Polls

IanRFerguson:
This PR adds the following functions to the `Zoom` connector:
* `get_meeting_poll_metadata` - Returns metadata for a specific poll from a given meeting
* `get_meeting_all_polls_metadata` - Returns metadata for all polls from a given meeting
* `get_past_meeting_poll_metadata` - Returns metadata for a poll from a past meeting
* `get_webinar_poll_metadata` - Returns metadata for a specific poll from a given webinar
* `get_webinar_all_polls_metadata` - Returns metadata for all polls from a given webinar
* `get_past_webinar_poll_metadata` - Returns metadata for a poll from a past webinar
* `get_meeting_poll_results` - Returns poll results from a past meeting
* `get_webinar_poll_results` - Returns poll results from a webinar meeting

[Meeting API Reference](https://developers.zoom.us/docs/api/rest/reference/zoom-api/methods/#tag/Meetings)
[Webinar API Reference](https://developers.zoom.us/docs/api/rest/reference/zoom-api/methods/#tag/Webinars)

I'd like to request that @Jason94 / Indivisible test drive these functions before merging, as I've had a hard time finding polls in past TMC Zoom meetings

---
  - https://app.asana.com/0/0/1205295200749537


parsons/zoom/zoom.py
```diff
@@ -275,3 +275,103 @@ def get_webinar_registrants(self, webinar_id):
         tbl = self._get_request(f"webinars/{webinar_id}/registrants", "registrants")
         logger.info(f"Retrieved {tbl.num_rows} webinar registrants.")
         return tbl
+
+    def get_meeting_poll(self, meeting_id: str, poll_id: str) -> Table:
```
Jason94, line 279: The typing for the meeting ID's is off in the connector. Webinars do have string UUIDs, but meetings have integer IDs. Normally I'm in favor of adding type annotations, but not if it will conflict with other areas of the code. Maybe you could just remove the "str" type annotation for the meeting_ids in the code, and change the type to int in the docstrings? Then we could clean the rest of the file later?

parsons/zoom/zoom.py
```diff
@@ -275,3 +275,103 @@ def get_webinar_registrants(self, webinar_id):
...
+
+        `Args`:
+            meeting_id: str
+                Unique identifier for Zoom meeting
+
+        `Returns`:
+            Parsons Table of all polling responses
+        """
+
+        endpoint = f"meetings/{meeting_id}/polls"
```
Jason94, line 318: This looks light to me, but it doesn't seem to be working. I'm running this script (The account ID is a properly named ENV VAR so I don't need to reference it here):

``` python
from parsons.zoom import Zoom
import os

ZOOM_CLIENT_ID = os.environ["ZOOM_CREDENTIAL_API_USERNAME"]
ZOOM_CLIENT_SECRET = os.environ["ZOOM_CREDENTIAL_API_PASSWORD"]

MEETING_ID = 83783398800


def main():
    zoom = Zoom(client_id=ZOOM_CLIENT_ID, client_secret=ZOOM_CLIENT_SECRET)
    print(zoom.get_past_meeting(MEETING_ID))
    print(zoom.get_meeting_all_polls(MEETING_ID))


if __name__ == "__main__":
    main()
```

The first print, getting the meeting details, is working. So I know that the meeting does exist. But the second print, trying to get all of the polls, is failing. 

```
requests.exceptions.HTTPError: HTTP error occurred (404): Not Found, json: {'code': 3001, 'message': 'Meeting does not exist: 83783398800.'}
```

Jason94 (changes requested):

parsons/zoom/zoom.py
IanRFerguson: This is a great flag - seems like Zoom's REST API has a whole different endpoint for "past" meeting polls:
https://developers.zoom.us/docs/api/rest/reference/zoom-api/methods/#operation/listPastMeetingPolls

IanRFerguson: For the sake of consistency / verbosity I suppose I'll just write new functions for `past meetings` and `past webinars`?

IanRFerguson: my guess is that if you run `.get_meeting(MEETING_ID)` (instead of `get_past_meeting()`)  it'll throw an exception too, is that right?

parsons/zoom/zoom.py
```diff
@@ -275,3 +275,149 @@ def get_webinar_registrants(self, webinar_id):
...
+        logger.info(f"Retrieved {tbl.num_rows} polls for meeting ID {meeting_id}")
+
+        return tbl
+
+    def get_past_meeting_poll(self, meeting_id) -> Table:
+        """
+        List poll results of a past meeting.
+
+        `Args`:
+            meeting_id: str
```
Jason94, line 334: Here's a meeting_id type that's still string

IanRFerguson: good catch! I updated a few from existing functions too for consistency

parsons/zoom/zoom.py
IanRFerguson: ack, I should I have caught that that's my bad. Maybe I'll change the existing functions to `get_meeting_all_polls_metadata` and then the `get_meeting_poll` would return data from the reports endpoint? What do you think about that?

Trying to imagine when you'd want the metadata vs. the actual results, I reckon the latter is much more common but I could be wrong

Jason94: That sounds good to me. I don't know when you would want the metadata, but we might as well offer the option!

Jason94: That works! I was able to get poll results. I have two pieces of feedback on it now. 

1) Could we mention in the documentation that you need the report:read:admin scope in your Zoom authentication setup for any of the functions that get reports?

2) I think the return type _could_ be improved, maybe. Here is an example row:

```
{'name': 'John Doe', 'email': 'johndoe@gmail.com', 'question_details': [{'question': 'Do you think this could help you?', 'answer': 'Yes', 'polling_id': 'abcde', 'date_time': '2023-01-20 01:15:50'}, {'question': 'Do you have an upcoming event?', 'answer': 'No', 'polling_id': 'efgh', 'date_time': '2023-01-20 01:16:50'}], 'first_name': 'John', 'last_name': 'Doe'}
```

The question_details column would need some manual unpacking into some kind of _xf table or something.

@shaunagm do you have thoughts on this? Should the Parsons connector focus on shaping API data to make it easier to load into a database? Or should it focus on providing more direct access to the API data?

Jason94 (changes requested):

IanRFerguson:
I can only speak for myself, but it makes plenty of sense to make this connector as usable out-of-the-box as possible. I can't foresee any situation where community members would prefer the nested dictionaries rather than the unpacked values ... if y'all see it differently we can make the `__handle_nested_json` helper optional via keyword argument

parsons/zoom/zoom.py
```diff
@@ -275,3 +295,197 @@ def get_webinar_registrants(self, webinar_id):
...
+        endpoint = f"report/meetings/{meeting_id}/polls"
+        tbl = self._get_request(endpoint=endpoint, data_key="questions")
+
+        if type(tbl) == dict:
+            logger.debug(f"No poll data returned for meeting ID {meeting_id}")
+            return Table(tbl)
+
+        logger.info(f"Retrieved {tbl.num_rows} reults for meeting ID {meeting_id}")
+
+        return self.__handle_nested_json(table=tbl, column="question_details")
```
Jason94, line 473: This was better, but it was still putting JSON data into the table for me. I took the liberty of massaging it around until I got a schema that I liked for the data coming back. Here is the code I wound up with:

```
    def __process_poll_results(self, tbl: Table) -> Table:
        tbl.add_column("poll_taker_id", lambda _: str(uuid.uuid4()))
        tbl = tbl.unpack_nested_columns_as_rows(
            "question_details", key="poll_taker_id", expand_original=True
        )
        tbl.remove_column("poll_taker_id")
        tbl.remove_column("question_details")

        tbl = tbl.unpack_dict(
            "question_details_value", include_original=True, prepend=False
        )
        tbl.remove_column("question_details_value")

        return tbl


    def get_meeting_poll_results(self, meeting_id) -> Table:
        """
        Get a report of poll results for a past meeting

        Required scopes: `report:read:admin`
        """

        endpoint = f"report/meetings/{meeting_id}/polls"
        tbl = self._get_request(endpoint=endpoint, data_key="questions")

        if type(tbl) == dict:
            logger.debug(f"No poll data returned for meeting ID {meeting_id}")
            return Table(tbl)

        logger.info(f"Retrieved {tbl.num_rows} reults for meeting ID {meeting_id}")

        if tbl.num_rows > 0:
            tbl = self.__process_poll_results(tbl)

        return tbl

    def get_webinar_poll_results(self, webinar_id) -> Table:
        """
        Get a report of poll results for a past webinar

        Required scopes: `report:read:admin`
        """

        endpoint = f"report/webinars/{webinar_id}/polls"
        tbl = self._get_request(endpoint=endpoint, data_key="questions")

        if type(tbl) == dict:
            logger.debug(f"No poll data returned for webinar ID {webinar_id}")
            return Table(tbl)

        logger.info(f"Retrieved {tbl.num_rows} reults for webinar ID {webinar_id}")

        if tbl.num_rows > 0:
            tbl = self.__process_poll_results(tbl)

        return tbl
```

The data comes out pretty nice, with each answer getting a full separate row to make it easy to query in Python or in SQL

Jason94: Here is what the table in my script looked like after pushing to Redshift (I appended the three columns with the event information.)

image

Jason94 (changes requested):

Jason94 (approved):

PR #887, opened 2023-09-06
Action Builder Entity Record Insert Name Key Hotfix

ydamit:
Action Builder's API is not consistent about what key you need in your data for the name of the entity. For People (and other Entities configured like them) you need at least a `given_name` key. Other Entity types can be configured just to take a single name. These used to be insertable with the `name` key, but now seem to require `action_builder:name`. Rather than trying to force that prefix to be added, I'm just making sure all three keys pass the error handling in `insert_entity_record()`.


IanRFerguson (approved):
love it

PR #889, opened 2023-09-09
chore(actionnetwork): Edge case and cleaning up type checking

talevy42:
Handle edge case where acitonnetwork doesn't return a person_id, and use pythonic style for testing types


sharinetmc (approved):
@talevy42, thank you for adding the person_id check and for making the if patterns more inline with our parsons patterns! lgtm!

PR #890, opened 2023-09-14
Action Builder Remove Tag Method

ydamit:
Removing tags is not an option through the person signup helper endpoint, so this allows us to remove tags by providing certain combinations of arguments (the best being the tag's interact ID and the individual tagging's interact ID, as those are direct components of the DELETE endpoint).


IanRFerguson (approved):
Crispy clean! Looks great

PR #900, opened 2023-10-01
Nirs actionnetwork changes

NirTatcher:
Added some get functions for ActionNetwork module, I have more functions to add (create and update functions) after this batch is approved.
Please let me know if it looks ok :)

### DIFF
--- a/parsons/action_network/action_network.py
+++ b/parsons/action_network/action_network.py
@@ -433,7 +433,15 @@ class ActionNetwork(object):
     def get_events(self,limit=None, per_page=25, page=None, filter=None):
         """
              `Args:`
-                 self:
+            limit:
+                The number of entries to return. When None, returns all entries.
+            per_page
+                The number of entries per page to return. 25 maximum.
+            page
+                Which page of results to return
+            filter
+                The OData query for filtering results. E.g. "modified_date gt '2014-03-25'".
+                When None, no filter is applied.
              `Returns:`
                  A  JSON with all the events entries
          """
@@ -444,7 +452,6 @@ class ActionNetwork(object):
     def get_event(self,event_id):
         """
              `Args:`
-                 self:
                  event_id: the unique id of the event
              `Returns:`
                  A  JSON with event entry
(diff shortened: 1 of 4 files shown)


shaunagm:
Hi @NirTatcher sorry it's taking so long to get to this. I plan to review next week!

NirTatcher:
Hi @shaunagm, the PR didn't pass the linting tests, I fixed it (it should now pass the linting tests), updated the branch by the latest commit in main branch.
Thank you!

parsons/action_network/action_network.py
```diff
@@ -51,7 +51,7 @@ def _get_entry_list(self, object_name, limit=None, per_page=25, filter=None):
         while True:
             response = self._get_page(object_name, page, per_page, filter=filter)
             page = page + 1
-            response_list = response["_embedded"][f"osdi:{object_name}"]
+            response_list = response["_embedded"][list(response["_embedded"])[0]]
```
shaunagm, line 54: Can you explain what's going on here? I want to make sure this isn't a breaking change.

test/test_action_network/test_action_network.py
```diff
@@ -348,6 +348,46 @@ def setUp(self, m):
             },
```
shaunagm, line 348: You've added a number of methods with no test coverage. I realize that having test coverage is of relatively marginal value given that we have to mock the endpoints, and I don't want to create a ton of busywork for you, but I would love it if we could check that the call is what we expect, given the parameters input.

NirTatcher: Hi @shaunagm,
I reorganized the code and added tests for all functions added by me. 
Please let me know if it looks ok to you.

NirTatcher:
> Overall this looks great, had one quick question and a request about test coverage. Sorry for the delay in reviewing.

Hi @shaunagm,

I reorganized the code and added tests for every function. Everything should be working fine now please let me know how it looks like to you.

parsons/action_network/action_network.py
shaunagm: Thank you for the explanation! Super helpful. So it sounds like this will not break any existing code, because only `object_name`s like "donations" was working anyway, and your approach still works for them. This just expands the valid object_names. And I didn't realize this before, but `object_name` really only seems to be used internally, and is not exposed to the end-user, so that makes worrying about breaking changes way less important anyway. :)

test/test_action_network/test_action_network.py
shaunagm: These look great! Thank you for adding all these tests, Nir.

shaunagm (approved):
These look great. Thanks so much for adding all these methods, @NirTatcher!

shaunagm:
Merged! Thank you so much @NirTatcher

PR #902, opened 2023-10-07
Add Empower Connector

jburchard:
This connector adds support for the [Empower](https://empowerproject.us/empower-app/) relational app. As of my most recent testing, it includes support for all of the different json objects that are included in their API.


docs/empower.rst
```diff
@@ -0,0 +1,47 @@
...
+********
+Overview
+********
+
+The Empower class allows you to interact with the Empower API. Documentation for the Empower API can be found
+in their `GitHub <https://github.com/getempower/api-documentation/blob/master/README.md>`_ repo.
+
+.. note::
+    The Empower API only has a single endpoint to access all account data. As such, it has a very high overhead. This
+    connector employs caching in order to allow the user to specify the tables to extract without additional API calls.
```
sharinetmc, line 13: More of a personal question, coming from lack of experience, but wondering what caching is?

sharinetmc (commented):
@jburchard, everything here looks good to me, honestly! I just have a personal question! From what I understand, it seems that the connector basically grabs an object, which contains all the relevant data and then splits it up between the relevant tables that each method is associated with. Everything looks clean and makes sense. And all the documentation and mock tests and everything looks great.

@shaunagm, not sure if you might have any other comments or suggestions? Just marking this as a comment for now in case you might have anything to add before approval?

docs/empower.rst
```diff
@@ -0,0 +1,47 @@
...
+    You can disable caching as an argument when instantiating the class.
+
+==========
+Quickstart
+==========
+
+To instantiate the Empower class, you can either store your ``EMPOWER_API_KEY`` an environment
+variables or pass them in as arguments:
+
+.. code-block:: python
```
shaunagm, line 23: Might be nice to have an example of disabling caching in the quickstart

parsons/empower/empower.py
```diff
@@ -0,0 +1,256 @@
...
+    def _unix_convert(self, ts):
+        """
+        Converts UNIX timestamps to readable timestamps.
+        """
+
+        ts = datetime.utcfromtimestamp(int(ts) / 1000)
+        ts = ts.strftime("%Y-%m-%d %H:%M:%S UTC")
+        return ts
+
+    def _empty_obj(self, obj_name):
```
shaunagm, line 67: Is there is a reason this is a separate method rather than just using this code in the one place it's referenced? You could write it with just that first line - the return True and return False aren't necessary.

parsons/empower/empower.py
```diff
@@ -0,0 +1,256 @@
...
+    def get_profiles_active_ctas(self):
+        """
+        Get active ctas assigned to Empower profiles.
+
+        `Returns:`
+            Parsons Table
+                See :ref:`parsons-table` for output options.
+        """
+
+        tbl = Table(self.data["profiles"]).long_table("eid", "activeCtaIds")
```
shaunagm, line 103: `long_table` is a new one on me! Mind sharing what it is and why you're using it here? Thank you!

parsons/empower/empower.py
```diff
@@ -0,0 +1,256 @@
...
+        tbl = Table(self.data["ctaResults"])
+        tbl.convert_column("contactedMts", lambda x: self._unix_convert(x))
+        tbl = tbl.unpack_nested_columns_as_rows(
+            "answerIdsByPromptId", key="profileEid", expand_original=True
+        )
+        tbl.unpack_list("answerIdsByPromptId_value", replace=True)
+        col_list = [v for v in tbl.columns if v.find("value") != -1]
+        tbl.coalesce_columns("answer_id", col_list, remove_source_columns=True)
+        tbl.remove_column("uid")
+        tbl.remove_column("answers")  # Per docs, this is deprecated.
```
shaunagm, line 137: I believe that remove_column errors if the column name isn't found. Is it possible that this column name will stop getting returned? Or do you mean it's deprecated in some other way?

docs/empower.rst
jburchard: So, there is only one endpoint and it grabs a giant JSON blob that is a bit of a pain to parse through and convert to a tabular-ish format. I've broken up that JSON into multiple functions so that folks can grab the bits of data that they want.

However, that means that every single time you call it function, it is grabbing a lot of extraneous data. So, the caching just stores the blob and extracts from it rather than calling the Empower server again.

At the time, it seemed like a cute idea, but perhaps its unnecessary.

docs/empower.rst
jburchard: Can do.

shaunagm:
I added a couple of comments/questions @jburchard but nothing major. Should be pretty close to merging this.

parsons/empower/empower.py
jburchard: This column is still being returned by Empower, but their docs indicate that it isn't being used for anything any longer. So, I decided to not surface it for the user.

jburchard:
Thanks @shaunagm @sharinetmc for the feedback. I'll take another pass. There are also a few slight modifications (a few errors surfaced) when I started using a forked version of this, that I do want to include in the PR before it is merged.

shaunagm (changes requested):
Just formally requesting changes so it shows up with that status in the pull request list view :)

shaunagm:
Hi Justin, zero pressure - just wanted to see if you have plans to get to this soon?

KasiaHinkson:
TMC has been using Airbyte for our Empower sync and it's no longer working so we're looking for an alternative. This looks like good work, can TMC do anything to help get this over the line? 

Also, we were not doing any unnesting at all at the load step, so I'd like to include a function/option to just return the results as a 1 row table with all that nested json that we would write to GCS and then load into BQ, our dbt pipelines handle the rest.

shaunagm:
Merged in #1191 so closing this.

PR #913, opened 2023-10-31
BigQuery - Add row count function to connector

IanRFerguson:
Adds a `get_row_count` function to the BigQuery connector. The native version of this function in the BigQuery API client only works with tables, wondering if we should stick with that logic here and add another function for views


parsons/google/google_bigquery.py
```diff
@@ -1018,6 +1018,26 @@ def get_columns_list(self, schema: str, table_name: str) -> list:
 
         return [x for x in first_row.columns]
 
+    def get_row_count(self, schema: str, table_name: str) -> int:
```
shaunagm, line 1021: Kind of a bummer that you have to fetch the whole table in order to get the number of rows! But presumably there's not a better way to do this.

shaunagm:
I will defer to Willy etc on matters of substance, such as how to handle views. The only change I'd like to see is a test for the new method, otherwise lgtm.

parsons/google/google_bigquery.py
IanRFerguson: I believe this is more memory efficient than `SELECT COUNT(*) FROM {{ table }}` but if others have thoughts here we should update this accordingly

Jason94: I'd think that doing a `SELECT COUNT(*)` would be more efficient? If we select the whole table, then it needs to transfer all of the data from the database to the computation site over the network. If we do a `SELECT COUNT(*)` then it just needs to transfer the single integer.

IanRFerguson: You're right, I assumed that running `select count(*)` would scan the table in its entirety but it doesn't! I'll change this

IanRFerguson: Pushed! Still need to add a formal unit test before merging but this approach does seem more efficient

image

shaunagm:
Just want to flag that PRs against major-release don't have tests run. I'm going to try and fix that next week, but leaving a comment here to double-check that tests are run before this is merged.

PR #914, opened 2023-11-02
Google BigQuery - Clean Up Autodetect Logic

IanRFerguson:
The job config logic was written such that incoming `job_config` objects were set to `autodetect=True` ... we cleaned up the logic that's applied such that an explicitly passed-in schema will take precedence


willyraedy (approved):
LGTM!

PR #915, opened 2023-11-07
Fix incorrect documentation on ActionNetwork upsert_person argument

austinweisgrau:
Existing documentation is wrong. When email addresses are passed as described in a dict, they are ignored by the method.

I added a type check that will raise an error if a dict is passed, in order to make it more clear to the user that this is an invalid argument.

I also updated the documentation to clarify that a list of dicts is acceptable but not a plain dict. Also, the keys described for the dict are wrong - those are also updated.

See https://actionnetwork.org/docs/v2/people/


shaunagm:
Thanks for flagging this! I applied the 'breaking change' label just as a reminder to whoever writes the release note to mention this one specifically - it's better to fail with an error than fail silently, but it nonetheless might cause some previously running scripts to break.

I'll merge once the tests pass.

shaunagm (approved):
lgtm

PR #916, opened 2023-11-09
Enable passing arbitrary additional fields to NGPVAN people match API

austinweisgrau:
The existing arguments for using the NGPVAN people match API are quite limited. There are many more fields that would be useful to have access to. Currently it is possible to access those by using the `find_person_json()` method, but this requires reproducing a lot of the behavior around structuring the JSON response. Allowing kwargs to be passed through the stack would simplify the usage here.

If I want to add an employer to a contact record, the current use looks like:
```
parsons.VAN(...).upsert_person_json({
    "first_name": "Austin",
    "last_name": "Weisgrau",
    "emails": [{
        "email": "austinweisgrau@gmail.com"
    }],
    "employer": "Working Families Party"
})
```

After this change, it looks like this:
```
parsons.VAN(...).upsert_person(first_name='Austin', last_name='Weisgrau', email='austinweisgrau@gmail.com', employer='Working Families Party')
```

The more fields are used, the more the added simplicity of this new implementation.

### DIFF
--- a/parsons/ngpvan/people.py
+++ b/parsons/ngpvan/people.py
@@ -6,7 +6,6 @@ logger = logging.getLogger(__name__)
 
 class People(object):
     def __init__(self, van_connection):
-
         self.connection = van_connection
 
     def find_person(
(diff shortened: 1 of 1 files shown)


KasiaHinkson (approved):
LGTM, love it!

PR #917, opened 2023-11-09
Enhancement: Action Network Connector: Added unpack_statistics param in get_messages method

justicehaze:
This PR introduces an enhancement to the get_messages method in the action_network.py file. The update includes an optional parameter unpack_statistics which, when set to True, unpacks the statistics dictionary into the returned Parsons Table. This feature provides users with more granular data access, allowing them to directly interact with the statistics related to each message.

The unpack_statistics parameter defaults to False, ensuring backward compatibility for users who do not require this level of detail. When unpack_statistics is True, the method uses the unpack_dict function on the statistics field of the Parsons Table.

This enhancement improves the flexibility and usefulness of the get_messages method, providing users with more control over the data they retrieve.

Please review and provide any feedback.

### DIFF
--- a/parsons/action_network/action_network.py
+++ b/parsons/action_network/action_network.py
@@ -583,7 +583,7 @@ class ActionNetwork(object):
         return self.api.get_request(f"lists/{list_id}")
 
     # Messages
-    def get_messages(self, limit=None, per_page=25, page=None, filter=None):
+    def get_messages(self, limit=None, per_page=25, page=None, filter=None, unpack_statistics=True):
         """
         `Args:`
             limit:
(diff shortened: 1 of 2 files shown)


justicehaze:
Both these accounts are me but looks like I submitted it under a different account. Confusing. Let me know if there are any questions. I decided to make the default behavior to be not unpacking for backwards compatibility. I know it's a new connector but I thought if someone had used this API endpoint themselves and is switching to this, I wanted the native behavior to be the same.

parsons/action_network/action_network.py
```diff
@@ -595,14 +597,21 @@ def get_messages(self, limit=None, per_page=25, page=None, filter=None):
...
+            A Parsons Table with all the messages related entries
         """
+        # Fetch messages
         if page:
             return self._get_page("messages", page, per_page, filter)
-        return self._get_entry_list("messages", limit, per_page, filter)
+        tbl = self._get_entry_list("messages", limit, per_page, filter)
+        # Unpack statistics
+        if unpack_statistics:
+            tbl.unpack_dict("statistics", prepend=False, include_original=True)
```
shaunagm, line 613: I assume that based on the existing structure of the messages data Action Network returns, there's no chance of the unpacked statistics duplicating a column name? If there _is_ a chance the unpacked names would be duplicates, that might cause an error or just overwrite a column that the user might need. If you can reassure me I'll go ahead and merge this. :)

shaunagm (approved):

parsons/action_network/action_network.py
justicehaze: That's correct. Here is the api documentation about what is returned. The statistics section has no overlapping field names with the other columns. 

[Action Network API Docs => Messages Endpoint](https://actionnetwork.org/docs/v2/messages)

shaunagm: Great, thanks! I'll go ahead and merge.

PR #918, opened 2023-11-09
Generalize the oauth_api_connector to accept authorization kwargs

austinweisgrau:
The oauth_api_connector can simplify authorization with APIs that use OAuth2, but requires some modifications to be general enough to work with different implementations of OAuth2.

This PR enables the use of authorization kwargs that are passed to authorization endpoints when requesting a token. The updated connector is then implemented in the Zoom connector as a test / demonstration.

Someone who has Zoom credentials should check if this update works.

Prior to this PR, the only connector that used the oauth_api_connector is the controlshift connector. That one should be unaffected, because the new argument is optional and doesn't change behavior if it is not used.


austinweisgrau:
Rebasing onto main and adding a commit that updates the catalist match connector to also use the oauth_api_connector

austinweisgrau:
This is verified to work with the Catalist Match connector

PR #919, opened 2023-11-12
Enhancement: Parsons Table: added rename_columns for multiple cols

mkrausse-ggtx:
This pull request introduces a new method, rename_columns, to the ETL process. This method allows for the renaming of multiple columns at once in a Parsons Table.

The rename_columns method takes a dictionary as an argument (column_map), where the keys represent the current column names and the values represent the new column names. The method checks if the old column name exists and if the new column name does not already exist in the table. If these conditions are met, the method uses the underlying petl.rename function to perform the renaming operation.

The method returns the updated Parsons Table and also updates the self.table attribute of the class instance.

Unit tests have been added to ensure the rename_columns method works as expected and handles edge cases appropriately. These tests cover scenarios such as renaming columns with a valid column map, renaming only some columns, and attempting to rename a column that does not exist or to a name that already exists.

This enhancement provides a more efficient way to rename multiple columns, improving the flexibility and usability of the ETL process.

### DIFF
--- a/parsons/etl/etl.py
+++ b/parsons/etl/etl.py
@@ -80,7 +80,13 @@ class ETL(object):
 
         `Args:`
             column_map: dict
-                A dictionary of columns and new names
+                A dictionary of columns and new names.
+                The key is the old name and the value is the new name.
+
+                Example dictionary:
+                {'old_name': 'new_name',
+                'old_name2': 'new_name2'}
+
         `Returns:`
             `Parsons Table` and also updates self
         """
--- a/parsons/etl/etl.py
+++ b/parsons/etl/etl.py
@@ -73,7 +73,7 @@ class ETL(object):
         self.table = petl.rename(self.table, column_name, new_column_name)
 
         return self
-    
+
     def rename_columns(self, column_map):
         """
         Rename multiple columns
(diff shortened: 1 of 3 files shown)


parsons/etl/etl.py
```diff
@@ -74,6 +74,29 @@ def rename_column(self, column_name, new_column_name):
 
         return self
 
+    def rename_columns(self, column_map):
+        """
+        Rename multiple columns
+
+        `Args:`
+            column_map: dict
+                A dictionary of columns and new names
```
shaunagm, line 83: Can you specify here that the old names are keys and the new names are values? ie `{'old_name': 'new_name'}`?

shaunagm (changes requested):
Overall this looks good, just a couple small docs changes. One is an adjustment to your docstring, the other is to this file: https://github.com/move-coop/parsons/blob/main/docs/table.rst#parsons-table-transformations

You can see there's a table that lists available Parsons Table transformations. If you can add a row underneath rename_column for your new rename_columns method, that should allow people to find your new method more easily.

parsons/etl/etl.py
mkrausse-ggtx: Sure things. Resolved.

shaunagm:
Why did you close this? Do you need help updating your PR?

mkrausse-ggtx:
> Why did you close this? Do you need help updating your PR?

Looks like when I tried to rename the branch to the naming convention I'm trying to keep, it closed it. Reopened.

PR #922, opened 2023-11-14
Add QuickBooks connector and related updates

mkrausse-ggtx:
This pull request adds a QuickBooks connector to the Parsons library, which allows users to pull time tracking data from QuickBooks Time. 

This includes methods for getting data out of the software like users, groups, jobcodes, timesheets, schedules.

### DIFF
--- a/parsons/quickbooks/quickbooks.py
+++ b/parsons/quickbooks/quickbooks.py
@@ -2,6 +2,7 @@ from parsons import Table
 from parsons.utilities import check_env
 from parsons.utilities.api_connector import APIConnector
 import logging
+import collections
 
 # Set up logger
 logger = logging.getLogger(__name__)
@@ -33,37 +34,49 @@ class QuickBooks:
 
     def qb_get_request(self, end_point: str, querystring=None):
         """
-        This functions handles the pagination of the request
+        This function handles the pagination of the request
         """
 
+        # If no querystring is provided, initialize it as an empty dictionary
         if querystring is None:
             querystring = {}
-        output_list = []
+        output_list = []  # This list will hold the results
 
-        page = 1
-        more = True
+        page = 1  # Start from the first page
+        more = True  # This flag indicates if there are more pages to fetch
         while more:
+            
             # After every 10 pages, log the progress
             if page % 10 == 0:
                 logger.info(f"Retrieved {len(output_list)} records from {end_point} endpoint.")
                 logger.info(f"Currently on page {page}.")
-
+                
+            # Add the current page to the querystring
             querystring = {**querystring, **{"page": page}}
 
+            # Send the GET request
             response = self.client.get_request(end_point, params=querystring)
 
-            temp_list = [response['results'][end_point][record] for record in response["results"][end_point]]
+            # Extract the key of the results
+            endpoint_key = list(response['results'].keys())[0]
+            # Extract the records from the results
+            temp_list = [response['results'][endpoint_key][record] for record in response["results"][endpoint_key]]
 
+            # If the response indicates there are more pages, update the flag and increment the page number
             if "more" in response:
                 more = response["more"]
                 page += 1
 
+            # Add the records from the current page to the output list
             output_list += temp_list
 
+            # If there are no more pages, break the loop
             if not "more" in response or response["more"] == False:
                 break
 
+        # Log the total number of records retrieved
         logger.info(f"Retrieved {len(output_list)} records from {end_point} endpoint.")
+        # Return the results as a Table
         return Table(output_list)
 
     def get_groups(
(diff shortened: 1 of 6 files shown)


mkrausse-ggtx:
I am running all the linting commands and it's still not passing the linting tests. Not sure what to do at this point.

mkrausse-ggtx:
Success! I had pasted in the API docs from QB Time and they had all these problems with line length and trailing whitespaces that took forever to fix. But there is a vscode extension for flake8 that highlights all the issues. That helped as opposed to reading that long list of lines in the flake8 output.

parsons/quickbooks/quickbooks.py
```diff
@@ -0,0 +1,734 @@
...
+logger = logging.getLogger(__name__)
+
+QB_URI = "https://rest.tsheets.com/api/v1/"
+
+
+class QuickBooks:
+    """
+    Instantiate the QuickBooks class.
+
+    TODO: Add information about the difference between accounting vs time
```
shaunagm, line 16: Not a huge deal, but would be nice not to publish the docs with a 'todo' in them. You can always link to a guide distinguishing the two if you don't want to have to write up the distinction yourself.

parsons/quickbooks/quickbooks.py
```diff
@@ -0,0 +1,734 @@
...
+
+    `Args:`
+        token: str
+            A valid QuickBooks Auth Token. Not required if ``QB_AUTH_TOKEN`` env
+            variable set. Find instructions to create yours here:
+            https://developer.intuit.com/app/developer/qbo/docs/develop/authentication-and-authorization/oauth-2.0
```
shaunagm, line 22: You can use the restructured text to markup docstrings to (ie https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links). Here's an example of someone doing this in a Parsons connector: https://github.com/move-coop/parsons/blob/main/parsons/aws/aws_async.py#L30

parsons/quickbooks/quickbooks.py
```diff
@@ -0,0 +1,734 @@
...
+
+            Quickbooks API Documentation: https://tsheetsteam.github.io/api_docs/#introduction
```
shaunagm, line 24: Again, if you can reformat so this is just a link with text "Quickbooks API Documentation" that'd be great - much easier for the docs reader

parsons/quickbooks/quickbooks.py
```diff
@@ -0,0 +1,734 @@
...
+                Only jobcodes with an id set to one of these values will be returned.
+                If omitted, all jobcodes matching other specified filters are returned.
+
+            parent_ids: Int
+                Default is -1 (meaning all jobcodes will be returned regardless of parent_id).
+                Comma separated list of one or more jobcode parent_ids you'd like to filter on.
+                Only jobcodes with a parent_id set to one of these values will be returned.
+                Additionally you can use 0 to get only the top-level jobcodes.
+
+                Then get the id of any results with has_children=yes
```
shaunagm, line 187: The formatting of this looks a little weird. Have you built the docs and checked that it looks okay?

test/test_quickbooks/test_quickbooks.py
```diff
@@ -0,0 +1,275 @@
...
+        # Act
+        result = self.qb.get_schedule_events()
+
+        # Assert
+        self.assertIsInstance(result, Table)
+
+    @requests_mock.Mocker()
+    def test_get_jobcodes_with_params(self, mock_request):
+        # Arrange
+        mock_request.get(requests_mock.ANY, json=mock_jobcodes_data)
```
shaunagm, line 113: I'm a little confused why this data is here and not in `test_quickbooks_data.py`. It's not so much data as to make the tests unreadable, but it is inconsistent.

test/test_quickbooks/test_quickbooks.py
```diff
@@ -0,0 +1,275 @@
...
+    @requests_mock.Mocker()
+    def test_get_groups(self, mock_request):
+        # Arrange
+        mock_request.get(requests_mock.ANY, json=mock_groups_data)
+
+        # Act
+        result = self.qb.get_groups()
+
+        # Assert
+        self.assertIsInstance(result, Table)
```
shaunagm, line 53: These tests only test whether the result is a Table, but you return an empty table for most of the methods. So you might also want to test the number of rows > 0 for all of these, and/or test the specific data being returned for each test.

shaunagm (changes requested):
Hi Matthew! So sorry this ended up taking so long to review, we didn't end up communicating well about who was reviewing what. Overall looks great, just a few minor changes need to merge.

parsons/quickbooks/quickbookstime.py
```diff
@@ -0,0 +1,732 @@
...
+        supplemental_data=None,
+        per_page=None,
+        page=None,
+    ):
+        """
+            This function allows you to call the /groups endpoint of the QuickBooksTime Time API.
+            All Args are optional.
+
+            `Args:`
+                ids: Int
```
shaunagm, line 100: It seems like there might be an indentation issue here? Can you fix it so all args are at the same level of indentation?

parsons/quickbooks/quickbookstime.py
```diff
@@ -0,0 +1,732 @@
...
+            "group_ids": group_ids,
+            "not_group_ids": not_group_ids,
+            "payroll_ids": payroll_ids,
+            "active": active,
+            "first_name": first_name,
+            "last_name": last_name,
+            "modified_before": modified_before,
+            "modified_since": modified_since,
+            "supplemental_data": supplemental_data,
+            "per_page": per_page,
```
shaunagm, line 460: This seems like the querystring dict is missing the "page" parameter listed in the docstring?

You could maybe do something programmatic with these functions, where you accept **kwargs and then pass them along to `qb_get_request` (if there are any functions where there's a param that doesn't go into the querystring, you could pop it in that function). That would decrease the size of the functions and also prevent accidentally missing parameters.

shaunagm (commented):
Overall looks good, just a minor indentation issue and then a possible bug with those parameters - whether or not you want to implement the design of switching out explicit paremters for **kwargs is up to you.

parsons/quickbooks/quickbookstime.py
shaunagm: Sorry @matthewkrausse I'm still confused. The default is 0 but is there any way to override the default? The page parameter getting passed in here seems to be overwritten in `qb_get_request`.

matthewkrausse (approved):
> Sorry @matthewkrausse I'm still confused. The default is 0 but is there any way to override the default? The page parameter getting passed in here seems to be overwritten in `qb_get_request`. 

I see what you are saying. I went ahead and fixed the issue in the qb_get_request method and did a little bit of refactoring to make it a bit more clear.

parsons/quickbooks/quickbookstime.py
matthewkrausse: I see what you are saying. I made a change in the `qb_get_request` method to address that.

shaunagm:
Looks good to me. Ideally the functions will get refactored to be more concise someday, but that's something that shouldn't be a breaking change when it happens so it's not a blocker on this PR.

PR #924, opened 2023-11-15
Add http response to update_mailer

sjwmoveon:
Without returning the response, or at least the status code, it's impossible to check for errors.


shaunagm (approved):
This looks straightforward, and not a breaking change because the method previously returned nothing.

Issue #925, opened 2023-11-22
[Bug] Error on apply_canvas_result() in VAN for lack of phone even though phone is provided


retacg:
The `apply_canvas_result()` method fails saying phone is needed even when phone is provided. It looks like there are many phone contacttypes that might require phone ever since VAN made the update to require phone numbers for phone contact types. There was a previous pr that only addressed the situation for a couple contacttypes - https://github.com/move-coop/parsons/issues/525
I am trying to load canvas results for contacttypeid = 19 (Auto Dial) and it is errorring saying that phone is now required. 

I queried and looks like there are potentially a lot more contacttypes that we should include in the section of requiring phone but that are not documented in VAN's documentation as requiring phone...? 
<img width="949" alt="Screenshot 2023-11-22 at 10 19 38 AM" src="https://github.com/move-coop/parsons/assets/62267774/6b0c4acf-da74-46d8-979a-d4bccef996b0">

## Detailed Description

When I try and load canvas results with the contacttypeid of 19, i get the error that 

`INFO The error was: HTTP error occurred (400): Bad Request, json: {'errors': [{'code': 'INVALID_PARAMETER', 'text': "'phone.phoneNumber' is required.", 'properties': ['canvassContext']}]}`

## To Reproduce

        van.apply_canvass_result(
            row['vanid'],
            result_code_id=row['resultid'], 
            date_canvassed=str(row['datecanvassed']),
            contact_type_id = 19,
            input_type_id=11,
            phone=str(row['phone'])
        )

## Your Environment

* Version of Parsons used (if you know): current
* Environment name and version (e.g. Chrome 39, node.js 5.4): parsons docker image?
* Operating System and version (desktop or mobile):

## Priority
medium/ high this is for a member that would like to be able to use a script using this parsons function!

justicehaze:
I'll go ahead and get this fixed. It looks like if the contact type is phone or sms, it sends the phone. If it is not phone or sms, it does not add phone to the data being sent. I'll handle this to include the phone no matter which contact type but give an error if phone is not being provided. This way if they add more contact types it wont not send the phone and have data loss.

mkrausse-ggtx:
I have opened a PR that fixes this bug. #927

shaunagm:
Closed by #927

PR #927, opened 2023-11-24
Fix bug in NGPVAN apply_canvass_result method

mkrausse-ggtx:
This pull request fixes a bug in the `apply_canvass_result` function. The problem was that the phone was only being sent when the contact_type_id was either 1 (phone) or 37 (SMS text) even though it was required for other contact types. Now it raises an error if the contact type requires phone. But, if the phone is provided, it sends it regardless of the contact type id.


mkrausse-ggtx:
@shaunagm this addresses https://github.com/move-coop/parsons/issues/925. I think this is the best way to handle this bug.

shaunagm (approved):
had a conversation about this at a parsons party, lgtm

PR #928, opened 2023-11-24
Add utility method to format phone numbers

mkrausse-ggtx:
This pull request adds a new utility method, `format_phone_number`, which formats a phone number in E.164 format. The method takes a phone number as input and an optional country code, and returns the formatted phone number. It removes non-numeric characters and leading zeros, adds the country code prefix if necessary, and formats the number in E.164 format. The pull request also includes unit tests for the `format_phone_number` method to ensure its correctness and handle different scenarios.

### DIFF
--- a/parsons/utilities/format_phone_number.py
+++ b/parsons/utilities/format_phone_number.py
@@ -3,15 +3,19 @@ import re
 
 def format_phone_number(phone_number, country_code="1"):
     """
-    Formats a phone number in E.164 format.
+    Formats a phone number in E.164 format, which is the international standard for phone numbers.
+    Example: Converts "555-555-5555" -> "+15555555555"
 
-    Args:
-        phone_number (str): The phone number to be formatted.
-        country_code (str): The country code to be used as a prefix.
+    `Args:`
+        phone_number (str):
+            The phone number to be formatted.
+        country_code (str):
+            The country code to be used as a prefix.
         Defaults to "1" (United States).
 
-    Returns:
-        str: The formatted phone number in E.164 format.
+    `Returns:`
+        str:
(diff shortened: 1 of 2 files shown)


austinweisgrau:
IMO a better default behavior for a format phone number method would be to return None for an invalid phone number rather than raise an error.

mkrausse-ggtx:
That makes sense to me. I can update that.

mkrausse-ggtx:
I'm happy to do either one. Do folks think my approach to not adding another dependency makes sense? I'm also willing to write a function that uses the phonenumbers package.

austinweisgrau:
Yeah I think your approach is good! The dependency seems unnecessary.

A few other validations worth incorporating potentially - valid US phones can't have an area code that starts with 0 or 1, and the first digit after the area code also can't be 0 or 1. I think 999 is also an invalid area code.

FWIW the phonenumbers package doesn't invalidate those patterns so there's no advantage there either.

parsons/utilities/format_phone_number.py
```diff
@@ -0,0 +1,33 @@
+import re
+
+
+def format_phone_number(phone_number, country_code="1"):
+    """
+    Formats a phone number in E.164 format.
```
Jason94, line 6: Could you add some documentation showing what it'll do? That would be helpful to someone who doesn't know what the e.164 format is, and might have no idea what this does.

Even one line like:

`Example: Converts "(123) 123-1212" to "+01231231212"`

would be very helpful!

matthewkrausse: Sounds good. Went ahead and made that change.

PR #930, opened 2023-11-28
Fix gcs hidden error

sharinetmc:
Changes the object that gets pulled when an error gets raised.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,3 @@
 [flake8]
-max-line-length = 200
+max-line-length = 100
 
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,3 @@
 [flake8]
-max-line-length = 100
+max-line-length = 200
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
         language_version: python3
         args: [
           '--extend-ignore=E203,W503',
-          '--max-line-length=200'
+          '--max-line-length=100'
         ]
   - repo: https://github.com/psf/black
     rev: 22.3.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
         language_version: python3
         args: [
           '--extend-ignore=E203,W503',
-          '--max-line-length=100'
+          '--max-line-length=200'
         ]
   - repo: https://github.com/psf/black
     rev: 22.3.0
--- a/parsons/google/google_cloud_storage.py
+++ b/parsons/google/google_cloud_storage.py
@@ -460,11 +460,13 @@ class GoogleCloudStorage(object):
                     logger.debug("Operation still running...")
 
                 else:
+                    logger.info(operation.error.code)
                     if int(operation.error.code) not in StatusCode["OK"].value:
                         raise Exception(
                             f"""{blob_storage} to GCS Transfer Job {create_result.name} failed with error: {operation.error.message}
                             """
                         )
+                    logger.info(operation.response)
                     if operation.response:
                         logger.info(f"TransferJob: {create_result.name} succeeded.")
                         return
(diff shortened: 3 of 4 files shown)


.flake8
```diff
@@ -1,3 +1,3 @@
 [flake8]
-max-line-length = 100
+max-line-length = 200
```
Jason94, line 2: Intentional? If so, this should probably be a separate PR because it's pretty opinionated.

.pre-commit-config.yaml
```diff
@@ -6,7 +6,7 @@ repos:
         language_version: python3
         args: [
           '--extend-ignore=E203,W503',
-          '--max-line-length=100'
+          '--max-line-length=200'
```
Jason94, line 9: Same comment about line length

parsons/google/google_cloud_storage.py
```diff
@@ -439,12 +437,11 @@ def copy_bucket_to_gcs(
 
         # Create the transfer job
         create_result = client.create_transfer_job(create_transfer_job_request)
-        logger.info(f"Created TransferJob: {create_result.name}")
 
         polling = True
         wait_time = 0
         wait_between_attempts_in_sec = 10
-        max_wait_in_sec = 60 * 10  # Ten Minutes
+        # max_wait_in_sec = 60 * 10  # Ten Minutes
```
Jason94, line 444: If this variable isn't being used in the final version of the code, we should probably delete the line.

parsons/google/google_cloud_storage.py
```diff
@@ -453,27 +450,33 @@ def copy_bucket_to_gcs(
...
-                        return
+                    return
 
             else:
                 logger.info("Waiting to kickoff operation...")
                 get_transfer_job_request = storage_transfer.GetTransferJobRequest(
                     {"job_name": create_result.name, "project_id": self.project}
                 )
                 get_result = client.get_transfer_job(request=get_transfer_job_request)
+                logger.info(f"get_result: {get_result}")
```
Jason94, line 479: Could this be a debug log, instead of an info? Is this something that would be helpful to someone looking through the log of a higher-level ETL script after it had an error, or something?

parsons/google/google_cloud_storage.py
```diff
@@ -453,27 +450,33 @@ def copy_bucket_to_gcs(
...
+                    )
+                    error_output = operation_metadata.error_breakdowns
+                    if len(error_output) != 0:
                         raise Exception(
-                            f"""{blob_storage} to GCS Transfer Job {create_result.name} failed with error: {operation.error.message}
+                            f"""{blob_storage} to GCS Transfer Job {create_result.name} failed with error: {error_output}
                             """
                         )
-                    if operation.response:
+                    else:
```
Jason94, line 469: Strictly speaking, you don't need the `else` here because the `if` clause above it will terminate execution of the method. Sometimes people like to leave the `else` in those cases to make it clearer, which is fine. (I usually do that myself, actually.)

I do think it's strange that the `return` is _not_ in the else, though. The execution in the non-error path will be to skip the if, run one line in the else, and then immediately run the return. It seems like either the log and the return should be in the else, or it shouldn't be there at all.

Jason94 (changes requested):

sharinetmc:
@Jason94, thank you for the feedback!

.flake8
Jason94: Looks like this is still present. Did we get approval from the Parsons community on doubling the maximum line length to 200?

I noticed you had an issue with a long string. Python allows you to separate strings inside parenthesis with a line break to avoid long strings breaking line length limitations. Like this:

```python
foo(f"pretend this is {a} a really long string {b}", 1, 2)

# Can be written:
foo(
  f"pretend this is {a}"
  " a really long"
  f" string {b}",
  1,
  2
)
```

Jason94 (changes requested):

Jason94 (approved):
Looks good! Thanks, Sharine :D

Issue #929, opened 2023-11-28
Documentation docstring in BigQuery client query method is wrong


austinweisgrau:
Documentation docstring in BigQuery client query() method is copied from the parsons.Redshift.query() method docstring and is not fully modified to reflect differences. 

The example is based on parsons.Redshift.query() syntax and does not demonstrate how to parametrize a query using the BigQuery client.

https://github.com/move-coop/parsons/blob/90c0f43b2a0207991f58d22dcad4c08aef81223e/parsons/google/google_bigquery.py#L199

austinweisgrau:
This is fairly high priority for us as it is blocking our migration to BigQuery

austinweisgrau:
Parameterization of queries is actually not currently possible with the current implementation

PR #931, opened 2023-11-28
Enable parameterization of bigquery queries

austinweisgrau:
Addresses #929

Documentation on BigQuery query parameterization was wrong, but also with the existing implementation of the query method, parameterization was not possible.

Updates are made to enable parameterization of queries using the bigquery client, and documentation is updated to reflect correct usage.

BigQuery python query parameterization requires the use of `client.query()` rather than `cursor.query()`

See documentation here: https://cloud.google.com/bigquery/docs/parameterized-queries


austinweisgrau:
This is fairly high priority for us as it is blocking our migration to BigQuery

austinweisgrau:
#938 significantly refactors the bigquery connector query method, although not in a way that fixes the underlying issue here. This PR will likely need to be redone on top of the state of the code in that PR.

PR #933, opened 2023-11-28
Enable Table.to_bigquery()

austinweisgrau:
This creates parity with Table.to_redshift()

### DIFF
--- a/parsons/etl/tofrom.py
+++ b/parsons/etl/tofrom.py
@@ -3,6 +3,7 @@ import json
 import io
 import gzip
 from parsons.utilities import files, zip_archive
+from typing import Optional
 
 
 class ToFrom(object):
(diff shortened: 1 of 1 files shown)


austinweisgrau:
This is high priority for us, blocks our migration to BigQuery

Jason94 (approved):
This looks good to me. Approving.

Just curious, is there a reason you can't just use BigQuery.copy() instead of needing to have a Table.to_bigquery() method?

(I'm on record as saying we should deprecate and eventually get rid of the `Table.to_my_database()` methods, so I'm curious what the use case is.)

austinweisgrau:
I use a lot of `Table.to_redshift()` and so it simplifies migration. I appreciate the method, it's simple and ergonomic. In general I think we should have parity between features in Redshift and BigQuery connectors.

Using the copy() method is fine as well in general, I have no fundamental preference at an abstract level beyond parity.

Jason94:
> I use a lot of `Table.to_redshift()` and so it simplifies migration. I appreciate the method, it's simple and ergonomic. In general I think we should have parity between features in Redshift and BigQuery connectors.
> 
> Using the copy() method is fine as well in general, I have no fundamental preference at an abstract level beyond parity.

I agree there should be parity. Personally I'd rather delete the Redshift method than add a BigQuery one, but that's just my own battle :D Given that we can't do that anytime soon, definitely good to add the to_bigquery one in the meantime.

austinweisgrau:
This feature seems to be included already in #938

austinweisgrau:
Closing this out since it is included in #938

PR #936, opened 2023-11-29
GoogleCloudStorage - Add GCS Destination Path Param

IanRFerguson:
The `s3 -> gcs` component of the `copy_bucket_to_gcs()` function allows for the user to supply a destination path argument, such that blobs are stored in a nested sub-bucket.

This PR replicates that logic in `gcs -> gcs` transfers.


KasiaHinkson (approved):
LGTM

PR #937, opened 2023-11-29
GoogleCloudStorage - Handle zip / gzip files flexibly

IanRFerguson:
The current `gcs.unzip_blob()` implementation only handles `gzip` files, we're updating to handle other forms of compression


parsons/google/google_cloud_storage.py
```diff
@@ -568,4 +606,26 @@ def unzip_blob(
...
 
-        return self.format_uri(bucket=bucket_name, name=decompressed_blob_name)
+    def __zip_decompress_and_write_to_gcs(self, **kwargs):
+        compressed_filepath = kwargs.pop("compressed_filepath")
+        decompressed_blob_name = kwargs.pop("decompressed_blob_name")
+        decompressed_blob_in_archive = decompressed_blob_name.split("/")[-1]
+        bucket_name = kwargs.pop("bucket_name")
+        new_file_extension = kwargs.pop("new_file_extension")
+
+        staging_file = "./staging"
```
IanRFerguson, line 616: I'd like to do something different here

IanRFerguson: As I understand the `zipfile` process here we need to (1) unzip the archive (2) open the file to be read, which has the same filename as the decompressed blob and (3) copy the contents into a file to be put in storage

If we could combine steps 2 and 3 and simply stream the contents of the file to GCS that would obviously be preferable, but this current iteration is functional at the very least

parsons/google/google_bigquery.py
```diff
@@ -371,6 +373,8 @@ def copy_from_gcs(
                 on the BigQuery client. The function will create its own if not provided. Note
                 if there are any conflicts between the job_config and other parameters, the
                 job_config values are preferred.
+            force_unzip_blobs: bool
+                If True, target blobs will be unzipped before being loaded to BigQuery
```
Jason94, line 377: Needs a period for consistency.

Jason94 (changes requested):

parsons/google/google_cloud_storage.py
IanRFerguson: love this suggestion!

parsons/google/google_cloud_storage.py
```diff
@@ -543,29 +548,92 @@ def unzip_blob(
...
+            with path_.open(decompressed_blob_in_archive) as f_in:
+                # Write bytes to temporary file
+                with tempfile.TemporaryFile("wb+") as f_out:
+                    shutil.copyfileobj(f_in, f_out)
+                    logger.debug(
+                        f"Uploading uncompressed file to GCS: {decompressed_blob_name}"
+                    )
+                    bucket = self.get_bucket(bucket_name=bucket_name)
+                    blob = storage.Blob(name=decompressed_blob_name, bucket=bucket)
+                    blob.upload_from_file(file_obj=f_out, rewind=True)
```
Jason94, line 639: What happens if you just pass `f_in` here? Does it work?

Jason94 (approved):
Looks good!

IanRFerguson:
Ok one issue to address @Jason94 ... for "really big files" (e.g., the New York Catalist voterfile) I'm getting a timeout trying to stream the data into GCS (`requests.exceptions.ConnectionError: ('Connection aborted.', timeout('The write operation timed out'))`)

Do you think it's worth trying to handle this (either by checking the size of the file or with a try / catch), or should we just stick with the local file -> blob method that exists in the current function?

IanRFerguson:
> Ok one issue to address @Jason94 ... for "really big files" (e.g., the New York Catalist voterfile) I'm getting a timeout trying to stream the data into GCS (`requests.exceptions.ConnectionError: ('Connection aborted.', timeout('The write operation timed out'))`)
> 
> Do you think it's worth trying to handle this (either by checking the size of the file or with a try / catch), or should we just stick with the local file -> blob method that exists in the current function?

There IS a `timeout` param in the Google API, maybe I'll just mess with that

parsons/google/google_cloud_storage.py
```diff
@@ -547,29 +551,82 @@ def unzip_blob(
...
+            compressed_filepath=compressed_filepath,
+            decompressed_filepath=decompressed_filepath,
+            decompressed_blob_name=decompressed_blob_name,
+            bucket_name=bucket_name,
+            new_file_extension=new_file_extension,
+        )
+
+        return self.format_uri(bucket=bucket_name, name=decompressed_blob_name)
+
+    def __gzip_decompress_and_write_to_gcs(self, **kwargs):
```
willyraedy, line 594: Dig the two different, private functions for the two different cases here

Wonder if a single, public function `decompress and write_to_gcs` would be useful for folks and would make the above code simple b/c you could just call that public api and pass in the compression type

parsons/google/google_bigquery.py
```diff
@@ -371,6 +374,12 @@ def copy_from_gcs(
                 on the BigQuery client. The function will create its own if not provided. Note
                 if there are any conflicts between the job_config and other parameters, the
                 job_config values are preferred.
+            force_unzip_blobs: bool
+                If True, target blobs will be unzipped before being loaded to BigQuery.
+            compression_type: str
```
willyraedy, line 379: This is only relevant for the decompression and not the job config? I wonder if we make that clear by calling it "force_unzip_compression_type" and "force_unzip_new_file_extension

willyraedy (commented):
Couple small style things but love how generalized this is!

parsons/google/google_cloud_storage.py
IanRFerguson: That's essentially the `unzip_blob()` function, no? You pass in `zip` or `gzip` and the private function is accessed under the hood

willyraedy: yup you're totally right. this is good to go

willyraedy (approved):
🤠

parsons/google/google_cloud_storage.py
IanRFerguson: 🎸 🎸 🎸

PR #938, opened 2023-12-01
Merge major-release into main

shaunagm:
I just merged main into major-release, per @Jason94's advice, so this should be clean.


shaunagm:
@IanRFerguson @willyraedy it looks like there were some uncaught issues with your BigQuery PRs, I think because we're not testing the same (or at all???) when things get merged to major-release. Can you take a look and let me know whether this is a problem with the original PR (and if so, your timeline for fixing it) or whether this is actually still a versioning/release issue. Thanks!

austinweisgrau:
This PR seems to include problematic code in the ActionNetwork connector from #861 that was reverted with #876

austinweisgrau:
The major-release branch is exclusively behind the main branch, there is no work in the major-release branch that needs to be merged into main. e.g. try rebasing this branch on top of main - there are no new commits.

austinweisgrau:
Nevermind I had the wrong branch checked out lol

austinweisgrau:
#945 fixes the issue with #876 getting lost

austinweisgrau:
#946 should fix the test failure issue related to the storage_transfer import

.gitignore
```diff
@@ -110,6 +110,7 @@ venv.bak/
 # scratch
 scratch*
 old!_*
+test.ipynb
```
IanRFerguson, line 113: [nit] we should probably just git ignore all `ipynb` files but this is fine

PR #940, opened 2023-12-03
Nirs actionnetwork changes

NirTatcher:
Hi @shaunagm, Round 2 of functions (POST, PUT and even DELTE just for taggings) to create or update any resource that is available to be updated.
Please let me know if there is any feedback!
Thanks


parsons/action_network/action_network.py
```diff
@@ -169,6 +169,32 @@ def get_person_attendance(self, person_id, attendance_id):
...
 
+    def create_attendance(self, event_id, payload):
+        """
+        `Args:`
+            event_id: the unique id of the event
+            payload: the payload for creating the event attendance
+        `Returns:`
+            A JSON response after creating the event attendance
+        """
+        return self.api.post_request(f"events/{event_id}/attendances", payload)
```
shaunagm, line 180: I'm noticing now that a lot of these methods just say "the payload". I imagine that the payloads for a lot of these items are very detailed and we don't want to put all that in the docstrings, but perhaps we could link to the payload details? For example this one could link to https://actionnetwork.org/docs/v2/attendances

parsons/action_network/action_network.py
```diff
@@ -1139,6 +1412,34 @@ def get_person_signature(self, person_id, signature_id):
...
+            petition_id:
+               The unique id of the petition
+            data:
+               The data for creating the signature
+        `Returns:`
+            A JSON with the created signature entry
+        """
+        return self.api.post_request(f"petitions/{petition_id}/signatures", data)
+
+    def update_signature(self, petition_id, signature_id, data):
```
shaunagm, line 1427: Why is this one "data" and not "payload"?

test/test_action_network/test_action_network.py
```diff
@@ -3874,6 +3985,39 @@ def test_get_message(self, m):
             self.fake_message,
         )
 
+    @requests_mock.Mocker()
+    def test_create_message(self, m):
+        payload = {
+            "subject": "Stop doing the bad thing",
```
shaunagm, line 3991: 😂 I love test data

shaunagm:
@NirTatcher this looks great overall, my only concern is about having (a) consistency and (b) enough details in the parameters to the various methods.

shaunagm (changes requested):

parsons/action_network/action_network.py
NirTatcher: Hi @shaunagm! 
I hope that is ok I've added a "Documentation Reference" for each function for example:
        """
        `Args:`
            tag_id:
                The unique id of the tag
            payload:
                The payload for creating the tagging
                {
                    "_links" : {
                        "osdi:person" : { "href" : "https://actionnetwork.org/api/v2/people/id" }
                    }
                }
        `Returns:`
            A JSON response after creating the tagging
        **`Documentation Reference`:
            https://actionnetwork.org/docs/v2/taggings**
        """
Please let me know if this is acceptable or if you prefer another way to attach a reference from the documentation!

parsons/action_network/action_network.py
NirTatcher: You're right, didn't see it! I changed it to payload :)

parsons/action_network/action_network.py
shaunagm: This is perfect, thanks!

shaunagm (approved):
Thanks for making the changes, these look great! Apologies for the delay - hope you enjoyed the holidays :)

Issue #942, opened 2023-12-04
[Feature/Addition] parsons.databases.database.DatabaseCreateStatement.detect_data_type() should detect more data types than str, float, int


austinweisgrau:
`parsons.databases.database.DatabaseCreateStatement.detect_data_type()` is only able to currently detect data types str, float, and int. As a result, downstream methods (e.g. `parsons.databases.redshift.rs_create_table.RedshiftCreateTable.data_type()`) that generate DDL statements to create database tables are only able to infer data types for database tables for these 3 types.

It should be possible to infer types for more data types, including datetimes, dates, and booleans.

## Context

When using `parsons.Redshift.copy()` to upload a parsons Table into a Redshift database, parsons should be able to infer data types for most column types and pass those into Redshift accurately. However, the current implementation only accomplishes this for string, float, and integer types. All other types are coerced into strings. There should be enough information available in a parsons table to detect and accurately represent other types, including datetimes, dates, and booleans.

## Possible Implementation

postgres, mysql, and redshift database connectors all use the underlying `detect_data_type()` method, so whatever change is made here will need to be compatible with all 3 downstream connectors and SQL generators.

Jason94:
Yes, I can confirm this is really annoying for boolean columns. I have to use `bool_col = 'True'` as a workaround.

austinweisgrau:
Figured out that parsing BOOLs works if `parsons.databases.database.constants.DO_PARSE_BOOLS` is set to True. There doesn't seem to be any documentation about why this defaults to False or anything.

austinweisgrau:
The feature was added with 766cfaedc5652af8c39fde890067a99b6fa58518 and it seems to have been turned off by default for "reverse compatibility," which doesn't really make sense to me 🤔

Jason94:
Oh dear, they couldn't have even made that an argument to the function or something? This would definitely be a breaking change, but a really good one... Maybe we can target flipping the default for the next major release? (I know Shauna is cutting one right now.)

austinweisgrau:
Ya I made the PR #943 on the major-release branch

Jason94:
Are we ready to close this issue, @austinweisgrau ?

austinweisgrau:
I think it would still be nice to have support for loading datetime and date type objects to databases where possible!

dannyboy15:
Has any work been done on this? If not I can jump in. I think I had already started working on some code that did this.

[...]

PR #943, opened 2023-12-04
Parse Boolean types by default

austinweisgrau:
Commit 766cfaedc5652af8c39fde890067a99b6fa58518 created a feature for parsing boolean types but turned it off by default. This commit turns that feature on by default and adds a comment about how to turn it off and what that does.

When this feature is turned on, parsons will detect boolean column types when copying a parsons Table to a database and create a boolean column in the database. The default behavior without this change is to coerce booleans to be strings when loading to database.

Partially addresses #942


austinweisgrau:
This will change default behavior for loading booleans from parsons Table to a database, so implementations based on the prior default behavior will likely break. e.g.

```
tbl = Table([{"bool_col": True}])
Redshift.copy(tbl, "tablename")
Redshift().query("select * from tablename where bool_col = 'True'")
```
will no longer work, instead the more normal code will be needed:
```
Redshift().query("select * from tablename where bool_col ")
```

IanRFerguson:
> Ok so digging deeper, it looks like the current implementation will accept any of the following case-insensitive python values as True: `[True, "true", "t", 1, "1", "yes"]` and any of the following as False: `[False, "false", "f", 0, "0", "no"]`
> 
> I think this is bad and the only values we should translate into boolean are python booleans, `[True, False]`
> 
> If someone wants a set of strings or integers to be coerced into booleans when they're loading a parsons Table to a database, they should do that type conversion explicitly themself. It seems sketchy to have strings like "t" or "yEs" getting coerced into booleans by default.
> 
> These changes are made with [1cc6bd1](https://github.com/move-coop/parsons/commit/1cc6bd1e44d855cc9853f954666ffad189db9ff4)
> 
> What do you think?

I agree philosophically, we shouldn't be coercing data types implicitly so this is a valuable change. Going to take one more look at the diff but overall looks good

parsons/databases/database/constants.py
```diff
@@ -153,11 +153,9 @@
 
 VARCHAR = "varchar"
 FLOAT = "float"
-
-DO_PARSE_BOOLS = False
 BOOL = "bool"
-TRUE_VALS = ("TRUE", "T", "YES", "Y", "1", 1)
```
IanRFerguson, line 159: wowza

parsons/databases/database/database.py
```diff
@@ -161,64 +138,45 @@ def detect_data_type(self, value, cmp_type=None):
         # Stop if the compare type is already a varchar
         # varchar is the highest data type.
         if cmp_type == self.VARCHAR:
```
IanRFerguson, line 140: So much cleaner

parsons/databases/database/constants.py
austinweisgrau: Oh yeah I should delete those lines lol, they're not used anymore

Jason94 (approved):
This looks really great. Thanks @austinweisgrau for getting this over the finish line.

PR #947, opened 2023-12-06
Enable parameterization of bigquery queries

austinweisgrau:
Addresses https://github.com/move-coop/parsons/issues/929

Documentation on BigQuery query parameterization was wrong, but also with the existing implementation of the query method, parameterization was not possible.

Updates are made to enable parameterization of queries using the bigquery client, and documentation is updated to reflect correct usage.

BigQuery python query parameterization requires the use of client.query() rather than a more generic database connection & cursor.

See documentation here: https://cloud.google.com/bigquery/docs/parameterized-queries

This PR replaces #931 to make these updates on top of the major-release branch version of the bigquery client, rather than main.

TODO:
- [ ] Tests on google_bigquery module will need to be fixed, but they were already broken on the major-release branch so require some work there first

### DIFF
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -176,38 +176,6 @@ class GoogleBigQuery(DatabaseConnector):
 
         return self._client
 
-    @contextmanager
-    def connection(self):
-        """
-        Generate a BigQuery connection.
-        The connection is set up as a python "context manager", so it will be closed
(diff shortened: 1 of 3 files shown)


austinweisgrau:
My bad! This was not an issue, I was wrong. I closed #931 and I'm closing this out too.

The change made here works more closely with the BigQuery preferred python interaction, but I'm perfectly happy with and support the choices made in favor of database interoperability.

PR #948, opened 2023-12-07
Resolve GCP Test Failures For Major Release

IanRFerguson:


IanRFerguson:
I'm going to keep updating the docs but all tests are passing locally now @willyraedy

austinweisgrau:
Wait why are these tests passing without the fix to the google storage_transfer import? I can't get this to work locally...

IanRFerguson:
> Wait why are these tests passing without the fix to the google storage_transfer import? I can't get this to work locally...

uh oh ... you're seeing the same `ImportError` we were dealing with before?

austinweisgrau:
locally yeah but something weird is going on - maybe just for me? - github actions clearly doesnt have an issue with anything here

austinweisgrau:
Ah it works if you do `pip install -r requirements.txt` (which the tests do) but not if you do `pip install parsons[google]`

Something in the dependency versions...

austinweisgrau:
Got it - the `google-cloud-storage-transfer` wasn't included as a dependency in the extras for parsons[google] in the setup.py file.

Fixed that here: https://github.com/move-coop/parsons/pull/946

austinweisgrau:
The tests passed in github actions but failed for me locally due to a misconfiguration of test environment, see [here](https://github.com/move-coop/parsons/pull/951)

PR #949, opened 2023-12-07
Add Tests to major-release Branch

IanRFerguson:
Adds `major-release` branch to GitHub workflow to run tests / `black` / `flake8` checks


willyraedy (approved):
Not sure if there's a way to test this but I think this is what we want.

We may have to also merge main into the major-release branch as well after we merge this PR

austinweisgrau:
If we make the target of this PR the major release branch instead of main, it should run the tests if it works

PR #951, opened 2023-12-07
Install dependencies for tests based on package install

austinweisgrau:
Parsons dependencies can be installed in two different ways.
`pip install -r requirements.txt` is equivalent to pip install parsons
UNLESS the environment variable `PARSONS_LIMITED_DEPENDENCIES` is set,
in which case the dependencies are installed based on the explicit
configuration in `setup.py`.

There can be drift between what is defined in requirements.txt and
what is defined in `setup.py`.

These changes ensure that both sets of dependencies are used to run
tests, ensuring alignment between test environments and actual usage.

An example is tests passing on [this PR](https://github.com/move-coop/parsons/pull/948) even though the same tests would fail when using `PARSONS_LIMITED_DEPENDENCIES=1 pip install parsons[all]` (as explained and fixed in [this PR](https://github.com/move-coop/parsons/pull/946))

### DIFF
--- a/.github/workflows/test-linux-windows.yml
+++ b/.github/workflows/test-linux-windows.yml
@@ -14,6 +14,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.8', '3.9', '3.10']
+        limited-dependencies: ['','TRUE']
         os: [ubuntu-latest]   # add in windows-latest to add windows testing
         include:
         - os: ubuntu-latest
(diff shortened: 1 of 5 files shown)


shaunagm:
My only concern about this is that it doubles the total number of tests, and might push us out of the free tier of Github actions. But we can always revisit if we run out of memory/minutes/compute/whatever-it-is-they-limit

PR #953, opened 2023-12-08
Kasiahinkson/extend list blobs

KasiaHinkson:


parsons/google/google_cloud_storage.py
```diff
@@ -136,7 +136,14 @@ def delete_bucket(self, bucket_name, delete_blobs=False):
...
         logger.info(f"{bucket_name} bucket deleted.")
 
-    def list_blobs(self, bucket_name, max_results=None, prefix=None, match_glob=None):
+    def list_blobs(
+        self,
+        bucket_name,
+        max_results=None,
+        prefix=None,
+        match_glob=None,
+        include_file_details=False,
```
Jason94, line 145: Could you update the documentation of the return type to account for `include_file_details`?

parsons/google/google_cloud_storage.py
```diff
@@ -158,6 +165,10 @@ def list_blobs(self, bucket_name, max_results=None, prefix=None, match_glob=None
         blobs = self.client.list_blobs(
             bucket_name, max_results=max_results, prefix=prefix, match_glob=match_glob
         )
+
+        if include_file_details:
+            return [b for b in blobs]
+
         lst = [b.name for b in blobs]
         logger.info(f"Found {len(lst)} in {bucket_name} bucket.")
```
Jason94, line 173: We should probably still do the logger statement in case we pass in `include_file_details`, right? I think it's weird that behavior would change just because we want to return more info about the blobs.

Jason94 (changes requested):

parsons/google/google_cloud_storage.py
IanRFerguson: Definitely, good catch

parsons/google/google_cloud_storage.py
IanRFerguson: Added to the docstring - I linked to Google's reference docs here but can't remember if that's bad practice or not (happy to remove if so)

Jason94 (approved):
Looks good!

PR #956, opened 2023-12-11
Fix BigQuery type map for datetime objects

austinweisgrau:
Source types ultimately come from `petl.typeset`, which calls `type(v).__name__`. This call does not include source module, but only the type name itself. e.g. `date` and not `datetime.date`


austinweisgrau:
Closing this in favor of consolidated fixes in #959

PR #958, opened 2023-12-12
Fix nullable column type parsing by BigQuery connector

austinweisgrau:
If a Parsons Table column has values like `[None, None, True, False]`,
the BigQuery connector will infer that the appropriate type for this
column is NoneType, which it will translate into a STRING type.

This change ensures that types returned by petl.typecheck() will
choose the first available type that isnt `NoneType` if that is
available.

Adds a test that checks if the connector is able to successfully parse a column type when nulls are included. First commit here will allow test to run and fail, second commit will make the fix and demonstrate the test succeeds.


austinweisgrau:
Closing this in favor of consolidated fixes in #959

PR #960, opened 2023-12-19
Adding TMC canalespy query helpers

sharinetmc:
Adds two sql querying helper functions (one gets sql from a file and the other dedups a given table using the partition order by logic).

### DIFF
--- a/parsons/databases/table.py
+++ b/parsons/databases/table.py
@@ -1,4 +1,5 @@
 import logging
+import datetime
 
 logger = logging.getLogger(__name__)
 
@@ -164,3 +165,70 @@ class BaseTable:
 
         self.db.query(f"TRUNCATE TABLE {self.table}")
         logger.info(f"{self.table} truncated.")
+
+    def dedup_table(
+        self,
+        order_by_column_name=None,
+        order_by_direction=None,
+        cascade=False,
+        columns_to_ignore=None,
+    ):
+        """
+        Description:
+            This function re-creates a deduped version of a table by grabbing
+            all columns and inserting those into a partition statement for
+            row_number().
+        Args:
+            order_by_column_name: str
+                Column name of specific column that you would like to dedup using order by
+            order_by_direction: str
+                Order by direction, if you would like to dedup by ordering by a specific column,
+                this is the direction of the order by
+                example: 'asc'
+            cascade: bool
+                Set to True if you want any dependent views to be dropped -
+                queries will fail if there are dependent views and this is set to False.
+           columns_to_ignore: list
+                List any columns that should be ignored in the dedup
+        """
+        current_timestamp = datetime.datetime.now().strftime("%Y%m%d%H%M%S")
+        run_cascade = "CASCADE" if cascade else ""
+        order_by_column_name = (
+            "random()" if order_by_column_name is None else order_by_column_name
+        )
+        if order_by_direction is None and order_by_column_name is not None:
+            raise Exception("order_by_direction argument is blank")
+
+        columns_list = self.columns
+
+        # remove order_by columns
+        columns_list.remove(
+            order_by_column_name
+        ) if order_by_column_name is not None else None
+
+        # remove ignore columns
+        if columns_to_ignore is not None:
+            for column in columns_to_ignore:
+                columns_list.remove(column)
+
+        partition = ", ".join(columns_list)
+
+        dedup_query = f"""
+            alter table {self.table}
+            rename to {self.table}temp_{current_timestamp};
+            create table {self.table} as
+            select * from
+            (select *
+            , row_number() over (partition by {partition}
+            order by {order_by_column_name} {order_by_direction}) as dup
+            from {self.table}_temp_{current_timestamp})
+            where dup=1;
+            alter table {self.table}
+            drop column dup;
+            drop table {self.table}temp_{current_timestamp} {run_cascade};
+        """
+
+        self.db.query(dedup_query)
+        logger.info(f"Finished deduping {self.table}...")
+
+        return None
--- a/parsons/databases/table.py
+++ b/parsons/databases/table.py
@@ -215,7 +215,7 @@ class BaseTable:
 
         dedup_query = f"""
             alter table {self.table}
-            rename to {self.table}temp_{current_timestamp};
+            rename to {self.table}_temp_{current_timestamp};
             create table {self.table} as
             select * from
             (select *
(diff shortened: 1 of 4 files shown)


parsons/databases/table.py
```diff
@@ -164,3 +165,70 @@ def truncate(self):
...
+        order_by_direction=None,
+        cascade=False,
+        columns_to_ignore=None,
+    ):
+        """
+        Description:
+            This function re-creates a deduped version of a table by grabbing
+            all columns and inserting those into a partition statement for
+            row_number().
+        Args:
```
Jason94, line 181: All of these args should be documented as Optional. (That's something we haven't always done a great job of following, especially in the older code.)

Parsons doesn't follow the Google styleguide exactly, but I like to generally follow these: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

parsons/databases/table.py
```diff
@@ -164,3 +165,70 @@ def truncate(self):
...
+        ) if order_by_column_name is not None else None
+
+        # remove ignore columns
+        if columns_to_ignore is not None:
+            for column in columns_to_ignore:
+                columns_list.remove(column)
+
+        partition = ", ".join(columns_list)
+
+        dedup_query = f"""
```
Jason94, line 216: In addition to having issues with views, this is going to reset permissions of the table anytime you dedup it, which I think is really counterintuitive. 

What about instead of renaming the table, we do the select query with the filtered row_number calculation into a temp table, wipe the original table, and then copy the data from the temp table back into the original?

parsons/databases/table.py
```diff
@@ -164,3 +165,70 @@ def truncate(self):
...
+            create table {self.table} as
+            select * from
+            (select *
+            , row_number() over (partition by {partition}
+            order by {order_by_column_name} {order_by_direction}) as dup
+            from {self.table}_temp_{current_timestamp})
+            where dup=1;
+            alter table {self.table}
+            drop column dup;
+            drop table {self.table}temp_{current_timestamp} {run_cascade};
```
Jason94, line 228: I think you forgot the `_` before the temp. (Although making my above suggestion would remove this line anyway.)

parsons/utilities/sql_helpers.py
```diff
@@ -12,3 +12,17 @@ def redact_credentials(sql):
     sql_censored = re.sub(pattern, "CREDENTIALS REDACTED", sql, flags=re.IGNORECASE)
 
     return sql_censored
+
+
+def get_sql_from_file(sql_file):
```
Jason94, line 17: Is this function really worth having? It's just wrapping f.read. I think a code-smell here is that this could be reading _any_ file, from a jpg or a blog post, not just a SQL file. That would mean this function isn't adding value beyond just using the `File.reader()` method.

test/test_databases/test_mysql.py
```diff
@@ -108,6 +108,11 @@ def test_truncate(self):
         self.tbl.truncate()
         self.assertEqual(self.tbl.num_rows, 0)
 
+    def test_dedup_table(self):
+
+        self.tbl.dedup_table(order_by_column_name="user_name")
```
Jason94, line 113: Does this test actually pass? IT looks like the table defined in the set up doesn't have any duplicates in it, so it should still have 3 rows? Maybe this test isn't running because the MySQL database isn't set up?

Jason94 (changes requested):

test/test_utilities.py
```diff
@@ -150,6 +150,20 @@ def test_redact_credentials():
     assert sql_helpers.redact_credentials(test_str) == test_result
 
 
+def test_get_sql_from_file():
+
+    # Test query string
+    test_str = "select * from schema.tablename limit 10"
+
+    # Create fake file.
+    os.mkdir("tmp")
```
Jason94, line 159: The GMail tests are failing because they're also trying to create this directory, which isn't being cleaned up in the test. If we do decide to retain the `get_sql_from_file` method, the test should be rewritten to use the `tempfile` module (https://docs.python.org/3/library/tempfile.html).

parsons/utilities/sql_helpers.py
```diff
@@ -12,17 +12,3 @@ def redact_credentials(sql):
     sql_censored = re.sub(pattern, "CREDENTIALS REDACTED", sql, flags=re.IGNORECASE)
 
     return sql_censored
-
-
-def get_sql_from_file(sql_file):
```
shaunagm: Why are we removing this function?

parsons/databases/table.py
```diff
@@ -214,18 +214,17 @@ def dedup_table(
...
-            from {self.table}_temp_{current_timestamp})
+            from {self.table})
             where dup=1;
-            alter table {self.table}
+            alter table {self.table}_temp_{current_timestamp}
             drop column dup;
-            drop table {self.table}temp_{current_timestamp} {run_cascade};
+            truncate table {self.table}
+            insert into {self.table} (select * from {self.table}_temp_{current_timestamp})
+             {run_cascade};
```
shaunagm, line 227: Apologies as my SQL is rusty, but do we need to drop the temp table after copying it to the main table?

parsons/databases/table.py
shaunagm: @Jason94 does this approach fix the views issues too? (I don't know what the views issues are, as a non-expert SQL person, just want to make sure we're not merging buggy code.)

parsons/utilities/sql_helpers.py
sharinetmc: Mostly because, as Jason noted, it seems to just be wrapping f.read. That would mean this function isn't adding value beyond just using the File.reader() method. So, I thought that having a whole new function just for this isn't of much use, perhaps?

parsons/databases/table.py
sharinetmc: Good point!

parsons/utilities/sql_helpers.py
shaunagm: Makes perfect sense, but I'm hesitant to remove anything that someone else might be depending on. It looks like you removed this removal from the PR, but if you wanted to put it back in (or make a separate PR) we should probably put a deprecating warning on and then remove it later.

shaunagm: Not sure why github is doing this, at this point it might make sense to just do a fresh PR (my solution for whenever github is confusing me lol)

PR #961, opened 2023-12-19
Canalespy gsheet helper functions

sharinetmc:
Added some internal tmc canalespy gsheet functions to parsons! One function recursively re-attempts a specified gsheet method, while the other combines data from multiple gsheets into one parsons table. Also, not quite sure why there's the import issue in the unit tests?


parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
...
 
+    def attempt_gsheet_method(self, method, i=1, max=6, wait_time=15, **kwargs):
+        """
+        The Google Sheets API has notoriously strict rate limits (e.g. 60 calls per minute). This
+        function calls itself (i.e. is recursive) to help configure wait times and retry attempts
+        needed to wait out rate limit errors instead of letting them derail a script.
+        `Args:`
+            method: str
+                The name of the Parsons GoogleSheets method to be attempted
+            i: int
```
Jason94, line 410: Instead of exposing recursion control parameters to the public API, it is probably going to be cleaner to define an inner-function like this:

```python
def foo(a, b):
    def inner_foo(a, b, i):
         # <Do Recursive Stuff Here...>
    inner_foo(a, b, 0)
```

parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
...
+        except (APIError, HTTPError, ReadTimeout, ConnectionError) as e:
+            # Lets get the ordinals right, because why not
+            if i % 10 == 1:
+                ordinal = "st"
+            elif i % 10 == 2:
+                ordinal = "nd"
+            else:
+                ordinal = "th"
+
+            logger.info(f"trying to {method} for the {i}{ordinal} time")
```
Jason94, line 453: Could this be a debug log instead of an info log?

parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
...
+            if i < max:
+                time.sleep(wait_time)
+                i += 1
+                output = self.attempt_gsheet_method(method, i, max, wait_time, **kwargs)
```
Jason94, line 457: Can we just do `return self.attempt_gsheet_method(...)` instead of saving it as `output`?

(It doesn't _really_ matter because Python doesn't support tail call optimization anyway, but I do think it's cleaner...)

parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
...
+                ordinal = "st"
+            elif i % 10 == 2:
+                ordinal = "nd"
+            else:
+                ordinal = "th"
+
+            logger.info(f"trying to {method} for the {i}{ordinal} time")
+            if i < max:
+                time.sleep(wait_time)
+                i += 1
```
Jason94, line 456: Instead of incrementing the argument, can we call the recursive call with `i+1`? That way we're not mutating our arguments, which is bad-ish practice.

parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
         ws.format(range, cell_format)
         logger.info("Formatted worksheet")
 
+    def attempt_gsheet_method(self, method, i=1, max=6, wait_time=15, **kwargs):
```
Jason94, line 402: Can we adjust this to also take an `*args` splat variable? That's the standard way to solve the positional arguments problem that you note in the docstring. The Python standard for function passthrough like you're doing here is to do:

```python
def call_foo(my_function, *args, **kwargs):
    # do stuff...
    my_function(*args, **kwargs)
```

parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
...
+                output = self.attempt_gsheet_method(method, i, max, wait_time, **kwargs)
+
+            else:
+                raise e
+
+        return output
+
+    def combine_multiple_sheet_data(self, sheet_ids, worksheet_id=None):
+        """
+        Combines data from multiple Google Sheets into a Parsons Table.
```
Jason94, line 466: Could you give a little more description about how the data will be combined? Will they be treated as if the spreadsheets are stacked on top of each other (concatenated)? Do the columns have to align positionally or by name? Do the data types need to match? Etc.

parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
...
+        `Args:`
+            sheet_ids: str, list
+                The IDs of the Google Spreadsheets with that data to be combined. Can be a
+                comma-separated string of IDs or a list.
+
+            worksheet_id: str
```
Jason94, line 472: Should be marked `(optional)`

parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
...
+            # Comma-separated string
+            sheet_id_list = [x.strip() for x in sheet_ids.split(",")]
+
+        else:
+            raise ValueError(f"{sheet_ids} is not a valid string or list GSheet IDs")
+
+        # Non-DB table options yield a list, convert to Parsons table with default worksheet col
+        if sheet_id_list:
+            sheet_id_tbl = Table(
+                [{"sheet_id": x, "worksheet_id": 0} for x in sheet_id_list]
```
Jason94, line 496: Should this be `id_col` instead of the hardcoded `"sheet_id"`? That way we keep consistent and in case it gets changed, this doesn't cause a bug.

parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
...
+        elif "," in sheet_ids:
+            # Comma-separated string
+            sheet_id_list = [x.strip() for x in sheet_ids.split(",")]
+
+        else:
+            raise ValueError(f"{sheet_ids} is not a valid string or list GSheet IDs")
+
+        # Non-DB table options yield a list, convert to Parsons table with default worksheet col
+        if sheet_id_list:
+            sheet_id_tbl = Table(
```
Jason94, line 495: What is the benefit of making this a table instead of a list of dictionaries?

Jason94: Could it just be `sheet_id_tbl = [{"sheet_id": x, "worksheet_id": 0} for x in sheet_id_list]`? I think that would even work as-is. (Although we'd probably want to change the name...)

parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
...
+            # Retrieve sheet title (with attempts to handle rate limits) and add as a column
+            globals()["sheet_obj"] = self.attempt_gsheet_method(
+                "gs.gspread_client.open_by_key", key=sheet_id[id_col]
+            )
+            sheet_title = str(self.attempt_gsheet_method("sheet_obj.title"))
+            data.add_column("spreadsheet_title", sheet_title)
+
+            # Accumulate and materialize
+            combined.concat(data)
+            temp_files.append(combined.materialize_to_file())
```
Jason94, line 540: What is the benefit to using the `materialize_to_file` method? Looking at the source, it looks like it's still updating the table in memory?

parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
...
+
+        return combined
```
Jason94, line 542: We should probably close out all of the remaining temp files before we return. In fact, we might want to wrap that in a `finally` block to make sure we don't leave temp files around in case this function errors.

parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
...
+                wait_time=60,
+                spreadsheet_id=sheet_id[id_col],
+                worksheet=sheet_id[worksheet_id],
+            )
+
+            # Add the sheet ID as a column
+            data.add_column("spreadsheet_id", sheet_id[id_col])
+
+            # Retrieve sheet title (with attempts to handle rate limits) and add as a column
+            globals()["sheet_obj"] = self.attempt_gsheet_method(
```
Jason94, line 532: Instead of polluting the global namespace, we should be able to store it in the current instance. This line and the next should probably be something like:
```python
self.__sheet_obj = self.attempt_gsheet_method(...)
sheet_title = str(self.attempt_gsheet_method("self.__sheet_obj.title")
del self.__sheet_obj
```

parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
...
+                "gs.gspread_client.open_by_key", key=sheet_id[id_col]
+            )
+            sheet_title = str(self.attempt_gsheet_method("sheet_obj.title"))
+            data.add_column("spreadsheet_title", sheet_title)
```
Jason94, line 536: Could you document the two columns that this function will add to the data in the docstring?

Jason94 (changes requested):

parsons/google/google_sheets.py
```diff
@@ -395,6 +399,148 @@ def format_cells(self, spreadsheet_id, range, cell_format, worksheet=0):
...
+                spreadsheet_id=sheet_id[id_col],
+                worksheet=sheet_id[worksheet_id],
+            )
+
+            # Add the sheet ID as a column
+            data.add_column("spreadsheet_id", sheet_id[id_col])
+
+            # Retrieve sheet title (with attempts to handle rate limits) and add as a column
+            globals()["sheet_obj"] = self.attempt_gsheet_method(
+                "gs.gspread_client.open_by_key", key=sheet_id[id_col]
```
Jason94, line 533: I think this is the culprit of your test failures. `gs` isn't in scope here. I think that it's supposed to be some GoogleSheets related module that gets imported as `gs`?

parsons/google/google_sheets.py
sharinetmc: Wondering what you think about this? Would python just automatically close the sheet itself, then?

sharinetmc:
@Jason94 @shaunagm, I think I addressed all the comments, but am still getting failed checks. However, I'm really not to sure how to resolve these issues pasted below:

ImportError: cannot import name 'GoogleSheets' from 'parsons' (/home/runner/work/parsons/parsons/parsons/__init__.py)

shaunagm:
Maybe you can combine the changes from this and #960 into a fresh PR? These are out of date enough that simply updating to the most recent version of main might fix the existing issues.

PR #963, opened 2023-12-20
Enable email dict to be passed to VAN.upsert_person()

austinweisgrau:
This allows for greater configurability of email data when upserting person records in EveryAction. This change is backwards compatible, but allows, for example, setting emails to not be subscribed by default when loading into EveryAction.

### DIFF
--- a/parsons/ngpvan/people.py
+++ b/parsons/ngpvan/people.py
@@ -1,4 +1,5 @@
 from parsons.utilities import json_format
+from typing import Union, List, Dict
 import logging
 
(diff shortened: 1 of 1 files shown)


Jason94 (approved):
Awesome!

shaunagm:
@austinweisgrau is this good to merge? For the record, I trust either/both of you to hit the "squash and merge" button if you feel a PR's ready (and you should both have github permissions)

austinweisgrau:
Ok great will do!

PR #965, opened 2024-01-02
Action Builder Deactivate & Reactivate Connections

ydamit:
Connection records in Action Builder cannot be deleted. Instead, there's an "Inactive" flag that determines whether they appear in the UI or not on either of the two Entity records they connect. This PR introduces a new method, `deactivate_connection()`, which deactivates an existing Connection. 

The method requires the "interact ID" for one of the two Entities as a positional argument. One of two additional (keyword) arguments is required to make the change: either the interact ID for the other Entity (POST request to the connection helper endpoint), or the interact ID of the connection itself (PUT request to the specific connection endpoint).

This PR also updates the existing `upsert_connection()` method to include by default an explicit `False` value for the `inactive` flag when upserting Connection data, under the assumption that upserting to a given Connection implies that the user wants that Connection to be active. The `reactivate` argument can be set to `False` to prevent this and make changes to inactive Connections _without_ reactivating them.

Tests were written and adapted for these additions and changes, and are running successfully. Furthermore, these changes were successful in live tests done by defining the two methods locally and overriding the Class definitions before instantiation.

A good future PR might be to adapt the existing `upsert_connection()` method to accept the Connection ID as an optional keyword arg, to trigger a PUT request directly to the specific Connection endpoint. Argument validation could be updated such that a single-item list (or even a string) could be accepted for the positional `identifiers` arg if the Connection ID is provided.

### DIFF
--- a/parsons/action_builder/action_builder.py
+++ b/parsons/action_builder/action_builder.py
@@ -425,11 +425,11 @@ class ActionBuilder(object):
         from_identifier,
         connection_identifier=None,
         to_identifier=None,
-        campaign=None
+        campaign=None,
     ):
         """
         Deactivate an existing connection record in Action Builder between two existing entity
-        records. Only one connection record is allowed per pair of entities, so this can be done 
+        records. Only one connection record is allowed per pair of entities, so this can be done
         by supplying the ID for the connection record, or for the two connected entity records.
         `Args:`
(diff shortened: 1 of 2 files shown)


IanRFerguson (approved):
Looks great!

PR #969, opened 2024-01-18
add client_options with default scopes to client

graemelorimer:
Add a client_options parameter to the client function, including default scopes. These scopes are necessary to allow querying of external tables stored in Google Drive. The formatting should make it possible to pass in other client options as required.


willyraedy (approved):
Hoping to get this release soon-ish to unblock some migration work. @shaunagm do you know what might be the problem with the build dependency?

shaunagm:
Looks like we need to bump the sphinx version to 5.0 or later

PR #973, opened 2024-01-21
Update get_columns_list() to accept tables with spaces in the name

coastlines:
Updates `get_columns_list()` to work with table names that include spaces or begin with invalid characters.

### DIFF
--- a/parsons/databases/redshift/rs_table_utilities.py
+++ b/parsons/databases/redshift/rs_table_utilities.py
@@ -428,7 +428,8 @@ class RedshiftTableUtilities(object):
         `Returns:`
             A list of column names.
         """
-
+        schema = f'"{schema}"'
+        table_name = f'"{table_name}"'
         first_row = self.query(f"select * from {schema}.{table_name} limit 1")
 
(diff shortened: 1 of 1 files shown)


parsons/databases/redshift/rs_table_utilities.py
```diff
@@ -428,7 +428,8 @@ def get_columns_list(self, schema, table_name):
         `Returns:`
             A list of column names.
         """
-
+        schema = f'"{schema}"'
```
Jason94, line 431: What if someone was already working around this issue by passing in the name of the schema wrapped in quotes? Should we check if `schema` and `table_name` are already wrapped in quotes, and only add them if they aren't already?

Jason94 (changes requested):

parsons/databases/redshift/rs_table_utilities.py
coastlines: Ah, good catch. Yes, I think you are right.

Jason94 (approved):
Looks good! Thanks, Erica!

PR #974, opened 2024-01-22
BigQuery - Call .result() regardless of compression

IanRFerguson:
We found that some of the `try`/`except`/`finally` logic employed in the `GoogleBigQuery` connector is suppressing exceptions. This PR explicitly returns a `LoadJobConfig` object when a flat file is forcibly decompressed, such that any errors raised are encountered in the `try` block

### DIFF
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -420,7 +420,7 @@ class GoogleBigQuery(DatabaseConnector):
 
         try:
             if force_unzip_blobs:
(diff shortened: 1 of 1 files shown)


shaunagm:
This looks good to me. No problems with the tests - just the issue with the Sphinx versioning that still needs to be fixed (not an issue with this PR)

shaunagm (approved):

Jason94 (approved):

PR #975, opened 2024-01-22
Bump Multiple Docs Deps Around Sphinx

SorenSpicknall:
Several docs-related dependencies in Parsons are pretty far behind their latest project versions, and began causing [CI failures](https://github.com/move-coop/parsons/pull/969) related to Sphinx. This bumps Sphinx up to the latest 5.x version before introduction of 6.0, bumps sphinx-rtd-theme correspondingly, and attempts to upgrade myst-parser as well. If build steps fail, we will likely need to be a little less aggressive with the version number leaps of the latter two packages.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,5 @@
 -r ../requirements.txt
-Sphinx==4.3.2
-sphinx-rtd-theme==1.0.0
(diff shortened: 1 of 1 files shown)


willyraedy (approved):
Thanks Soren!

PR #976, opened 2024-01-22
fixes copy_s3 by using generic GCS function

willyraedy:
During the migration, we generalized a function to use the Transfer Service to copy file(s) from an arbitrary bucket rather than having bespoke functions for S3, GCS etc. However, we missed one reference to the old file. 

Austin this should get you started. Could stand some more manual testing and we should add a unit test here obviously. I can get to that probably post the Feb 5th release of BQ to the rest of the members but wanted to get you what I had.


elyse-weiss (approved):
LGTM!

shaunagm:
Looks like there's a test failure here

shaunagm (changes requested):
Just changing the review status to 'request changes' instead of 'approved' due to the test failure

bmos:
I will investigate in the next few days.

shaunagm:
Thanks to @bmos the tests are now passing. Is this otherwise ready to merge? @austinweisgrau it seems like you have more context than me, if you can give it the thumbs up I will merge it.

shaunagm:
@willyraedy @austinweisgrau what's the status of this PR? The tests are passing now, but it's not clear whether this is still needed. Happy to help get it over the line if so.

bmos (approved):
This is still an issue, and the fix in this PR is appropriate.

PR #983, opened 2024-02-05
Refactor Table initialization to avoid iterating over every row

austinweisgrau:
The prior implementation for initializing a Parsons table involved a check of the truthiness of a Parsons Table's underlying petl Table, which involves iterating over every row of the table. For very long tables, this can take a very long time.

These changes refactor how a parsons Table is initialized slightly to avoid this iteration and more robustly and performatively set & check for the validity of the underlying petl Table.

The result involves one bit of esoteric code (the use of a sentinal as a default argument) in order to allow existing behavior to remain the same, specifically that `Table()` should work but `Table(None)` should raise an Error. I think that `Table(None)` should work equivalently to `Table()` and would simplify the implementation somewhat, but that would be a breaking change so I will make that in a separate PR to the major-release branch.


Jason94:
What's the relationship between this PR and #984? It looks like that commit removes code that's added here? Could they be consolidated?

austinweisgrau:
This one does not make any breaking changes, so it can be merged into main without a major release.

Currently Table(None) would raise an error, while Table() is okay.

PR #984 is one commit ahead of this branch, it's just one commit that makes a breaking change.  It makes Table(None) work the same as Table().

Jason94 (approved):
Thanks for clarifying. This looks great to me, then. Thanks Austin!

PR #984, opened 2024-02-05
Allow Table(None)

austinweisgrau:
This change allows `Table(None)` to work, equivalent to `Table()`. This is reasonable behavior and also simplifies the implementation to avoid the necessity of a sentinal as a default argument. 

This may be a breaking change if code depends on `Table(None)` raising a ValueError.

### DIFF
--- a/parsons/etl/table.py
+++ b/parsons/etl/table.py
@@ -1,6 +1,5 @@
 import logging
 import pickle
-from enum import Enum
 from typing import Union
 
 import petl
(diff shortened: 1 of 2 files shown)


austinweisgrau:
Just one commit here, which will be more clear once major-release is brought up to date. no urgency on my end there.

Jason94:
I think this change makes sense, and I took a look at the operative commit (https://github.com/move-coop/parsons/pull/984/commits/62ceca67a32bffc3dcacec4653a5c04c7f398102) and it looks good to me.

Jason94 (approved):

shaunagm:
Sorry about the delay - there's a merge conflict in major-release so it can't be trivially merged back into main. I've asked Willy and Kasia to look into it, and hopefully we'll have that handled (and this PR merged) within the week.

austinweisgrau:
The major release branch needs to be hard reset to match main.
Alternatively, we start major release branches that are particular to the
version - so we could have a major-release-4 branch

shaunagm:
Worth talking about switching our process in the next contributor meeting (two weeks from Thursday). In the meantime I can reset major-release.

shaunagm:
Actually switching to using a schema like `major-release-4` was logistically easier. But unfortunately there's still a few merge conflicts.

shaunagm:
Hmm, it's possible I didn't think this through - this should probably be `major-release-5`. Would've been easier to make a new branch for that of main then change the target of the PR.

shaunagm:
I think we can get away with just renaming `major-release-4` to `major-release-5`.

PR #994, opened 2024-02-19
992 bug parsons bigquery upsert is broken

willyraedy:
Fixes the upsert bug by adding "template_table" back into the API for the three copy methods (copy, copy_s3, copy_from_gcs). 

Moved the schema determination logic into a single private function and moved some error handling into a lower level wrapper of the BQ client's `load_table_from_uri`. Also, did some basic tidying in a few functions.

These are some of the most central methods in the class. Obviously the unit tests are passing, your test script, and I got halfway through our Catalist load locally. But if people can do some meaningful kicking the tires with integration tests in your projects, I think that would be a good use of time.


parsons/google/google_bigquery.py
```diff
@@ -1136,6 +1109,46 @@ def get_row_count(self, schema: str, table_name: str) -> int:
 
         return result["row_count"][0]
 
+    def get_table_ref(self, table_name):
+        # Helper function to build a TableReference for our table
+        parsed = parse_table_name(table_name)
+        dataset_ref = self.client.dataset(parsed["dataset"])
+        return dataset_ref.table(parsed["table"])
+
+    def _get_job_config_schema(
```
willyraedy, line 1118: moved all the schema logic, including the stuff Austin added a couple PRs ago, into this one function

parsons/google/google_bigquery.py
```diff
@@ -1136,6 +1109,46 @@ def get_row_count(self, schema: str, table_name: str) -> int:
...
+                return bigquery_table.schema
+            except google.api_core.exceptions.NotFound:
+                logger.warning(
+                    f"template_table '{template_table}' not found. Unable to set schema."
+                )
+        # if load is coming from a Parsons table, use that to generate schema
+        if parsons_table:
+            return self._generate_schema_from_parsons_table(parsons_table)
+
+        return None
```
willyraedy, line 1150: this logic is probably the most important to review closely

parsons/google/google_bigquery.py
```diff
@@ -1267,6 +1295,38 @@ def _fetch_query_results(self, cursor) -> Table:
...
+        except exceptions.BadRequest as e:
+            for idx, error_ in enumerate(load_job.errors):
+                if idx == 0:
+                    logger.error(
+                        "* Load job failed. Enumerating errors collection below:"
+                    )
+                logger.error(f"** Error collection - index {idx}:")
+                logger.error(error_)
+
+            raise e
```
willyraedy, line 1328: moved this logic to loop through the error collections into the lower level function because I think we'll always want to do this. @IanRFerguson I know we had that weird edge case where some of this logic was throwing. Do you remember what that was? (should have made a unit test for it)

willyraedy (commented):
I think this is definitely better but still doesn't feel all that clean. A few notes

parsons/google/google_bigquery.py
```diff
@@ -349,6 +342,7 @@ def copy_from_gcs(
         force_unzip_blobs: bool = False,
         compression_type: str = "gzip",
         new_file_extension: str = "csv",
+        template_table: str = None,
```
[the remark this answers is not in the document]
Jason94: This should be an `Optional[str]`, right?

parsons/google/google_bigquery.py
```diff
@@ -1136,6 +1109,46 @@ def get_row_count(self, schema: str, table_name: str) -> int:
...
+
+    def _get_job_config_schema(
+        self,
+        job_config: LoadJobConfig,
+        destination_table_name: str,
+        if_exists: str,
+        parsons_table: Optional[Table] = None,
+        custom_schema: Optional[list] = None,
+        template_table: Optional[str] = None,
+    ) -> LoadJobConfig:
```
Jason94, line 1126: I think that the return type of this function is wrong. It looks like it should return an `Optional[List[SchemaField]]`, not a `LoadJobConfig` instance.

test/test_databases/test_bigquery.py
```diff
@@ -338,8 +338,9 @@ def test_copy(self):
         gcs_client = self._build_mock_cloud_storage_client(tmp_blob_uri)
         tbl = self.default_table
         bq = self._build_mock_client_for_copying(table_exists=False)
-        bq.copy_from_gcs = mock.MagicMock()
-        table_name = ("dataset.table",)
+        bq._load_table_from_uri = mock.MagicMock()
```
Jason94, line 341: To be honest, I'm not sure if these unit tests have any value. I'm a big fan of tests, and even TDD where it makes sense, but some code just doesn't have any testable surface area. (Outside of integration tests that do actually hit the external service.)

There's a few code smells here like mocking private methods, and asserting that private methods were called a certain number of times.

This might be too big of a change for this commit, but my vote would be that we just scrap these tests.

Jason94: At the very least, I think a better long term direction for these tests would be to mock out the `client` object that the BQ connector holds on to, but _not_ to mock the private methods of the BQ connector itself.

Jason94: (Again, I don't think that's worth changing in this PR, though.)

Jason94 (changes requested):
I think this is working. I tested in on my simple script in the issue I posted, and that worked fine. I have a longer script that is failing, but I think that's because of a separate issue unrelated to `upsert` specifically. I'm going to do a little more testing on that and, assuming it's unrelated, then I'll approve after these few small changes are made.

test/test_databases/test_bigquery.py
austinweisgrau: As an aside, this is an endemic problem for many parsons tests. There many tests that are essentially
```
Foo.bar = mock.MagicMock()
Foo.bar.return_value = 'fibble'
result = Foo.bar()
assert result == 'fibble'
```

shaunagm: Yeah, more than willing to rethink how we do tests, but don't let it get in the way of this particular PR.

Jason94: Yeah, I guess that Will should probably just leave this as-is in this PR. More of a philosophical point, then.

parsons/google/google_bigquery.py
```diff
@@ -814,16 +790,14 @@ def copy(
...
-                gcs_blob_uri=temp_blob_uri,
-                table_name=table_name,
-                if_exists=if_exists,
-                max_errors=max_errors,
+            self._load_table_from_uri(
+                source_uris=temp_blob_uri,
+                destination=self.get_table_ref(table_name=table_name),
                 job_config=job_config,
                 **load_kwargs,
             )
```
willyraedy, line 803: @Jason94 I think this might be the source of the bug you reported in slack. `copy` used to call the `copy_from_gcs` under the hood which meant it used the defaults set in the `copy_from_gcs` function signature, which are different from the defaults of the underlying google api

I think our options are:
1. add those same parameters and defaults to the `copy` function
2. modify the `copy` function to still call `copy_from_gcs` under the hood

I think I'd vote number 1. By calling the lower level `_load_table_from_uri` function, it made it a little easier to manage the schema stuff but we could probably figure it out.

parsons/google/google_bigquery.py
IanRFerguson: yeah that was when the Catalist flat file size exceeds Google's maximum compressed file size ... any of the state-level models are a good test case for that

parsons/google/google_bigquery.py
Jason94: Yeah, I agree. #1 seems fine to me.

Jason94: For reference, here's the script:

```python
import dotenv
from parsons import Table
from parsons.google.google_bigquery import GoogleBigQuery

dotenv.load_dotenv(override=True)


def main():
    bq = GoogleBigQuery()

    data = Table([{"id": 1, "text": "Test\ntext"}])

    bq.copy(data, "test.newlines", if_exists="drop")


if __name__ == "__main__":
    main()
```

Jason94 (approved):
Everything looks good to me, except for the issue that all of the instances of the `template_table` parameter in the `google_bigquery.py` file need to be `Optional[str]` not `str`.

I've tested it against the upsert and newlines bug and it passes both! Once you fix the issue above, it looks good to me, so I'm going to go ahead and approve.

Issue #995, opened 2024-02-20
[Bug] GoogleBigQuery get_row_count() method is inefficient and should be changed


matthewkrausse:
The current code runs this sql to get the row data. 

sql = f"SELECT COUNT(*) AS row_count FROM `{schema}.{table_name}`"

But we should change it to this:

SELECT table_name, row_count 
FROM project.dataset.INFORMATION_SCHEMA.TABLES 
WHERE table_name = 'your_table_name'

## Detailed Description

SELECT COUNT(*):

- Scans all rows in the table to count them individually.
- Reads the entire data size of the table.
- Can be computationally expensive, especially for large tables.

INFORMATION_SCHEMA.TABLES:

- Retrieves metadata about the table from the INFORMATION_SCHEMA.TABLES view.
- This view stores pre-aggregated information about tables, including row count.
- Only reads a small amount of data from the INFORMATION_SCHEMA.TABLES view.
- Is significantly faster and cheaper than scanning the entire table.

Therefore, by using the INFORMATION_SCHEMA.TABLES method, you only need to read a small amount of metadata instead of the entire table data, resulting in less data usage and improved performance.

## Your Environment

* Version of Parsons used (if you know):
* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):

## Priority
Please indicate whether fixing this bug is high, medium, or low priority for you. If the issue is time-sensitive for you, please let us know when you need it addressed by.

I believe this should be high priority to change as it could be costly for people using this method. I opened an issue mostly for discussion on this issue before writing the PR.

matthewkrausse:
I'm going to keep looking into this to see if there may be a better way.

austinweisgrau:
Nice catch

matthewkrausse:
@austinweisgrau I'm not sure if you saw my additional comments but this seems to be not as straightforward. There is another method I see where we can query the metadata another way via the api. 

```
from google.cloud import bigquery

client = bigquery.Client()
table_ref = client.dataset("your_dataset").table("your_table_name")
table = client.get_table(table_ref)
row_count = table.num_rows

print(f"Row count: {row_count}")
```

However, this only works on tables, not views, and apparently is potentially delayed by a few seconds. 

At this point in research it seems COUNT(*) may be the only up-to-date, accurate method for getting the row count.

PR #998, opened 2024-02-21
Add cautionary note about performance in get_row_count method

matthewkrausse:
This pull request adds a cautionary note about the performance of the get_row_count method in the BigQuery materialization. The note warns that using SELECT COUNT(*) can be expensive for large tables, especially those with many columns, as BigQuery scans all table data to perform the count. The note aims to inform developers about the potential performance impact and encourage them to consider alternative approaches when dealing with large tables.

### DIFF
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -1121,6 +1121,10 @@ class GoogleBigQuery(DatabaseConnector):
         """
         Gets the row count for a BigQuery materialization.
 
+        Caution: This method uses SELECT COUNT(*) which can be expensive for large tables,
+        especially those with many columns. This is because BigQuery scans all table data
+        to perform the count, even though only the row count is returned.
+
         `Args`:
             schema: str
                 The schema name


Jason94:
@matthewkrausse Do you think it might be worth changing this method to use the `INFORMATION_SCHEMA` if it detects it's a table? Apparently that can be out of date sometimes, so maybe we could add a parameter to force it to do a `COUNT(*)` query?

https://www.metaplane.dev/blog/four-efficient-techniques-to-retrieve-row-counts-in-bigquery-tables-and-views

Jason94:
> @Jason94 Yeah, that was what prompted this PR, I made an issue #995. I think that could be a good option but not sure how useful it is given the use case. Usually when I want the row count of a table it is to make a comparison. Thoughts?

That's a good point. I'm sure how much of a concern the computational resources of running a big COUNT(*) would be. Maybe make COUNT(*) a default, and add an extra boolean flag for something like `fast_mode=False` that would switch it to use INFORMATION_SCHEMA?

Maybe that's not something that would actually be helpful to the way people are using this method in production code? Maybe we could keep it simple and just always use COUNT.

matthewkrausse:
Yeah, I think it's useful when someone is just looking to see if a table has 0 records, 10 records or 10 million records.But, I don't feel like anyone is doing that programatically. They are likely just using the BQ Studio.

shaunagm:
So I'm happy to merge this as is, but another approach would be to create another method like `get_row_count_estimate` that uses the `information_schema.tables` mentioned in https://github.com/move-coop/parsons/issues/995. Then in this doc warning, we could say "if an estimate is fine, use X other method" (and the docstring for estimate could point people to `get_row_count` for when accuracy is important.

matthewkrausse:
> So I'm happy to merge this as is, but another approach would be to create another method like `get_row_count_estimate` that uses the `information_schema.tables` mentioned in https://github.com/move-coop/parsons/issues/995. Then in this doc warning, we could say "if an estimate is fine and you don't want the performance hit, use `get_row_count_estimate`" (and the docstring for `get_row_count_estimate` could point people to `get_row_count` for when accuracy is important or they don't care about performance.

Let's go ahead and merge this as is.

shaunagm (approved):
LGTM

Issue #989, opened 2024-02-08
[Feature/Addition] Parsons table head


elyse-weiss:
I would love for petl's head function to be built into Parsons. This allows you to just get the first X number of rows from a table. Sometimes you want to test code on a smaller subset of data, or something like that, and this is very helpful.

shaunagm:
Thanks @matthewkrausse

PR #1001, opened 2024-02-24
Add head and tail methods to Parsons Table

matthewkrausse:
This pull request adds the `head` and `tail` methods to the `Parsons Table` class in the `ETL` module. The `head` method returns the first n rows of the table, while the `tail` method returns the last n rows of the table. Tests and documentation have also been added for these new methods.

### DIFF
--- a/parsons/etl/etl.py
+++ b/parsons/etl/etl.py
@@ -9,6 +9,37 @@ class ETL(object):
     def __init__(self):
         pass
 
+    def head(self, n=5):
+        """
+        Return the first n rows of the table
+
+        `Args:`
+            n: int
+                The number of rows to return. Defaults to 5.
+        `Returns:`
+            `Parsons Table`
+        """
+
+        self.table = petl.head(self.table, n)
+
+        return self
+
+    def tail(self, n=5):
+        """
+        Return the last n rows of the table. Defaults to 5.
+
+        `Args:`
+            n: int
+                The number of rows to return
+
+        `Returns:`
+            `Parsons Table`
+        """
+
+        self.table = petl.tail(self.table, n)
+
+        return self
+
     def add_column(self, column, value=None, index=None, if_exists="fail"):
         """
         Add a column to your table
@@ -724,7 +755,7 @@ class ETL(object):
                 The new long table
         """
 
-        if type(key) == str:
+        if type(key) is str:
             key = [key]
 
         lt = self.cut(*key, column)  # Create a table of key and column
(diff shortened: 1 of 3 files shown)


matthewkrausse:
This addresses issue #989 to add the head method to the Parsons Table. @elyse-weiss

test/test_etl.py
```diff
@@ -997,3 +997,15 @@ def test_deduplicate(self):
         tbl_expected = Table([["a", "b", "c"], [1, 2, 3], [1, 3, 2], [2, 3, 4]])
         tbl.deduplicate(["a", "b"])
         assert_matching_tables(tbl_expected, tbl)
+
+    def test_head(self):
+        tbl = Table([["a", "b"], [1, 2], [3, 4], [5, 6], [7, 8], [9, 10]])
+        tbl_expected = Table([["a", "b"], [1, 2], [3, 4]])
+        tbl.head(2)
```
shaunagm, line 1004: Looking at `tbl_expected`, this seems like it's returning three rows, not 2? Is the count zero-indexed? If so we should warn people in the docs.

shaunagm (changes requested):
Looks good, except for my question about whether n is zero-indexed or not.

parsons/etl/etl.py
```diff
@@ -9,6 +9,37 @@ class ETL(object):
     def __init__(self):
         pass
 
+    def head(self, n=5):
+        """
+        Return the first n rows of the table
+
+        `Args:`
+            n: int
```
shaunagm, line 17: Oh I guess there's also the question of whether `n` is the right variable name. Generally we try to avoid single letter variable names, but if we were going to make an exception this seems pretty reasonable as "n" is generally understood to mean "number". But it might be more readable to name this "rows".

matthewkrausse: I believe n=5 is the standard for this function in most cases. That is how they do it in Pandas: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.head.html

shaunagm: I'm fine with keeping it as is. If any single-letter variable is comprehensible, it's this one.

test/test_etl.py
Jason94: That's the `Table` constructor. The first array is the list of column names, then the subsequent arrays are the actual rows.

matthewkrausse: Thanks for reviewing. I agree.

shaunagm: Whoops 😂

shaunagm (approved):
Looks good to me, now that I've been reminded of how the Table constructor works

PR #1003, opened 2024-02-25
EveryAction: Add email endpoint methods to retrieve email stats from TargetedEmail

matthewkrausse:
This pull request adds an email endpoint to the VANConnector class, allowing users to retrieve email messages and their stats. It also includes test cases for getting email messages and email stats.


matthewkrausse:
For a little more context, the get_email_stats method uses the get_emails and get_email method in order to unpack a bunch of information. You need to get a list of all emails first with get_emails(). Then you use foreginMessageId with the get_email() method to get the stats. The data is nested because with any ab testing that was done. So for instance, inside the  emailMessageContentDistributions section, there is a list of dictionary objects with email_A, email_B, email_c and the email_winner. The get_email_stats method combines all these to get the final open count, unsubs, etc.

matthewkrausse:
Another thing to note, which I would take suggestions for, is that you need to request that this endpoint is turned on for your api key. They don't expose it by default. So we will want to document that somewhere.

parsons/ngpvan/email.py
```diff
@@ -0,0 +1,143 @@
...
+        orderby=None,
+    ):
+        """
+        Get emails.
+
+        Note that it takes some time for the system to aggregate opens and click-throughs,
+        so data can be delayed up to 15 minutes.
+
+        `Args:`
+            orderby : str
```
anzelpwj, line 23: Since the only real choice to be made is choosing to have things sorted ascending or descending by dateModified, would it make more sense to have `orderby` be something like `ascending=True`, and we then send the appropriate `dateModified%20asc` or `dateModified%20desc` value?

parsons/ngpvan/email.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def get_emails(
+        self,
+        orderby=None,
+    ):
+        """
+        Get emails.
+
+        Note that it takes some time for the system to aggregate opens and click-throughs,
+        so data can be delayed up to 15 minutes.
```
anzelpwj, line 20: Question - is this more of an issue for `get_emails` or the iteration in `get_email_stats`?

parsons/ngpvan/email.py
```diff
@@ -0,0 +1,143 @@
...
+
+        emails = self.get_emails()
+
+        foreign_message_ids = [email["foreignMessageId"] for email in emails]
+
+        for fmid in foreign_message_ids:
+            email = self.get_email(fmid)
+            email_list.append(email)
+
+        for email in email_list:
```
anzelpwj, line 92: Would it be helpful to have some progress indicator? `tqdm`?

parsons/ngpvan/email.py
```diff
@@ -0,0 +1,143 @@
...
+        return r
+
+    def get_email_stats(self):
+        """
+        Get stats for all emails.
+
+        `Args:`
+            emails : list
+                A list of email message details.
+
```
anzelpwj, line 74: Suggestion - maybe we could include `start_date` and `end_date` as optional parameters? I know that this won't affect `get_emails` but maybe might save some time for the L88 loop.

parsons/ngpvan/email.py
```diff
@@ -0,0 +1,143 @@
...
+            d["recipientCount"] = 0
+            d["bounceCount"] = 0
+            d["contributionCount"] = 0
+            d["contributionTotal"] = 0
+            d["formSubmissionCount"] = 0
+            d["linksClickedCount"] = 0
+            d["machineOpenCount"] = 0
+            d["openCount"] = 0
+            d["unsubscribeCount"] = 0
+            try:
```
anzelpwj, line 109: 1. Why do we need a try block here? What issues have we run into?
2. If we need it, would it maybe be better to have each of these assignments get their own try block so that - if something breaks with `bounceCount` (for example) - it doesn't affect all of the other assignment commands?

parsons/ngpvan/email.py
anzelpwj: Also, do we care about passing through `orderby`?

anzelpwj (commented):
Some thoughts looking at the code. Very excited for this!

parsons/ngpvan/email.py
matthewkrausse: This language comes directly from their docs. https://docs.ngpvan.com/reference/email-overview

parsons/ngpvan/email.py
matthewkrausse: I think it runs quick enough that this isn't necessary. Plus I don't believe we use tqdm or progress bars for many (if any) connectors in Parsons and not sure we want to add another dependency to our tangled mess.

parsons/ngpvan/email.py
matthewkrausse: I believe the problems I was running into was that when emails weren't sent yet, these fields did not exist. This was also code I wrote pretty early on in my career and a few months ago added it to parsons without looking too close into it because I was about to lose access to EveryAction and wanted to contribute before that went away.

parsons/ngpvan/email.py
matthewkrausse: I'm open to that. We have tried to keep our parameters to match the API docs params. But it does seems like a reasonable change.

parsons/ngpvan/email.py
anzelpwj: I'd imagine this is more of a get_email_stats problem, since we don't pull opens and click throughs until we put in the `$expand` parameter in `/email/message/email_id`.

parsons/ngpvan/email.py
anzelpwj: I'd suggest doing to simplify the API for the end user.

parsons/ngpvan/email.py
anzelpwj: Got it. Do we expect all of these fields to be missing simultaneously in the same `i` block?

Another question - are there cases where we would expect to loop multiple times?  Does this happen for emails that have A/B tests?

matthewkrausse: I believe so. 

Yes, if there are A/B tests it loops multiple times.

anzelpwj: Okay, then the try block should be fine.

Would it be possible to have the option of not aggregating over each email? For the reporting we do for Tech for Campaigns, we report on how each send did independently.

matthewkrausse: Yeah, we could create another method or could have a param that gives the option.

anzelpwj: As per our discussion on 2024-05-14 we can defer this request to a later PR.

shaunagm (approved):
LGTM

Issue #932, opened 2023-11-28
Documentation discrepancy regarding supported Python versions


mkrausse-ggtx:
The README.md file says "Parsons is only supported for Python 3.8-10." However on PyPI it says "Requires: Python >=3.7.0, <3.11.0"

Not sure which is correct.

Jason94:
Here is the code from `setup.py`. It looks like PyPi is correct. However, this brings up that it's high time we move on from Python 3.7. It was sunset from receiving security updates a little while ago, and even Python 3.8 only has 10 months left. ([Source](https://devguide.python.org/versions/)).

@shaunagm Has there been conversation about bumping the Python version? In the meantime, I suppose we should fix the README?

Jason94:
I guess that _allowing_ older versions isn't a problem. We should maybe consider adding 3.11, though. PETL (https://github.com/petl-developers/petl/blob/master/setup.py) still has Python 2.x listed as an allowed version! Although it does go up to 3.11.

I suppose the bigger issue with 3.7 is that it's still the Python version for the Dockerfile, but that's maybe a question for @willyraedy .

shaunagm:
@willyraedy are you okay with us no longer officially supporting 3.7? Will changing it in the setup.py file cause any issues for your automatic builds etc?

Ideally TMC doesn't need 3.7 and we can solve this issue by adding 3.11 to the test suite, fixing any bugs that appear in the tests under 3.11, and then updating the documentation and setup.py to reflect that we now officially support 3.8-3.11. 

If TMC wants to keep 3.7 listed we can. The only downside is if there are bugs that only occur when running parsons under 3.7, in which case we ought to handle them.

bmos:
I have been looking into 3.11 support.
At the moment, all the tests pass [except for one](https://github.com/bmos/parsons/actions/runs/8167109340/job/22326974373) and the error [appears to be](https://github.com/petl-developers/petl/pull/656/files) upstream.
I am [submitting a fix](https://github.com/petl-developers/petl/pull/656) there.

bmos:
Pytest does not pass on 3.7.
https://github.com/move-coop/parsons/actions/runs/8274714031/job/22640505554

jdw25:
I think this is now resolved with release of 3.2.0?

shaunagm:
Yes, correct! Thanks for the flag @jdw25

PR #1009, opened 2024-03-12
GitHub Actions & Dev Tools Cleanup + Python 3.11 Support

### LINKED ISSUE #1007: [Bug] Parsons 3.1.0 targets yanked 4.0.0 version of braintree with critical bugs
Parsons 3.1.0 depends on braintree 4.0.0 which has been yanked due to critical bugs.

## Detailed Description

When I install parsons 3.1.0 via pip, I get this message:
> WARNING: The candidate selected for download or install is a yanked version: 'braintree' candidate (version 4.0.0 at https://files.pythonhosted.org/packages/a5/1e/641306eec960161028e0a30d2f635abdfbdb89029330df1701bc106e5d8c/braintree-4.0.0-py2.py3-none-any.whl (from https://pypi.org/simple/braintree/))
Reason for being yanked: critical bugs

When I install it via [uv](https://github.com/astral-sh/uv) it doesn't even install for the same reason.

## To Reproduce

Create a new virtual environment and run `pip install parsons==3.1.0`

## Your Environment

* Version of Parsons used (if you know): **v3.1.0**
* Environment name and version (e.g. Chrome 39, node.js 5.4): 
* Operating System and version (desktop or mobile): **Pop!_OS 22.04 LTS**

## Additional Context

In a fork I'm playing around with, braintree==4.26.0 and google-cloud-bigquery==3.18.0 were needed to get pytests passing. I'm not sure why that isn't the case on this repo but it seems like these versions might be safe to update.

## Priority

For me it's low because I don't use braintree but "critical bugs" makes it sound rather serious for people who use that integration.

bmos:
I'm not really sure how these changes could be merged nicely with other open PRs.
I'm happy to restructure it however is desired, so perhaps treat this as a sort of menu.

## What's Included

- Simplified the linting workflow to use a single file for both mac and linux testing.
- Use the rust-based package installer [uv](https://github.com/astral-sh/uv) to significantly reduce the GitHub actions runtime to allow a second python version in macOS testing.
- Functionalized `__init__.py` (which is kind of pointless but I did it while debugging an error I was getting) and fixed a typo in it (one of the tuples was a list instead).
- Fix import of HttpError in gmail.py (which for some reason your pytest workflow was not catching).
- Updated packages that were needed to get pytest to pass or were missing patch-level updates.
- Restructured limited dependencies mode to use the same dependency versions as the usual install.
- Resolve the TypeError raised when running pytest under Python 3.11. Required [upstream fix](https://github.com/petl-developers/petl/pull/656).
- Flake8 was set to maintain a 100 character max line length, but Black didn't have a line length option configured which meant it was defaulting to a much lower 88. These are now both 100.
- Add flake8, pytest, and black configurations to pyproject.toml rather than having a mix of separate config files and command-line args -- also standardized the commands used in all places.
- Updated Black and Flake8 to latest versions and applied their updated linting suggestions.
- Pytest now finishes with 3 warnings instead of 13.

## Notes / Considerations

- Removed support for python 3.7 as it is not supported with uv. This could be added back as a supported version along with its own test workflow or no test workflow, if preferred. I know dropping 3.7 was recently discussed in #932.
- Closes #1007 by using a version of braintree that has not been yanked


requirements.txt
```diff
@@ -1,49 +1,48 @@
...
+braintree==4.26.0
+bs4==0.0.1
+censusgeocode==0.4.3.post1
+civis==1.16.1
+dbt_redshift==1.4.0
+docutils<0.18,>=0.14
 facebook-business==13.0.0
 google-api-python-client==1.7.7
+google-auth==2.6.2
+google-cloud-bigquery==3.18.0
```
bmos, line 14: updated from 3.4.0

requirements.txt
```diff
@@ -1,49 +1,48 @@
...
-slackclient==1.3.0
-psycopg2-binary==2.9.3
-xmltodict==0.11.0
-gspread==3.7.0
-oauth2client==4.1.3
-google-auth==2.6.2
+braintree==4.26.0
+bs4==0.0.1
+censusgeocode==0.4.3.post1
+civis==1.16.1
```
bmos, line 8: updated from 1.16.0

requirements.txt
```diff
@@ -1,49 +1,48 @@
...
+google-cloud-storage-transfer==1.9.1
+google-cloud-storage==2.10.0
 google-resumable-media==2.3.2
 grpcio==1.53.2
-httplib2==0.19.0
-validate-email==1.3
-paramiko==3.4.0
-xmltodict==0.11.0
+gspread==3.7.0
+httplib2==0.22.0
```
bmos, line 20: updated from 0.19.0

requirements.txt
```diff
@@ -1,49 +1,48 @@
...
-suds-py3==1.3.4.0
-newmode==0.1.6
 mysql-connector-python==8.0.18
-braintree==4.0.0
-python-dateutil==2.8.2
-azure-storage-blob==12.3.2
+newmode==0.1.6
+oauth2client==4.1.3
+paramiko==3.4.0
+petl==1.7.14
```
bmos, line 26: updated from 1.6.8

requirements.txt
bmos, line 20: updated httplib2 from 0.19.0

austinweisgrau:
I think this should be split into several different PRs for the different components here. Each PR should explain the motivation for the relevant changes and what is affected.

I see these are separate PR scopes:
- linting changes
- CICD changes
- package management changes
- 3.11 compatibility
- bug fixes

austinweisgrau:
Good work though, these look mostly like nice fixes. I think we'd want to discuss pros and cons of switching to uv, and I think it would be preferable to stick to the black line length default of 88 rather than switching it to the flake8 default of 100, but we should have those discussions in smaller scoped PRs.

bmos:
Thanks for those notes, I will restructure this into the suggested categories.

> I think it would be preferable to stick to the black line length default of 88 rather than switching it to the flake8 default of 100

100 is not the default for flake8, that would be 79.
I went with 100 because the flake8 [config file](https://github.com/move-coop/parsons/blob/main/.flake8) in your repo says 100 and the CONTRIBUTING.md file also [includes it](https://github.com/move-coop/parsons/blob/main/CONTRIBUTING.md?plain=1#L95) in the flake8 command. If 88 is the desired value I can change those values of 100 to 88 -- standardization was my goal, not a longer line length.

austinweisgrau:
> Thanks for those notes, I will restructure this into the suggested categories.
> 
> > I think it would be preferable to stick to the black line length default of 88 rather than switching it to the flake8 default of 100
> 
> 100 is not the default for flake8, that would be 79. I went with 100 because the flake8 [config file](https://github.com/move-coop/parsons/blob/main/.flake8) in your repo says 100 and the CONTRIBUTING.md file also [includes it](https://github.com/move-coop/parsons/blob/main/CONTRIBUTING.md?plain=1#L95) in the flake8 command. If 88 is the desired value I can change those values of 100 to 88 -- standardization was my goal, not a longer line length.

Oh word gotcha that makes sense! Yeah I think maintaining whatever the documented value is makes the most sense so that sounds right to me.

bmos:
Does the major release branch need to be brought up to main?

austinweisgrau:
It doth, yes. It will be eventually, PRs against it will look a little
messy until that happens.

PR #1015, opened 2024-03-14
move build config to pyproject.toml

bmos:
This is the new standard and it allows a lot of config to be done in this file, including configuration of dev tools like what I have suggested in #1013.

Doing this should also resolve the need for the LIMITED_DEPENDENCIES environment variable.
All packages have been grouped with an alias 'all' that installs everything but the dev requirements and tmc specific-packages.

I'm not sure about the Docker stuff though so I just made my best guess as to how to handle that.

It's also worth noting that LIMITED_DEPENDENCIES seems to have been implemented in such a way that most packages are not limited to a certain version, which means that you could end up with potentially quite different program behavior between LIMITED_DEPENDENCIES and normal install modes. This PR resolves this potential issue as all packages are declared per-module/in a single place.


bmos:
I need to recreate this to take into account the intended behaviors of [LIMITED_DEPENDENCIES](https://www.parsonsproject.org/pub/friendly-dependencies/release/1)

PR #1024, opened 2024-03-21
Return PATCH response when updating AK user

crayolakat:
Return the response after patching an AK user. This is necessary to parse the status code after calling the `update_user` function


shaunagm (approved):
Looks good to me - technically kind of a breaking change but I don't know who would be relying on this method returning None so I think it's okay to merge here instead of into the next major release.

PR #1029, opened 2024-04-02
New Feature: DBSync can upsert or append updated rows

austinweisgrau:
The DBSync method can already append new rows from a source db into a target db based on an incremental primary key.

This new method generalizes that behavior in a new feature which allows for appending or upserting rows that have an updated_at_column value in the source db greater than the value in the target db.

### DIFF
--- a/parsons/databases/db_sync.py
+++ b/parsons/databases/db_sync.py
@@ -1,5 +1,8 @@
+import datetime
 import logging
+from typing import Literal
 
+from parsons.databases.database_connector import DatabaseConnector
 from parsons.etl.table import Table
 
 logger = logging.getLogger(__name__)
@@ -30,11 +33,11 @@ class DBSync:
 
     def __init__(
         self,
-        source_db,
-        destination_db,
-        read_chunk_size=100_000,
-        write_chunk_size=None,
-        retries=0,
+        source_db: DatabaseConnector,
+        destination_db: DatabaseConnector,
+        read_chunk_size: int = 100_000,
+        write_chunk_size: int = None,
+        retries: int = 0,
     ):
 
         self.source_db = source_db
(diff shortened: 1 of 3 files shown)


parsons/databases/db_sync.py
```diff
@@ -373,3 +409,95 @@ def create_table(self, source_table, destination_table):
                 "Unable to create destination table based on source table; we will "
                 'fallback to using "copy" to create the destination.'
             )
+
+    def upsert_rows(
+        self,
+        source_table_name: str,
+        destination_table_name: str,
+        cutoff,
```
KasiaHinkson, line 417: type?

KasiaHinkson (approved):
yeah i think this looks great, super helpful, just one type missing I think.

parsons/databases/db_sync.py
austinweisgrau: Help me think through this? 

string works, numbers of any kind work, datetime works - really any type of python object that is going to be translated from a SQL query by parsons that can be compared with < , > , == 

I guess we'd want to look through all possible python objects that parsons might make from queries from any of the subclasses of parsons.databases.database_connector.DatabaseConnector

probably `Union[int, float, str, datetime.datetime, datetime.date]` would cover most or all of the cases - but I wasn't sure so I skipped it out of laziness and caution

KasiaHinkson: Yeah, that makes sense. I'd say, since it's not binding, we go with your suggestion that covers all the cases we can think of. It's at least a starting place for anyone looking into things, and if someone uses a type that isn't listed they can always submit a small PR.

austinweisgrau: Done with 91db890

austinweisgrau:
Hm linting

shaunagm:
This looks merge-able, so I'm going to go ahead and do so!

austinweisgrau:
Ah shoot I'm sorry I should have raised this earlier

I was testing this method and found some edge cases that make this
functionality fail that I found too tricky to resolve and stopped working
on it. I should have closed this PR or made a note of it

I don't remember off hand what those edge cases were.

IMO this should probably be reverted - I would consider the new behavior
insufficiently tested to be ready for merge

Sorry again for not noting that here

shaunagm:
Gotcha - I will revert. Not a big deal, feel free to open a new PR if you figure out a way around the edge cases.

PR #1030, opened 2024-04-03
Move variable definition outside try/except block

austinweisgrau:
A variable definition is used in the except block, so if the variable definition fails, the except block will raise another variable not defined exception.

### DIFF
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -1324,13 +1324,14 @@ class GoogleBigQuery(DatabaseConnector):
             raise ValueError(f"Only supports csv or json files [data_type = {data_type}]")
 
     def _load_table_from_uri(self, source_uris, destination, job_config, **load_kwargs):
+        load_job = self.client.load_table_from_uri(
+            source_uris=source_uris,
+            destination=destination,
+            job_config=job_config,
+            **load_kwargs,
+        )
+
         try:
-            load_job = self.client.load_table_from_uri(
-                source_uris=source_uris,
-                destination=destination,
-                job_config=job_config,
-                **load_kwargs,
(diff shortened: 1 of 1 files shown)


parsons/google/google_bigquery.py
```diff
@@ -1328,13 +1328,14 @@ def _validate_copy_inputs(self, if_exists: str, data_type: str):
             raise ValueError(f"Only supports csv or json files [data_type = {data_type}]")
 
     def _load_table_from_uri(self, source_uris, destination, job_config, **load_kwargs):
+        load_job = self.client.load_table_from_uri(
```
shaunagm, line 1331: My only concern with this move is whether it might be common to have errors in the `load_table_from_uri` call. It doesn't seem like it - it seems like they'd be caused by `result` but I'm not familiar with BigQuery so I want to double check before approving and merging.

austinweisgrau: It happens! The problem with how this try/except block is currently written is that when that happens, the except block tries to use `load_job` but that variable hasn't been defined yet, so a second exception is then raised on top of the first. Theoretically the call to `load_table_from_uri` could be in its own try/except block, but this PR is only trying to solve the first problem.

austinweisgrau: So for the situation where an exception is raised by the load_table_from_uri call, the try/except block as currently written makes the situation worse and obscures the actual error by raising a new one rather than catching the exception in any useful way.

shaunagm: That makes sense! This is definitely an improvement, so I'll approve and merge :)

shaunagm (approved):
LGTM

PR #1034, opened 2024-04-12
Removes bluelink connector because platform retired March 2024

elyse-weiss:
Removing the bluelink connector because the platform was retired in March of 2024.


parsons/__init__.py
```diff
@@ -93,7 +92,9 @@
     ("parsons.zoom.zoom", "Zoom"),
 ):
     try:
-        globals()[connector_name] = getattr(importlib.import_module(module_path), connector_name)
+        globals()[connector_name] = getattr(
```
austinweisgrau, line 95: I think this change here breaks the linting. potentially relevant is that a recent PR upgraded us from black 22 to black 24, which has some differences in how it lints files

austinweisgrau (changes requested):
otherwise LGTM

austinweisgrau:
Also I'd prefer if the commit message was more explicit like "Removing the bluelink connector because the platform has closed" or something

elyse-weiss:
@austinweisgrau I'm not sure how to edit a commit message, but added a description to the PR.

austinweisgrau (approved):
LGTM

PR #1035, opened 2024-04-12
Run upgraded black formatter on all files

austinweisgrau:
We recently upgraded the version of black we use from 22 to 24 with 44cacb9171660486aab60ddae59daf5ad9271a57, PR #1022. This version of black formats files a little differently, so this commit runs black on all files to make those changes so that they're in compliance.


shaunagm (approved):
LGTM

PR #1038, opened 2024-04-17
Remove obsolete argument to google storage client list_blobs()

austinweisgrau:
match_glob is not a valid argument to google.storage.Client().list_blobs()

I can't really tell when this argument was deprecated, it doesn't exist even in the pinned version of the google-cloud-storage package. Not sure how this broke.

### DIFF
--- a/parsons/google/google_cloud_storage.py
+++ b/parsons/google/google_cloud_storage.py
@@ -141,7 +141,6 @@ class GoogleCloudStorage(object):
         bucket_name,
         max_results=None,
         prefix=None,
-        match_glob=None,
         include_file_details=False,
     ):
         """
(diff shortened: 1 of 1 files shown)


austinweisgrau:
IDK if I'm right here - the argument is in the documentation https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.client.Client#google_cloud_storage_client_Client_list_blobs

But if I run
```
from google.cloud import storage

storage.Client().list_blobs('some_bucket', match_glob='something')
```
then I'm getting
```
TypeError: Client.list_blobs() got an unexpected keyword argument 'match_glob'
```

Might be some kind of packaging error on the google-cloud-storage project

Trying this with `google-cloud-storage` versions `2.8.0` and `2.10.0` which is pinned by parsons. Getting the same results

austinweisgrau:
Made an upstream issue https://github.com/googleapis/python-storage/issues/1266

austinweisgrau:
Nevermind, I got my versions confused somehow. This argument works with the pinned version of `google-cloud-storage>=2.10.0`

Issue #921, opened 2023-11-13
[Feature/Addition] Add support for BigQuery export/extract to Google Cloud Storage


austinweisgrau:
BigQuery can export a query result to a file in Google Cloud Storage. This is helpful for various workflows, e.g. running a query with a huge result. Having this feature will also lead to closer parity with the Redshift connector and its "unload" method.

## Possible Implementation

Here's the documentation on this including an example implementation in python: https://cloud.google.com/bigquery/docs/exporting-data#export-data-in-bigquery

## Priority

medium

PR #1041, opened 2024-04-23
BigQuery extract table method

austinweisgrau:
Addresses #921 

Unlike in Redshift, you can't directly export query results to GCS. You can only extract a table.


shaunagm (approved):
This looks good to me and seems fairly straightforward. A test would be nice but I trust it worked locally for you and we'd have to mock the test anyway, so it'd only be a help for a hypothetical future person debugging this. 

If you want to merge it as-is, go ahead. You can also wait for Jason to review if you want someone with BigQuery experience reviewing. Up to you!

PR #1042, opened 2024-04-24
Allow passing google auth Credentials directly to BQ connector

austinweisgrau:
This allows for using oauth2 to authenticate the connector, rather than a service account.


KasiaHinkson (approved):
nice!

PR #1044, opened 2024-04-28
OpenField Class

codygordon:


TheReFTW:
CCing: @drewnewell, @isik-kaplan

parsons/openfield/openfield.py
```diff
@@ -0,0 +1,536 @@
...
+    def parse_error(self, resp, exception_message):
+        """
+        Parse error responses from the API
+        """
+        message = f"Status {str(resp.status_code)}"
+        if exception_message:
+            message += exception_message
+
+        try:
+            json = resp.json()
```
codygordon, line 177: @sjwmoveon would like your advice on this: 

On 400 errors, some OpenField endpoints such as the bulk upsert people return a list of dicts that map to the indexes of the provided list, and will contain the columns containing bad data if there are any.

For example, a parsed response here could look like:
```[{}, {'phone1': ['The phone number entered is not valid.']}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {'phone1': ['The phone number entered is not valid.']}, {}, {}, {}]```

Right now, this just stringifies that response and adds it as a newline to the Exception message, under the status code.

In my end-consumer script which needs to use that list to identify the problem data and remove it before retrying the call, I end up needing to process the response like this:

```
if "400" in str(e):
  parsed_error_rows = ast.literal_eval(str(e).splitlines()[1].strip())
  
  for idx, row in enumerate(parsed_error_rows):
      for key in row:
          # do something with the bad data
```

Are you aware of a best practice that fits within the standards of how parsons returns errors that we could employ here to bubble both the status code and json in a way that's easier to parse for consumers?

Comment (codygordon) [reply parsons/openfield/openfield.py]: I think I figured out the optimal way to do this by extending `Exception` and appending relevant data from the response:
https://github.com/move-coop/parsons/pull/1044/files#diff-14899469f7551d91402e8a36a2d18b2ba509a00a8674092b7b7ab82eb7ef708bR11

This way, the consumer then looks like:
```
 except Exception as e:
            if e.status_code == 400:
                if e.json and isinstance(e.json, list):
                    for idx, row in enumerate(e.json):
                        for key in row:
                            # do something
```

**Comment by codygordon** [comment]
@shaunagm I probably won't be adding more to this, not sure if you want to merge it as-is so it can be added to in the future, or ask someone else to add more to it?

I can confirm these endpoints have been used in production so they are field-tested.

File: docs/openfield.rst [hunk 2]
```diff
@@ -0,0 +1,69 @@
+OpenField
```
shaunagm, line 1: This looks great but don't forget to add a link to the connector to [the sidebar](https://www.parsonsproject.org/pub/build-a-new-connector/release/6#sidebar)

parsons/openfield/openfield.py
```diff
@@ -0,0 +1,543 @@
...
+            The authorized OpenField user password.
+            Not required if ``OPENFIELD_PASSWORD`` env variable set.
+    """
+
+    _default_headers = {
+        "content-type": "application/json",
+        "accepts": "application/json",
+    }
+
+    def __init__(self, domain=None, username=None, password=None):
```
shaunagm, line 47: I'm surprised you didn't use the [API Connector class](https://github.com/move-coop/parsons/blob/main/parsons/utilities/api_connector.py) as the client here. Is that class not suitable for your purposes, or were you unaware it existed? If you chose not to use it, I'd love to get your feedback on what made it unworkable.

shaunagm:
@codygordon do you need this merged on any particular timeline? My preference would be to wait until tests are added to merge it, but if you need this merged fast we can make an exception, as the field-testing (which you've done) is the more important part.

shaunagm (changes requested):
Needs the docs link to the sidebar, plus ideally tests (see my other comment/question re: your timing - tests are not a dealbreaker if you're in a rush)

parsons/openfield/openfield.py
codygordon: Just unaware! I can refactor to use it as it seems better to have a standard in place like this now, for sure.

codygordon:
No rush, will write some tests 👍

Issue #1036, opened 2024-04-13
Feature request: Allow us to offset a set number of rows/columns in GoogleSheets append_to_sheet


anzelpwj:
## Detailed Description

Hi everyone! I'm putting in this request to allow users to set a number of offset rows and columns for `GoogleSheets.append_to_sheet`. For example, if I wanted my data block to start at C3, I'd give 2 skip rows and 2 skip columns.

## Context

I'm a volunteer with Tech for Campaigns, and I'd love to be able to use this tool to automate a process we have to download donation data from ActBlue and put it into a standardized spreadsheet that TFC uses for every campaigns. However, our sheets are organized in a fashion such that our data would need to start being added at position D6 (or a lower row, if I'm just updating donations after a given day).

## Possible Implementation

I'd suggest a method call something like:

```python
GoogleSheets.append_to_sheet(spreadsheet_id, table, worksheet=0, user_entered_value=False, skip_rows=0, skip_cols=0, **kwargs)
```

If `skip_rows` and `skip_cols` are both 0, the logic we have in lines 275-286 (https://github.com/move-coop/parsons/blob/3fc05e75d572a107f8c6b5af0609df0cc52564eb/parsons/google/google_sheets.py#L275) can proceed as normal. But if non-default values are provided, we simply drop the data in at skip_rows + 1, skip_cols + 1 in the spreadsheet.

Alternately, if it makes more sense to have something like `start_row=1, start_col=1` that would be fine as well.

## Priority

Medium - would simplify some grunt work for TFC, but it's not a hair-on-fire situation.

anzelpwj:
Actually just looked at the Pandas to_excel API (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_excel.html) and they use `startrow=0, startcol=0`, so probably best to match that functionality.

matthewkrausse:
Just a suggestion for an immediate workaround. Drop the data into a new sheet and then use a reference to that sheet like this in C3: `={DataFromParsons!A:C}`. 

But we'll look into this.

anzelpwj:
I'm happy to take on this PR and do the coding work, but would appreciate advice on how I should design the API to fit in best with the rest of the library.

Will probably want to add another parameter to the method call that lets you decide whether or not you want to include the header row.

shaunagm:
Hi @anzelpwj! Thanks for the feature request and for the offer to make a PR. 

[...]

PR #1045, opened 2024-04-28
Adding a data paste function to Google Sheets

anzelpwj:
Hi all,

Following up on issue #1036 . I tried initially modifying the `append_to_sheet` function, but it became clear that this is simply asking too much for one function, so I made a new method `paste_data_in_sheet` instead.

I don't have access to the Google Sheets sandbox, so for now `test_paste_data_in_sheet` has not been confirmed to work as expected (but I have done some testing with my own spreadsheet).

Any advice on how I've designed this/getting this code ready for merging would be much appreciated, thank you!


anzelpwj:
Why are imports failing for modules I haven't touched?

NirTatcher:
> Why are imports failing for modules I haven't touched?

What version of Python are you running?
I get the same error when I'm on an incompatible python version , make sure it matches the current python version requirements:
` python_requires=">=3.7.0,<3.12.0"`

anzelpwj:
> > Why are imports failing for modules I haven't touched?
> 
> What version of Python are you running? I get the same error when I'm on an incompatible python version , make sure it matches the current python version requirements: ` python_requires=">=3.7.0,<3.12.0"`
> 
> Hope this would be helpful!

It runs fine on my machine (Pop!_OS, Python 3.10.14). What I'm confused about is the CI/CD failure at https://github.com/move-coop/parsons/actions/runs/8869754280/job/24350778446?pr=1045

anzelpwj:
I will note that https://github.com/move-coop/parsons/pull/1044 also failed at the same step and in the same way.

parsons/google/google_sheets.py
```diff
@@ -293,6 +293,55 @@ def append_to_sheet(
...
+            startcol: int
+                Starting column position of pasted data. Counts from 0.
+        """
+        sheet = self._get_worksheet(spreadsheet_id, worksheet)
+
+        number_of_columns = len(table.columns)
+        number_of_rows = table.num_rows + 1 if header else table.num_rows
+
+        if not number_of_rows or not number_of_columns:  # No data to paste
+            return
```
shaunagm, line 323: Maybe some kind of log statement or warning that nothing is happening here? Instead of failing silently

parsons/google/google_sheets.py
```diff
@@ -293,6 +293,55 @@ def append_to_sheet(
...
+
+        data_range = (
```
shaunagm, line 325: I would add a code comment here, just for clarity's sake

parsons/google/google_sheets.py
```diff
@@ -293,6 +293,55 @@ def append_to_sheet(
         sheet.update_cells(cells, value_input_option=value_input_option)
         logger.info(f"Appended {table.num_rows} rows to worksheet.")
 
+    def paste_data_in_sheet(
+        self, spreadsheet_id, table, worksheet=0, header=True, startrow=0, startcol=0
+    ):
+        """
+        Pastes data from a Parsons table to a Google sheet. Note that this may overwrite
```
shaunagm, line 300: Might add a little here to make sure the user knows that what's special about this is it can paste to anywhere in the sheet (can even reference overwrite and append to show when you'd use this instead of those)

test/test_google/test_google_sheets.py
```diff
@@ -123,6 +123,57 @@ def test_append_user_entered_to_spreadsheet(self):
...
+        self.google_sheets.paste_data_in_sheet(
+            self.spreadsheet_id, paste_table3, worksheet=3, header=False, startrow=2, startcol=0
+        )
+        self.google_sheets.paste_data_in_sheet(
+            self.spreadsheet_id, paste_table4, worksheet=3, header=False, startrow=2, startcol=2
+        )
+
+        result_table = self.google_sheets.read_worksheet(self.spreadsheet_id, 3)
+        results_dict = result_table.to_dicts()
+
```
shaunagm, line 171: You can also assert equality of parsons table by accessing `table.data`

shaunagm (changes requested):

shaunagm (approved):
Looks great! Thanks for making these changes

PR #1047, opened 2024-05-01
Try different runner

anzelpwj:
Just want to see if using the `macos-latest-large` runner works (I think it's M1).


anzelpwj:
Nope, it doesn't

PR #1050, opened 2024-05-09
Sqlalchemy Update to  >= 1.4.22

matthewkrausse:
Will update the version to match the prefect reqs file. This just makes sure we can resolve a conflict and the prefect req seems to be the best way to handle sqlalchemy.


matthewkrausse:
@shaunagm It looks like only two tests ran and none of the different OS builds.

shaunagm:
Oh according to this, github actions (tests) don't run if there's a merge conflict: https://github.com/orgs/community/discussions/26304

anzelpwj:
I think you might also want to set a version requirement in `setup.py`.

matthewkrausse:
Deleted the wrong branch by mistake.

matthewkrausse:
@shaunagm I think this is ready to go

shaunagm (approved):
LGTM, thanks!

PR #1051, opened 2024-05-10
Psycopg2 version bump

anzelpwj:
This is a more permanent solution to our issue with the failing `macos-latest` CI test for Python 3.11. Turns out we were installing psycopg2 2.9.2, and the library does not officially support Python 3.11 until 2.9.5. I bumped it up to the latest version (2.9.9) and everything appears to work as expected, but I have not executed any of the live tests.


shaunagm (approved):
Looks good to me

PR #1055, opened 2024-05-15
Update GET helper function, add custom columns to create profile

cmdelrio:
1) Update GET helper function - There was a small bug in the logic for get endpoints where pagination is controled with a "num" page indicator. In short, line 140 was returning false when it should have been returning true! 
2) Correct phone_numbers aspect of get_profiles - turns out mobile commons accepts a comma separated string of phone numbers, not a list. 
3) Added the option to write custom column data to profiles in the create_profiles method. Also updated documentation to indicate that the create_profiles method can also be used to update existing records. Considered creating a new method all together for update_profile, but it seemed a little bit redundant...

### DIFF
--- a/parsons/mobilecommons/mobilecommons.py
+++ b/parsons/mobilecommons/mobilecommons.py
@@ -353,7 +353,8 @@ class MobileCommons:
         custom_cols = "true" if include_custom_columns else "false"
         subscriptions = "true" if include_subscriptions else "false"
 
-        phones = ",".join(phones)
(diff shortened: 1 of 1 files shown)


jeffjrodriguez (approved):
LGTM

shaunagm (approved):
This looks straightforward, approving

cmdelrio:
Thank you @jeffjrodriguez and @shaunagm !

PR #1062, opened 2024-05-23
If petl is loaded from a CSV, use the source file, don't re-write

austinweisgrau:
No need to rewrite a CSV for a non-transformed petl that was loaded from a CSV on disk.

This can save a lot of time when working with large CSVs

### DIFF
--- a/parsons/google/google_cloud_storage.py
+++ b/parsons/google/google_cloud_storage.py
@@ -4,10 +4,11 @@ from google.cloud import storage_transfer
 from parsons.google.utitities import setup_google_application_credentials
 from parsons.utilities import files
 import datetime
+import gzip
+import petl
 import logging
 import time
 import uuid
-import gzip
 import zipfile
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


parsons/google/google_cloud_storage.py
```diff
@@ -311,7 +312,10 @@ def upload_table(self, table, bucket_name, blob_name, data_type="csv", default_a
         blob = storage.Blob(blob_name, bucket)
 
         if data_type == "csv":
-            local_file = table.to_csv()
+            if isinstance(table.table, petl.io.csv_py3.CSVView):
```
shaunagm, line 315: Just want to make sure I understand this (and may make sense to add this as a comment, so future people can understand this) - does the check that `table.table` is instance of `CSVView` mean that `table.table` has not been transformed? Otherwise how would we make sure the table hasn't been altered and we're not using an out-of-date file?

shaunagm (commented):
Happy to approve this, just want to get a little clarification on how this works (and possibly have that clarification as a code comment, if you agree that it would be helpful)

parsons/google/google_cloud_storage.py
austinweisgrau: Yes that's right! Once you do any transformation on a parsons Table, the Table.table petl object will no longer be a CSVView. I'll add that as a comment

shaunagm (approved):
Thanks for the quick clarification and for adding the comment! I'll merge once the tests pass again.

Issue #1059, opened 2024-05-21
[Feature] Add optional parameter to disaggregate A/B tests in NGP email code


anzelpwj:
## Detailed Description

I'd like to add an optional parameter to `ngpvan.email` code that allows you to avoid aggregating A/B test results in the `get_email_stats`  method.

## Possible Implementation

New method call: `get_email_stats(aggregate_ab=True)`, and if aggregate_ab is false then the try block does not aggregate.

## Priority
Medium. I can do this.

PR #1065, opened 2024-05-28
Add option to disaggregate A/B test emails

anzelpwj:
Addresses #1059 - this gives us an optional parameter to the NGP email endpoint that lets us dis-aggregate A/B test results from the endpoint. I also did a bit of code cleanup and added some tests.

### DIFF
--- a/parsons/ngpvan/email.py
+++ b/parsons/ngpvan/email.py
@@ -70,13 +70,13 @@ class Email(object):
         logger.debug(f"Found email {email_id}.")
         return r
 
-    def get_email_stats(self) -> Table:
+    def get_email_stats(self, aggregate_ab: bool = True) -> Table:
         """
         Get stats for all emails, aggregating any A/B tests.
 
         `Args:`
-            emails : list
-                A list of email message details.
+            aggregate_ab : bool
+                If A/B test results for emails should get aggregated.
 
         `Returns:`
             Parsons Table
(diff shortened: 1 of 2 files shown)


anzelpwj:
Tagging @shaunagm and @matthewkrausse as interested parties.

parsons/ngpvan/email.py
```diff
@@ -70,13 +70,13 @@ def get_email(self, email_id: int, expand: bool = True) -> Table:
...
 
-    def get_email_stats(self) -> Table:
+    def get_email_stats(self, aggregate_ab: bool = True) -> Table:
         """
         Get stats for all emails, aggregating any A/B tests.
 
         `Args:`
-            emails : list
-                A list of email message details.
+            aggregate_ab : bool
```
anzelpwj, line 78: We were missing the emails parameter - is this something we wanted?

parsons/ngpvan/email.py
```diff
@@ -95,50 +95,48 @@ def get_email_stats(self) -> Table:
...
-                    ]
-            except TypeError as e:
-                logger.info(str(e))
-                pass
-
-            final_email_list.append(d)
+                    try:
+                        for field in inner_fields:
+                            inner[field] += i["emailMessageContentDistributions"][field]
+                    except KeyError as e:
```
anzelpwj, line 126: Error should be KeyError, not TypeError. Found this out in testing.

test/test_van/test_email.py
```diff
@@ -81,6 +122,15 @@ def assert_matching_tables(table1, table2, ignore_headers=False):
     },
 ]
 
+mock_response_enriched = deepcopy(mock_response)
```
anzelpwj, line 125: This enriched table includes the output you would get with `expand=True`. Last entry gets the A/B test.

test/test_van/test_email.py
```diff
@@ -102,10 +152,33 @@ def test_get_email_messages(self, m):
     def test_get_email_message(self, m):
         m.get(
             self.van.connection.uri + "email/message/1",
-            json=mock_response[0],
+            json=mock_response_enriched[0],
```
anzelpwj, line 155: Should be enriched here, because we call `get_email` with the default `expand` kwarg (`= True`). We could potentially add a test if we want `expand=False`, but I don't know if it's worth the effort.

test/test_van/test_email.py
```diff
@@ -102,10 +152,33 @@ def test_get_email_messages(self, m):
...
+            m.get(
+                self.van.connection.uri + f"email/message/{resp['foreignMessageId']}",
+                json=resp,
+                status_code=200,
+            )
+
+        response_t = self.van.get_email_stats(aggregate_ab=True)
+        response_f = self.van.get_email_stats(aggregate_ab=False)
+
+        assert len(response_t.to_dicts()) == 5
```
anzelpwj, line 180: Did not feel like hand-writing out the whole expected table here, so instead I just tested to see if we should get one more entry if we don't aggregate and ensure that we either aggregate or dis-aggregate on `recipientCount`.

test/test_van/test_email.py
anzelpwj: NVM, decided it was easy enough to add.

test/test_van/test_email.py
shaunagm: In our pair programming session today, @matthewkrausse pointed out that this sample/fake data may not correspond to the actual data. Do you have actual data to compare it too? If not, we can figure something out.

parsons/ngpvan/email.py
```diff
@@ -95,50 +95,48 @@ def get_email_stats(self) -> Table:
...
+            "bounceCount",
+            "contributionCount",
+            "contributionTotal",
+            "formSubmissionCount",
+            "linksClickedCount",
+            "machineOpenCount",
+            "openCount",
+            "unsubscribeCount",
+        ]
+        if aggregate_ab:
```
shaunagm, line 118: Can you add some comments to this code, so I can better understand what it's doing? It's hard for me to parse as-is

shaunagm (changes requested):

test/test_van/test_email.py
```diff
@@ -81,6 +189,15 @@ def assert_matching_tables(table1, table2, ignore_headers=False):
...
 ]
 
+mock_response_enriched = deepcopy(mock_response)
+mock_response_enriched[0]["emailMessageContent"] = sample_content_single
+mock_response_enriched[1]["emailMessageContent"] = sample_content_single
+mock_response_enriched[2]["emailMessageContent"] = sample_content_single
+mock_response_enriched[3]["emailMessageContent"] = sample_content_single
+mock_response_enriched[4]["emailMessageContent"] = sample_content_full
+
+print(mock_response_enriched)
```
shaunagm, line 199: Is there a reason we're printing something here, or is it left over from when you were debugging?

shaunagm (approved):
Can you confirm that you checked the data structures with @matthewkrausse? If yes, and if the print statement is meant to be there, I'll go ahead and merge.

test/test_van/test_email.py
anzelpwj: That was probably a holdover from testing things in a notebook. Will delete. The data structures I got directly from @matthewkrausse , just with a bit of field value changing (so it didn't have GGT data in it).

shaunagm (approved):

PR #1067, opened 2024-05-29
Hot fix for MobileCommons.create_profile() method

cmdelrio:
Made a silly mistake last round of updates and used dict.merge() when I should have used dict.update()


anonarai (approved):
Looks good!

PR #1068, opened 2024-05-30
Remove support for dict type, add helpful exception message

austinweisgrau:
Suggestion for how to resolve #1028


KasiaHinkson (approved):
I'm sorry this took so long, was there somewhere we could add documentation for our suggestion to convert dicts to json strings before this step?

austinweisgrau:
I was hoping the exception message would be explicit enough for that, super down for more language to be added though!

```
                    "Column type not supported for load to BigQuery. "
                    "Consider converting to another type. "
                    f"[type={best_type}]"
```

PR #1069, opened 2024-06-03
Enable support for python 3.12

austinweisgrau:
Addresses #1021

PR #1070, opened 2024-06-03
Support for python3.12

jdw25:
This incorporates 
- library bumps for installation
- introduction of Setuptools to requirements.txt (due to PEP 632 & gh-95299)
- updating workflows and setup.py to reflect 3.12 support
- some updates to test suite to handle erros due to syntax changes with 3.12 (in progress, a little more to go here with test_twilio.py; also then will need to reflect similar syntax updates with test_salesforce.py)


austinweisgrau:
Addresses #1069

anzelpwj:
```
 pytest introspection follows:

Kwargs:
assert {} == {'time_period': 'today'}
 ```
 
 IIRC we did discuss that the `time_period` value did not get passed in to the call. I think we can get rid of the ` {'time_period': 'today'}` is the test.

austinweisgrau:
> ```
>  pytest introspection follows:
> 
> Kwargs:
> assert {} == {'time_period': 'today'}
>  ```
>  
>  IIRC we did discuss that the `time_period` value did not get passed in to the call. I think we can get rid of the ` {'time_period': 'today'}` is the test.

See #1072 for a discussion of the underlying issue here. Unfortunately there are several misspecified tests facing the same issue that will need to be rewritten not just to pass but also to test what they are trying to test.

shaunagm:
Minor but important note: you'll also want to update the README.md file to say we support up to 3.12. (I was going to say we need to update the little blue button too but apparently it's dynamically generated from the versions listed on PyPI - not sure how that gets specified but we can figure that out separately)

shaunagm:
Thanks for updating the readme. Also, I looked into it and the blue buttons are getting dynamically generated from the classifiers section of setup.py, which you've already updated in this PR, so we should be all good there.

test/test_salesforce/test_salesforce.py
austinweisgrau, line 1: I'd personally really rather we fixed these broken tests rather than just turning them off, but I supposed since they already are basically not running, and we have an issue documenting the problem (#1072) it's not a big problem

anzelpwj:
Add to `requirements_dev.txt` `pycodestyle==2.12.0`

austinweisgrau:
Should be able to remove the test changes from this branch in favor of the more thorough fix in #1072

bmos:
I wonder why docs build is failing to checkout the PR...
https://app.circleci.com/pipelines/github/move-coop/parsons/5443/workflows/01bf4ad9-0f62-48c8-bd31-8720a0505e2d/jobs/7030

jdw25:
@bmos yeah, idk maybe! 

Based on convo in group pairing meeting yesterday, will retract this PR and create a new one with the cleaner commit sequence (while ending up with same intended changes) and based on the now-merged [1072](https://github.com/move-coop/parsons/pull/1072) (thanks @austinweisgrau!) that should hopefully create a more straightforward path.

Issue #1071, opened 2024-06-03
[Bug] Tests and test documention use nonexistent method "MagicMock().called_with()` leading to erroneous test passes


austinweisgrau:
`called_with` is not a real test method on MagicMock, so it returns a MagicMock instance by default, which is truthy and evaluates to true.

Tests using this method `assert MagicMock().called_with()` are essentially running `assert MagicMock()` which is the same as `assert True`. These tests aren't checking for anything and pass no matter what actually happens in the code.

The documentation on how to write tests incorrectly advocates using this syntax:
```
docs/write_tests.rst
136:    assert self.sf._client.query_all.called_with('FAKESOQL')
```

PR #1072, opened 2024-06-03
Fix misspecified tests using MagicMock().called_with()

austinweisgrau:
See issue #1071 

The tests which use this method must be rewritten - some of them fail once the misspecified method is fixed, and will need to be rewritten to actually test what they intend to test.

TO DO:
- [x] Fix remaining twilio tests
- [x] Fix salesforce tests
- [x] Fix documentation recommending use of `called_with()`

BLOCKING: 
- [x] In python 3.12 it's no longer syntactical to use MagicMock().called_with(), so these fixes must be made before the python 3.12 migration can be made (#1070)


shaunagm (approved):
This looks good to me - thanks for catching & fixing these!

PR #1073, opened 2024-06-06
Fix VAN create_event location_ids list bug

codygordon:
Looks like this was defined as a list within a tuple, and events were not being associated with provided locations. Confirmed provided locations are properly associated with created events with this change.


austinweisgrau (approved):
nice catch

PR #1074, opened 2024-06-07
Add methods to update phone, order user detail, and import action

technicalex:
This pull request:
- Adds methods to update phone, order user detail, and import action in ActionKit
- Adds corresponding unit tests
- Fixes code comments/documentation for the `update_user()` function


codygordon (changes requested):
Looks good @technicalex! 

Just one suggestion: since these seem to be more esoteric endpoints that aren't well-documented by AK, it would be helpful to add a bit more to the docstring descriptions about what they do exactly and how they might be used.

technicalex:
Thanks for the suggestion, @codygordon! I improved the documentation.

codygordon (approved):

PR #1077, opened 2024-06-13
Airmeet Event Details connector with tests and documentation

strangeways:
This connector adds support for [Airmeet](https://www.airmeet.com/), a webinar platform.

Methods are included for everything listed in the [Event Details API](https://help.airmeet.com/support/solutions/articles/82000909768-1-event-details-airmeet-public-api) documentation. However, we don't currently have data in our account for booths, event tracks, or poll responses, so I haven't verified the data being returned from those.

This connector doesn't implement the [Manage Registrations](https://help.airmeet.com/support/solutions/articles/82000909769-2-manage-registrations-airmeet-public-api) or [Manage Events](https://help.airmeet.com/support/solutions/articles/82000909770-3-manage-event-airmeet-public-api) API methods.

Basic tests and documentation are included.

### DIFF
--- a/parsons/airmeet/airmeet.py
+++ b/parsons/airmeet/airmeet.py
@@ -219,8 +219,8 @@ class Airmeet(object):
         Get all attendees for an Airmeet, handling pagination based on the
         returned pageCount.
 
-        Results are available only for events with a status of `FINISHED`.
-        Maximum number of results per page = 50.
+        Results include attendance only from sessions with a status of
+        `FINISHED`. Maximum number of results per page = 50.
 
         "This is an Asynchronous API. If you get a 202 code in response,
         please try again after 5 minutes."
@@ -266,7 +266,7 @@ class Airmeet(object):
                 List of booths for this Airmeet
         """
         response = self.client.get_request(url=f"airmeet/{airmeet_id}/booths")
-        return Table(response["booths"])
+        return Table(response["booths"] or [])
 
     def fetch_booth_attendance(self, airmeet_id, booth_id) -> Table:
         """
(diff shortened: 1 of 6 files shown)


codygordon:
@shaunagm since this is a new class, similar to the OpenField connector, do you have any flags? Unlike my PR, looks like Ankur used the APIConnector which is good!

codygordon (approved):

parsons/airmeet/airmeet.py
```diff
@@ -0,0 +1,415 @@
...
+
+        `Args:`
+            airmeet_id: string
+                The id of the Airmeet.
+        `Returns:`
+            Parsons.Table
+                List of booths for this Airmeet
+        """
+        response = self.client.get_request(url=f"airmeet/{airmeet_id}/booths")
+        return Table(response["booths"] or [])
```
shaunagm, line 269: I'm curious about this syntax - does `response["booths"]` return something weird like None?

shaunagm:
Overall this looks good. I am a little concerned about including completely untested code, especially given that this particular API seems to be inconsistent at times in what it returns. Is there any way to add test data so you can double-check?

parsons/airmeet/airmeet.py
strangeways: @shaunagm Yes, if there are no booths for the Airmeet, the API returns `{"booths":null}`

shaunagm (changes requested):
I'm switching my status to "request changes" because it's my understanding that you're going to test out the last few endpoints in your production environment. Let me know once that's done and I'll approve + merge - or let me know if we need a different approach.

shaunagm:
@strangeways sorry about the failing tests, they're an unrelated issue we're trying to debug and shouldn't block merging (as long as the tests otherwise pass, which they do).

I appreciate the warnings in the docstrings. IIRC that was the only issue that needed addressed before merging, so I will go ahead and merge this

PR #1080, opened 2024-06-19
Skip empty tables in GSheets append & overwrite methods

codygordon:
If a table with no data is provided the overwrite method an error that is hard to debug is raised:
`ValueError: min() arg is an empty sequence`. We ought to just log a warning.

The [`paste_data_in_sheet` method already had a check like this](https://github.com/MoveOnOrg/parsons/blob/b02a67a34b3f9235dd2dafa89251b46dcb544ddb/parsons/google/google_sheets.py#L330).

### DIFF
--- a/parsons/google/google_sheets.py
+++ b/parsons/google/google_sheets.py
@@ -375,10 +375,6 @@ class GoogleSheets:
                 Otherwise, values will be entered as strings or numbers only.
         """
 
-        if not len(table.columns) and not table.num_rows:
-            logger.warning("No data provided to overwrite sheet, skipping.")
-            return
-
         # This is in here to ensure backwards compatibility with previous versions of Parsons.
         if "sheet_index" in kwargs:
             worksheet = kwargs["sheet_index"]
@@ -387,6 +383,10 @@ class GoogleSheets:
         sheet = self._get_worksheet(spreadsheet_id, worksheet)
         sheet.clear()
 
+        if not len(table.columns):
+            logger.warning("No data provided, worksheet is empty.")
+            return
+
         value_input_option = "RAW"
         if user_entered_value:
             value_input_option = "USER_ENTERED"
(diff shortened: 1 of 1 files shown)


parsons/google/google_sheets.py
```diff
@@ -370,6 +375,10 @@ def overwrite_sheet(
                 Otherwise, values will be entered as strings or numbers only.
         """
 
+        if not len(table.columns) and not table.num_rows:
+            logger.warning("No data provided to overwrite sheet, skipping.")
+            return
+
```
austinweisgrau, line 381: I don't think this is appropriate, if no data is returned that may be a valid result and should be allowed to overwrite a sheet.

parsons/google/google_sheets.py
```diff
@@ -262,6 +262,10 @@ def append_to_sheet(
                 Otherwise, values will be entered as strings or numbers only.
         """
 
+        if not table.num_rows:
+            logger.warning("No data provided to append, skipping.")
+            return
+
```
austinweisgrau, line 268: This looks good

parsons/google/google_sheets.py
austinweisgrau: I guess that is weird if the table has no columns, no idea what circumstances that would come up.

codygordon: Thanks for reviewing! Your comment made me look a bit closer at this and I realized that the error I mentioned is happening _after_ the columns are written anyway (which, you're right, the table should basically alway have) when the `cells` list passed to `update_cells` is empty.

I just changed this so:

- The sheet is always cleared.
- If there are no columns a warning is logged and it stops. Obviously this would almost never happen but no harm in handling it.
- If there are no rows a warning is logged and it finishes normally.

austinweisgrau (approved):

PR #1081, opened 2024-06-21
SFTP utility can include timeout

austinweisgrau:
This change allows setting a timeout on the SFTP connection. If set, a TimeoutExcept will be raised if the SFTP server does not respond for the number of seconds set by the timeout.


shaunagm (approved):
LGTM

Issue #950, opened 2023-12-07
Batch upsert method for Airtable class


lkesich:
## Detailed Description
It would help me if there was a Parsons method in the Airtable class that creates a [performUpsert request](https://airtable.com/developers/web/api/update-multiple-records). For my purposes, it's important to be able to supply arguments to the fieldsToMergeOn array. This would reduce the number of calls needed to achieve the same effect with existing update_record and insert_record methods.

## Priority
This is a low priority for me--I will use it when it's complete, but in the meantime I'll use another Python wrapper for the airtable API.

lkesich:
Thank you! The batch_upsert method from pyAirtable is exactly what I'm looking for.

Issue #758, opened 2022-10-13
Airtable get_records does not return columns when all values are null


mattkrausse:
The get_records method only returns a parsons table with columns that are not null for all records. When pulling a view with just a few records, I am not getting certain columns that happen to be blank. I imagine that’s just how the api returns the data but kind of annoying.

## Detailed Description
If you create a view with some records and all of them are missing a value for column "status", then status will not appear in the Parsons Table that is returned.

## To Reproduce
Call the get_records method on a table/view with a certain column that have all null values.

## Your Environment

* Version of Parsons used (if you know): 0.20.1
* Environment name and version (e.g. Chrome 39, node.js 5.4): Pycharm CE 2021, Python 3.9
* Operating System and version (desktop or mobile): MacOSX 12.5.1 Desktop

## Priority
low

codygordon:
It looks like the developer intends this behavior, or at least doesn't mind it.
https://github.com/gtalarico/pyairtable/issues/229#issuecomment-1529161449

I kind of see their point but I do agree default behavior should be that the `get_records` method always returns all columns in the table/selected view, even if they are empty like you'd get from a `select *` query result. 

To achieve this on our end would require fetching the table schema to get all field names and and explicitly appending the fields on the call every time, if specific fields hadn't already been defined for the call.

@shaunagm or @austinweisgrau how do you feel about implementing this "opinionated override" that adds some complexity to the method and tests?

shaunagm:
I'm for it, as long as it's well documented - to clarify, this would be an optional override that users could trigger via a parameter to the get_records method?

shaunagm:
I'm for it, if you think it'll make parsons easier to use. To clarify, this would be an optional parameter to `get_records` that users could pass in to get this behavior?

PR #1082, opened 2024-06-24
Airtable Updates

codygordon:
- [`airtable-python-wrapper` was renamed to `pyairtable` after v1.0](https://github.com/gtalarico/pyairtable/issues/125#issuecomment-891439661). Needed to update this eventually if we want to take advantage of newer features. Seems despite a couple method name changes that it's backwards compatible.
- Added new methods:
  - [update_records](https://pyairtable.readthedocs.io/en/stable/api.html#pyairtable.Table.batch_update)
  - [upsert_records](https://pyairtable.readthedocs.io/en/stable/api.html#pyairtable.Table.batch_upsert)
  - [delete_record](https://pyairtable.readthedocs.io/en/stable/api.html#pyairtable.Table.delete)
  - [delete_records](https://pyairtable.readthedocs.io/en/stable/api.html#pyairtable.Table.batch_delete) 
- Added `typecast` to `insert_record`


codygordon:
https://github.com/move-coop/parsons/issues/950

codygordon:
also working on addressing [this](https://github.com/move-coop/parsons/issues/758), which I can confirm is still an issue even in the updated lib.

the fix requires fetching the table schema and appending all `fields` to the call, which makes the tests a bit trickier

parsons/airtable/airtable.py
```diff
@@ -159,28 +160,148 @@ def insert_records(self, table, typecast=False):
...
+        `Args:`
+            table: A Parsons Table or list containing the record `id`s to delete.
+        `Returns:`
+            List of dicts with record `id` and `deleted` status
+        """
+
+        if isinstance(table, Table):
+            table = table.to_dicts()
+
+        if any(isinstance(row, dict) for row in table):
```
shaunagm, line 293: Can you add a comment here to clarify what's happening on these two lines?

shaunagm (changes requested):
This looks great, just one place that would be good to have a code comment as it's not clear what the code's doing. Once that's added I can merge this

shaunagm (approved):

PR #1083, opened 2024-06-24
Bug fix: pass custom delimiter from Table.from_csv to BigQuery load

austinweisgrau:
A bug was introduced by recent PR #1062 

If our source table is loaded from CSV with no transformations, the original source file will be directly loaded to GCS.

We may need to pass along a custom delimiter to BigQuery.

e.g. for this workflow:
```
tbl = Table.from_csv(filepath, delimiter='\t')
BigQuery().copy(tbl, 'my_dataset.my_table')
```
The copy job will fail because the delimiter for the file loaded to GCS will be a `\t`, but BigQuery().copy() expects a comma by default.

This change fixes that to pass along any custom delimiter set on the Table file load to the BigQuery copy job config.

### DIFF
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -787,6 +787,15 @@ class GoogleBigQuery(DatabaseConnector):
 
         self._validate_copy_inputs(if_exists=if_exists, data_type=data_type)
 
+        # If our source table is loaded from CSV with no transformations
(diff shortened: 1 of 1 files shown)


shaunagm (approved):
This looks good to me, although I'm not a BigQuery user so my ability to review here is limited. (Or delimited? 🥁)

PR #1089, opened 2024-07-09
Add campaignId parameter to apply_responses method in People class

matthewkrausse:
This pull request adds a new parameter, campaignId, to the apply_responses method in the People class. 

This handles issue #1075


matthewkrausse:
@shaunagm This is ready to go.

shaunagm (changes requested):
This looks good but can you add the campaign_id parameter to the docstring for the method? Once you do I can merge

shaunagm (approved):

retacg:
@shaunagm  I am confused why these changes don't show up when I look at the docs - https://move-coop.github.io/parsons/html/stable/_modules/parsons/ngpvan/people.html#People.update_person 

Is that because there hasnt been a formal "release"? But if I were to use the "latest" version of parsons in a docker image I would have access? Just a bit confused!

shaunagm:
Yes, the docs default to whatever the most recent release on PyPI is - not what's merged into main on Github. 

You can actually see the docs for the most recent version of Parsons on Github (ie not yet released to PyPi) if you look in the bottom left hand corner of the docs page. There's a version picker, and you can pick 'latest'. Here's a [direct link](https://move-coop.github.io/parsons/html/latest/ngpvan.html) to the 'latest' version of the NGPVAN.

If you build your docker image by writing `pip install parsons` that will get the most recent version on PyPI, not whatever's on main. But you can override that default and install directly from Github. [this link](https://stackoverflow.com/a/15268990) has instructions for how to do so using pip. 

Relatedly, @sharinetmc will be making a new Parsons release early next week - we don't usually let Github and the PyPi version get so out of sync.

PR #1090, opened 2024-07-10
Python3.8 Base Image

IanRFerguson:
The Parsons image officially broke ... this PR swaps out the base image in the Dockerfile and officially deprecates 3.7 as a supported version (not sure the latter bit here is necessary).

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=linux/amd64 python:3.7
+FROM --platform=linux/amd64 python:3.8
 
 ####################
 ## Selenium setup ##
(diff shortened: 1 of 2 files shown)


KasiaHinkson (approved):
Approved! I'd like to move this up to 3.11 in the next couple months, so we're really set, but that's a longer process.

shaunagm:
@KasiaHinkson Python 3.8 will [no longer be officially supported](https://devguide.python.org/versions/) as of October 14th of this year. Ideally we'd remove Parsons support around the same time (happy to wait until November/December though).

KasiaHinkson:
@shaunagm yep, I'm gonna push pretty hard to get TMC up to at least 3.9 if not 3.11 before September

PR #1092, opened 2024-07-11
Enhance and fix Zoom methods

austinweisgrau:
This PR makes two changes to the Zoom connector.
- get_meetings() takes new optional arguments to date filter results
- the post processing on the `get_past_meeting_poll_metadata` method is fixed

### DIFF
--- a/parsons/zoom/zoom.py
+++ b/parsons/zoom/zoom.py
@@ -1,5 +1,7 @@
+import datetime
 import logging
 import uuid
+from typing import Dict, Literal, Optional
 
 from parsons import Table
 from parsons.utilities import check_env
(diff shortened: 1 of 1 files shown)


IanRFerguson (approved):
Dig it!

parsons/zoom/zoom.py
```diff
@@ -158,7 +160,15 @@ def get_users(self, status="active", role_id=None):
...
         return tbl
 
-    def get_meetings(self, user_id, meeting_type="scheduled"):
+    def get_meetings(
+        self,
+        user_id,
+        meeting_type: Literal[
+            "scheduled", "live", "upcoming", "upcoming_meetings", "previous_meetings"
+        ] = "scheduled",
+        from_date: Optional[datetime.date] = None,
```
shaunagm, line 169: Can you add the two new input params, from_date and to_date, and the fact that they're datetime objects. to the docstring?

shaunagm (changes requested):
One minor docstring update needed, plus the tests will need to pass before we can merge

austinweisgrau:
Weirdly I'm not able to reproduce this test_gmail.py error locally but will dig in when I have a moment

shaunagm:
Oh interesting, we were getting this in #1132 and assumed it was from a change in that PR (cc @bmos)

parsons/zoom/zoom.py
austinweisgrau: Done with 9d988c9991af83a923951448bc602cb0be7161ef

shaunagm (approved):

shaunagm:
This looks ready to merge to me! Is it good on your end?

PR #1095, opened 2024-07-15
Refactor dbt utility for modularity

austinweisgrau:
These changes uncouple the code for the dbt utility that runs a dbt command from the code that logs dbt commands. This allows for modular use of different loggers and a very flexible ability to customize logging. This addresses the main concerns from the original PR (#841).

# Breaking changes
This PR does cause breaking changes with the previous implementation of the dbt utility, but those changes are well worth it for a much cleaner and clearer interface.

Using the new implementation looks like:
```{python}
import os
from pathlib import Path
from parsons.utilities.dbt import run_dbt_commands, dbtLoggerSlack, dbtLoggerPython

results = run_dbt_commands(
    commands=["dbt run", "dbt test"],
    dbt_project_directory=Path("/path/to/dbt/project"),
    loggers=[dbtLoggerPython, dbtLoggerSlack(webhook=os.environ["SLACK_WEBHOOK"])],
)
```

The prior implementation was potentially only compatible with Redshift, depending on how the user's dbt `profiles.yaml` was set up. The new implementation allows the dbt process to inherit the shell environment from the python parent process, and is therefore compatible with running dbt on any database, and the user can configure credential passing using environment variables (recommended) or any other method.

# Improved documentation
There is now much more thorough documentation throughout the 4 modules containing the dbt utility code.

TO DO:
- [x] Touch up the dbtLoggerMarkdown text formatting a bit
- [x] dbtLoggerDatabase splits artifacts into a run table and a node table


elyse-weiss:
I do not feel qualified to approve, but having the `requirements.txt` file updated as it is here, would be helpful for me separately!

austinweisgrau:
Finally fixed the build issue

parsons/utilities/dbt/__init__.py
```diff
@@ -0,0 +1,60 @@
+"""Utility for running and logging output from dbt commands
+
```
shaunagm, line 2: These docs are good, but ideally we want to be able to surface them on the docs website. Maybe add a hook into the autodocs on this page? https://move-coop.github.io/parsons/html/stable/utilities.html#

parsons/utilities/dbt/dbt.py
```diff
@@ -0,0 +1,126 @@
...
+        for command in self.commands:
+            result = self.execute_dbt_command(command)
+            results.append(result)
+
+        return results
+
+    def execute_dbt_command(self, command: str) -> Manifest:
+        """Runs dbt command and logs results after process is completed.
+
+        If raise_error is set, this method will raise an error if the dbt
```
shaunagm, line 50: Where is raise_error set? Where is it referenced? Is this an underlying dbt thing or is it part of this method/class?

shaunagm (changes requested):
A couple minor requests/questions but overall looks good. I know @matthewkrausse is looking forward to seeing this merged so hope to get it over the finish line soon and promise not to be a blocker on my end :)

parsons/utilities/dbt/__init__.py
austinweisgrau: I thiiink 8d56b8a5da5172597ab6b162cf34d0ec0f63a473 should do this

PR #1098, opened 2024-07-16
Move redshift and sshtunnel imports into ActionNetwork method

austinweisgrau:
Avoid making redshift and sshtunnel dependencies for all users of the ActionNetwork connector


austinweisgrau:
Ah shoot, that breaks the way the patch is set up for the test to mock the Redshift and SSHTunnel.

austinweisgrau:
Closing in favor of reverting #1025 and fixing issues there

PR #1099, opened 2024-07-16
Remove sshtunnel from requirements-dev.txt

austinweisgrau:
This belongs in requirements.txt and also in setup.py, but not here.


austinweisgrau:
Closing because we are reverting #1025 and fixing the issues there

PR #1101, opened 2024-07-16
adding unique id lists support in ActionNetwork module

NirTatcher:
- I've added only 2 functions: get_unique_id_lists, get_unique_id_list to fully support ActionNetwork's API.
@shaunagm @austinweisgrau I've separated the old problematic PR into 2 PR's so this is the first non problematic one and I would soon open the PR with the SSH utility PR.
Hope that is ok thank you and sorry for the hassle.


parsons/google/google_cloud_storage.py
austinweisgrau, line 1: Can you remove the changes on the google connectors? It looks like just linting but it's out of scope and wanna avoid any unexpected changes there

NirTatcher:
@austinweisgrau I restored the google_cloud_storage.py and the google_sheets.py files from main

austinweisgrau (approved):

austinweisgrau:
Looks like there are still 2 changes in the google connector files that are breaking linting

NirTatcher:
@austinweisgrau Yep I forgot the newline at the end of files for the google_sheets and google_cloud_storage files.

PR #1102, opened 2024-07-16
ActionNetwork SQL Mirror querying support + general SSH util to query any db through ssh

NirTatcher:
- Added SSH util function we can connect through SSH and query through, added a test for that.
- Added SQL Mirror support for ActionNetwork which lets us query our data/connect to our data we have in ActionNetwork (all tables read only access). 
@austinweisgrau, @shaunagm  I hope this won't break anything like my other PR, I recreated the utility and made it from a class to just a function we can use to query a DB through a SSH tunnel connection.
Please let me know if there is any feedback or any change request :)
Also, thank you again @austinweisgrau and @shaunagm for everything!


parsons/utilities/ssh_utilities.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+        logging.info("Database connection established successfully.")
+
+        cursor = con.cursor()
+        cursor.execute(query)
+        records = cursor.fetchall()
+        output = records
+        logging.info(f"Query executed successfully: {records}")
+    except Exception as e:
+        logging.error(f"Error during query execution: {e}")
```
austinweisgrau, line 72: Remove this except block - this will prevent exceptions from being raised. It's good to log the exceptions, but generally if an exception occurs here, it would be appropriate to raise it.

parsons/utilities/ssh_utilities.py
```diff
@@ -0,0 +1,80 @@
...
+    finally:
+        if con:
+            con.close()
+            logging.info("Database connection closed.")
+        if server:
+            server.stop()
+            logging.info("SSH tunnel closed.")
```
austinweisgrau, line 79: The finally block is good though

parsons/action_network/action_network.py
austinweisgrau, line 1: Technically, querying the ActionNetwork SQL mirror is a new connector. It has different dependencies, it queries data from a different source. I get the impulse to have it as a method on the existing ActionNetwork API connector, but I think that isn't the right move for a few reasons.

- It adds two dependencies to the ActionNetwork connector that API users don't need: the redshift extra and the sshtunnel package
- The method on the ActionNetwork connector to query the SSH mirror doesn't do anything - it wraps the query_through_ssh method without doing anything specific.

I think the query_through_ssh should be added as its own connector as you do here, but the ActionNetwork connector should be left unchanged. Anyone who wants to query the ActionNetwork SQL mirror can use the query_through_ssh method, and nothing needs to be done with the ActionNetwork connector.

parsons/action_network/action_network.py
```diff
@@ -1939,3 +1940,65 @@ def get_unique_id_list(self, unique_id_list_id):
...
+                mirror_port,
+                mirror_db_name,
+                mirror_username,
+                mirror_password,
+                query,
+            )
+            return result
+        except Exception as e:
+            print(e)
+            return None
```
austinweisgrau, line 2004: Again just in general, avoid this pattern - it results in Exceptions that should be raised not being raised.

test/test_action_network/test_action_network.py
austinweisgrau, line 1: Since the new method on the ActionNetwork connector doesn't do anything specific, it just calls the query_through_ssh method and passes all arguments to it, there's no need for an additional test. This test can be removed.

austinweisgrau (changes requested):

austinweisgrau (changes requested):
One last thought is that since the SSHTunnel utility adds a new dependency, you'll need to include it as an extra in the setup.py file. I think it would be appropriate to add a new extra called `ssh` where the `sshtunnel` dependency can be included. e.g. the new extra will be:
`"ssh": ["sshtunnel", "psycopg2-binary>=2.9.9", "sqlalchemy >= 1.4.22, != 1.4.33, < 2.0.0"]`
 because you'll need to include the postgres connector's dependencies as well

requirements-dev.txt
```diff
@@ -8,3 +8,5 @@ pytest-mock==3.12.0
 pytest==8.1.1
 requests-mock==1.11.0
 testfixtures==8.1.0
+sshtunnel==0.4.0
```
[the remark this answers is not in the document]
NirTatcher: Done.

parsons/utilities/ssh_utilities.py
NirTatcher: Done, we are now raising the exception after logging it. Hope this is ok and please let me know if you prefer the code without the except block all together.

parsons/action_network/action_network.py
NirTatcher: Makes a lot of sense. Just to be sure I've added to the ActionNetwork docs an example of how you would call `query_through_ssh` to query your SQL Mirror.

test/test_action_network/test_action_network.py
NirTatcher: Done.

NirTatcher:
> One last thought is that since the SSHTunnel utility adds a new dependency, you'll need to include it as an extra in the setup.py file. I think it would be appropriate to add a new extra called `ssh` where the dependencies can be included. e.g. the new extra will be: `"ssh": ["sshtunnel", "psycopg2-binary>=2.9.9", "sqlalchemy >= 1.4.22, != 1.4.33, < 2.0.0"]` because you'll need to include the postgres connector's dependencies as well

Done.
In general thank you again @austinweisgrau for the review, feedbacks I really appreciate it!

austinweisgrau (approved):
Thank you for making those changes! This looks much better to me.

PR #1103, opened 2024-07-19
Update old reference to Parsons API key type

### LINKED ISSUE #1084: NGPVAN authentication docs need to be updated
It was reported in the Slack that NGPVAN has removed the Parsons API key, and that people who want to use Parsons with Van will now have to request a custom integration.

This is a bummer, but it is what it is, so we should update [our docs](https://move-coop.github.io/parsons/html/stable/ngpvan.html).

Marking this as high priority since this is one of our most highly used connectors.

SorenSpicknall:
Resolves #1084 by updating the language previously used to refer to the now-nonexistent "Parsons" API key type provisioned by NGPVAN, pointing users to request a custom integration instead.

### DIFF
--- a/docs/ngpvan.rst
+++ b/docs/ngpvan.rst
@@ -13,9 +13,10 @@ additional details and information.
 .. note::
    API Keys
       - API Keys are specific to each committee and state.
-      - There is a Parsons type API Key that can be requested via the Integrations menu on the main page.
-      	If you have an issue gaining access to this key, or an admin has questions, please email
-	<parsons@movementcooperative.org>.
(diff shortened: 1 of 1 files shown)


rdhyee (approved):
I've read https://github.com/move-coop/parsons/issues/1084 and the text of the revision and as a project newbie, the change sounds good to me.  I'm going to be bold and hit "Approve" but I'd want someone like @shaunagm to say yes or no.

shaunagm (approved):
Tempted to add a 😞 emoji here lol but approved

shaunagm:
And thanks for the initial review @rdhyee!

PR #1104, opened 2024-07-19
Restore functionality of Google Sheets API tests

### LINKED ISSUE #1060: [Bug] Fix tests for `test_google_sheets`
## Detailed Description

There are some fixes we need to make to https://github.com/move-coop/parsons/blob/main/test/test_google/test_google_sheets.py. Specifically:
- tearDown is commented out - I presume we should re-enable it.
- Add a few strategic `time.sleep` calls so that you don't hit the [rate limit](https://developers.google.com/sheets/api/limits#quota).
  - Longer term maybe we want to look at https://docs.gspread.org/en/v6.0.0/api/http_client.html#gspread.BackOffHTTPClient ?
- `test_read_worksheet` and `test_read_sheet` read from fixed worksheets that many folks will not have access to. Can we have this point to the spreadsheet we make in setUp instead?
- `test_append_to_spreadsheet` literally says "BROKEN TEST", we should fix.

SorenSpicknall:
Resolves #1060 - I agree with [anzelpwj](https://github.com/anzelpwj) that it would be better to use backoff to avoid rate limiting in the long run (likely also true of other Parsons test scripts), but for the time being I've just instituted some judicious pauses along the test path.

### DIFF
--- a/test/test_google/test_google_sheets.py
+++ b/test/test_google/test_google_sheets.py
@@ -1,6 +1,7 @@
 import unittest
 import gspread
 import os
+import time
 
 from parsons import GoogleSheets, Table
 from test.utils import assert_matching_tables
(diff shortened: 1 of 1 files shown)


shaunagm (commented):
So this would add 40 seconds to the whole test suite, every time? That seems not ideal, although not a huge problem if this is just a quick fix.

It's not clear when gspread's BackOffHTTPClient will be ready but maybe we could mimic something like that with a decorator that looks for a given exception, waits, and retries? So that we're not adding 40+ seconds to our tests unless we have to?

This [stack overflow question](https://stackoverflow.com/questions/50246304/using-python-decorators-to-retry-request) gets at what I'm thinking.

SorenSpicknall:
> So this would add 40 seconds to the whole test suite, every time? That seems not ideal, although not a huge problem if this is just a quick fix.
> 
> It's not clear when gspread's BackOffHTTPClient will be ready but maybe we could mimic something like that with a decorator that looks for a given exception, waits, and retries? So that we're not adding 40+ seconds to our tests unless we have to?
> 
> This [stack overflow question](https://stackoverflow.com/questions/50246304/using-python-decorators-to-retry-request) gets at what I'm thinking.

I definitely agree from a design perspective, just trying to get things back up and running for now with limited time to contribute. FWIW, these tests are all _only_ live tests, so they won't run during a typical test run.

If that's a satisfactory limitation, we can merge as-is. If you prefer to implement a better approach right off the bat, it might take me awhile to get the time for it, but I'm down. My personal preference, given that these are live-only tests, is to merge this and then revisit for a better approach later since the current tests are fully broken.

shaunagm (approved):
I didn't notice that these were live tests. I think given that live tests don't run on each PR the added delays are not as problematic, and I'm ok to merge.

PR #981, opened 2024-01-27
Issue 910 - Added timeout to BigQuery load_job.result()

bear-jordan:
# Overview

This PR addresses [Issue #910](https://github.com/move-coop/parsons/issues/910). A timeout is added to prevent indefinite hanging.

From the original issue:

> Per [Google BQ query job documentation](https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.job.QueryJob#google_cloud_bigquery_job_QueryJob_result), it's possible to pass a default timeout parameter to the synchronous call that awaits a response.

> This change would require choosing a reasonable maximum default (30s?), setting that up as a configurable parameter, and passing that parameter into load_job.result.

# Changes
27/1/2024
- ~~Added `MAX_TIMEOUT = 30` constant~~ 
- ~~Updated `load_job.result()` to use `MAX_TIMEOUT`~~

31/1/2024 — @Jason94 
- Added `max_timeout: int = 30` as a keyword argument
- Updated `load_job.result()` to use `max_timeout`
- Added `DeadlineExceeded` error handling ([docs](https://googleapis.dev/python/google-api-core/latest/exceptions.html?highlight=deadlineexceeded#google.api_core.exceptions.DeadlineExceeded))

PR #1105, opened 2024-07-19
Add configurable timeout to BigQuery query jobs

### LINKED ISSUE #910: [Feature/Addition] Add Default Timeout to BigQuery Database Connections
Set a default `timeout` parameter for BigQuery queries.

## Detailed Description

Per [Google BQ query job documentation](https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.job.QueryJob#google_cloud_bigquery_job_QueryJob_result), it's possible to pass a default timeout parameter to the synchronous call that awaits a response.

This change would require choosing a reasonable maximum default (30s?), setting that up as a configurable parameter, and passing that parameter into `load_job.result`.

## Context

Generally, it's best practice to establish a timeout for database connections, to prevent code from indefinitely waiting for a response under unreliable network conditions. 

Per https://github.com/move-coop/parsons/issues/452, there was existing low-priority work to set this up with the built-in SQL connector.  Having a well-chosen default timeout and good notifications around errors will make everything dependent on parsons more robust.

## Possible Implementation

Should be as easy as changing 
https://github.com/move-coop/parsons/blob/55d371712193287fa775b23299e6e65756a26a17/parsons/google/google_bigquery.py#L417

and 
https://github.com/move-coop/parsons/blob/55d371712193287fa775b23299e6e65756a26a17/parsons/google/google_bigquery.py#L572

to `load_job.result(timeout=MAX_TIMEOUT)`

## Priority

High -- this is a very low lift now, but becomes much harder to implement and test once more code becomes dependent on the new BQ connector.

SorenSpicknall:
This picks up where somebody else's PR, #981, left off earlier this year. That PR closed because the original contributor deleted their fork after waiting unsuccessfully for merge.

This takes the spirit of those earlier proposed changes and adapts it to the current state of the BigQuery connector, hoping to resolve #910 once merged. I'm definitely open to a different approach if folks' thoughts on how to implement this default timeout have evolved since that last PR was active - @Jason94, I know you were part of that conversation at the time.


austinweisgrau:
Love it. Do you know if there's a default timeout in bigquery? 30 seconds seems perhaps a bit short to use as a default

shaunagm:
It looks like the default timeout is [6 hours](https://cloud.google.com/bigquery/quotas)? But @SorenSpicknall might know better than me, I don't use BigQuery and just got that from searching the documentation.

austinweisgrau (dismissed):
Shauna is right, the default BigQuery timeout is 6 hours. Definitely let's leave that default timeout at 6 hours and not shrink it to 30 seconds, that would cause some carnage

shaunagm:
@SorenSpicknall no rush, but this PR is ready to merge once the default timeout is set to 6 hours instead of 30 seconds. If you'd like I can change it for you & then merge.

SorenSpicknall:
All the test failures have to do with the Braintree version mismatch issue that @bmos has drafted a fix for in #1263, so once that's merged I can bring this branch up to date and we should be good to merge.

shaunagm (approved):

PR #1109, opened 2024-07-29
Allow to_gcs_csv function to accept a gcs instance instead of creating one

KasiaHinkson:
At TMC, we instantiate BigQuery and GCS outside of main, for dependency injection, but this function always creates a new GCS instance inside of it. This is breaking some of our code. This PR allows the user to send in a GCS instance, but should be backwards compatible and not breaking anyone else's code.

### DIFF
--- a/parsons/etl/tofrom.py
+++ b/parsons/etl/tofrom.py
@@ -464,6 +464,7 @@ class ToFrom(object):
         self,
         bucket_name,
         blob_name,
+        gcs=None,
         app_creds=None,
         project=None,
         compression=None,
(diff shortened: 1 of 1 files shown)


austinweisgrau (changes requested):
Appreciate this change! This is good.

Can you make this argument named `gcs_client` for more clarity and to match other methods that allow this argument? e.g. [this line](https://github.com/move-coop/parsons/blob/c7ac48e3bf2b568376ac3b178bb53e5e993d1426/parsons/google/google_bigquery.py#L739)

KasiaHinkson:
Absolutely, I was following TMC naming conventions, definitely want to use Parsons conventions here. Thanks for catching that!

austinweisgrau (approved):

PR #1111, opened 2024-08-03
Added method to GoogleBigQuery to copy between projects

Charlie-Kramer:


parsons/google/google_bigquery.py
```diff
@@ -1388,6 +1388,93 @@ def extract(
...
+            dataset_id = destination_project + "." + destination_dataset
+            self.client.get_dataset(dataset_id)  # Make an API request.
+            # if it exists: continue
+        except NotFound:
+            # if it doesn't exist: check the flag for creating it
+            if if_dataset_not_exists == 'create':  # create a new dataset in the destination
+                dataset = bigquery.Dataset(dataset_id)
+                dataset = self.client.create_dataset(dataset, timeout=30)
+            else:  # fail
+                print("BigQuery copy failed")
```
shaunagm, line 1444: I think it makes sense to use the logger here (and throughout the method) rather than print. [Here's](https://github.com/move-coop/parsons/blob/46e551edae0739007c682c70559f8a23532bbfee/parsons/google/google_bigquery.py#L110) an example of the logger in this file - if you've not used it before and are unfamiliar with log levels happy to explain a bit more how they work.

parsons/google/google_bigquery.py
```diff
@@ -1388,6 +1388,93 @@ def extract(
...
+
+        except NotFound:
+            # table doesn't exist
+            try:
+                job = self.client.copy_table(source_table_id, destination_table_id,
+                                             location='US', job_config=job_config)
+                result = job.result()
+                print(result)
+            except Exception as e:
+                print("Exception during copy between projects", e)
```
shaunagm, line 1476: While it makes sense to capture the NotFound exceptions and log the results without actually raising the exception, this exception here just captures any exception and saves the feedback. Generally speaking we want to raise any unexpected errors so we can notice and fix whatever the problem is sooner rather than later, so I would remove this specific try/except block and just call the code normally. If there's an exception, there's an exception - and we want to know about it!

shaunagm: Generally speaking, in software development, "don't catch general exceptions" is a good rule of thumb. [This blog post](https://jerrynsh.com/stop-using-exceptions-like-this-in-python/) has some good details of why.

shaunagm (changes requested):
Changes requested - happy to chat more about any of my review if you'd like. Also, ideally there would be a new test for this new method.

parsons/google/google_bigquery.py
```diff
@@ -1388,6 +1388,93 @@ def extract(
...
+                job_config = bigquery.CopyJobConfig()
+                job_config.write_disposition = "WRITE_TRUNCATE"
+                try:
+                    job = self.client.copy_table(source_table_id, destination_table_id,
+                                                 location='US', job_config=job_config)
+                    result = job.result()
+                except Exception as e:
+                    print("Exception during copy between projects", e)
+            else:
+                print("BigQuery copy failed")
```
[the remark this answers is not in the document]
Charlie-Kramer: Maybe I shouldn't do the try/except at all but just create the job and if there's a general exception just let the computer say so. (I think this would be an example of your guidance below re general exceptions?)

parsons/google/google_bigquery.py
Charlie-Kramer: Thanks--this looks like an error level (from what I know)--does that sound right?

shaunagm:
@Charlie-Kramer you seem to have accidentally added a bin folder with a bunch of files within - can you update the PR to remove them? Let me know if you need help

shaunagm:
@Charlie-Kramer overall this looks good, only two issues. First the extra files that got included in this PR that will need to be removed, and second it would be good to have a test for this new method. In particular, a test that tests the logic around missing datasets/tables/etc would be good.

shaunagm (changes requested):
Just updating the review status to reflect that you're working on making changes (let me know if you need help with anything, always happy to pair!)

Charlie-Kramer:
Hey thanks! I added the tests and cleaned up the formatting problems so I think it's ready for review. (Sorry prob should have put that in the message)

shaunagm:
Ah gotcha, I saw that the extra files are still there and assumed that meant it wasn't ready. Can we get rid of those? It'll make it much easier for me to review since I'll be able to see only the relevant files, and it will need to happen anyway before I can merge.

Charlie-Kramer:
Oh no! Sorry about that, let me fix it. Not sure what happened, I deleted them from my local rep using git and put them in .gitignore.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

shaunagm:
I think the issue is that the files were added before you put them in the .gitignore, so they're still there in the PR - they have to be manually removed. Adding them to the gitignore prevents this from happening in the future but does not fix the currently added files.

Charlie-Kramer:
> I think the issue is that the files were added before you put them in the .gitignore, so they're still there in the PR - they have to be manually removed. Adding them to the gitignore prevents this from happening in the future but does not fix the currently added files.

ah, I removed a bunch of them using git rm but there are others I missed. I did git rm bin -r, I hope this does the trick?

shaunagm:
When you did git rm, did you commit the changes? I don't see any new commits from you.

shaunagm:
Looks like it worked!

pyvenv.cfg
```diff
@@ -0,0 +1,3 @@
+home = /Users/charleskramer/opt/anaconda3/bin
```
shaunagm, line 1: looks like you missed one last file that should be gitignored/removed

Charlie-Kramer: ah let me get that one now. fixed the f string thing too.

test/test_google/test_bigquery_copy.py
```diff
@@ -0,0 +1,116 @@
+# testing bigquery copy method of google class
```
shaunagm, line 1: So I see why you put this in `test_google` but bigquery code is actually tested under `test_databases`. It's an unfortunate bit of organization since the bigquery *code* is not in the databases folder (I'm not entirely sure it makes sense to have a databases folder at all, but that's a decision that predates me.

Anyway, can you move this to the existing test file [here](https://github.com/move-coop/parsons/blob/main/test/test_databases/test_bigquery.py)? I'm not sure if there's anything in the existing test file you'd want to reuse to make your tests easier to read but otherwise these tests look fine. 

Apologies again for the confusing organization.

Charlie-Kramer:
You bet! Thanks for your patience, I’m kinda making a mess here.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

shaunagm:
You're doing great! It's Parsons that's messy :)

Can you move the bigquery copy test into the [test_bigquery.py](https://github.com/move-coop/parsons/blob/main/test/test_databases/test_bigquery.py) file? You can keep it as a separate test class although you'll need to rename your class so you don't overwrite the existing class (both are named `TestGoogleBigQuery`. `TestGoogleBigQueryProjectCopy` or similar would be perfectly fine

Charlie-Kramer:
Sure! But then it won’t run automatically yes? (Maybe that’s the idea?)

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

shaunagm:
As long as it's in a file that starts with `test_*` and is a class that
inherits from `TestCase` and as long as the individual methods start with
`test_*` it will run automatically. You can have multiple test classes in
the same test file and they'll all run as long as they each individually
satisfy the unittest syntax rules. And as long as they don't have the same
names (if they have the same names, whatever comes last will overwrite the
earlier versions)

Charlie-Kramer:
can we do a pair programming at some point? I reformatted using ruff locally but now it fails the ruff checks on github. I tried going back to a previous commit but that didn't work.

shaunagm:
Sure - you have my cal.com link in Slack DMs, right? Feel free to sign up
for whatever's next available (I'm off for holidays but I should have
normal times again in the new year)

Charlie-Kramer:
Ah I got it to pass all the tests! I did something with ruff that wasn't entirely sensible I think.

Anyway my main question is how to run the tests locally but I can post on the slack about this.

Happy Holidays! Hope you're having a good one.
________________________________

shaunagm (approved):
Looks great to me! Thanks for persevering through all the weird testing and gitignore issues :)

Charlie-Kramer:
> Looks great to me! Thanks for persevering through all the weird testing and gitignore issues :)

Thanks for all the great coaching!

PR #1112, opened 2024-08-05
Kasiah/community dot com connector

KasiaHinkson:
Creates a new connector for the Community.com API


parsons/community/community.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+        response = self.client.get_request(url=url)
+        return response
+
+    def get_resource(self, resource):
+        """
+        Get specified data from Community.com API as Parsons table.
+
+        `Args:`
+            resource: str
```
shaunagm, line 84: Can you rename the parameter and update the docs so we're using `file_name` instead of `resource`? As we chatted about on our call, this will make it easier for people to understand the `file_name error` response. Please also include a link to the underlying com api docs.

parsons/utilities/api_connector.py
```diff
@@ -96,9 +96,12 @@ def get_request(self, url, params=None):
 
         r = self.request(url, "GET", params=params)
         self.validate_response(r)
-        logger.debug(r.json())
 
-        return r.json()
```
shaunagm, line 101: This is a breaking change (though mostly internal) - can we switch this to passing in a `format` parameter that defaults to 'json', but if not 'json' but instead 'content', gets `r.content`?

test/test_community/test_community.py
```diff
@@ -0,0 +1,35 @@
...
+
+TEST_CLIENT_ID = "someuuid"
+TEST_CLIENT_TOKEN = "somesecret"
+
+TEST_ID = "12345"
+TEST_URI = "https://faketestingurl.com/example"
+
+TEST_CSV_TYPE = "campaigns"
+
+TEST_GET_RESPONSE = {
```
shaunagm, line 15: Can we rename this or leave a code comment that makes it clearer that this is mock data equivalent to the response from community.com that contains a csv string?

test/test_community/test_community.py
```diff
@@ -0,0 +1,35 @@
...
+        Table.from_csv = MagicMock(name="mocked from_csv", return_value=test_csv_data)
+
+    def tearDown(self):
+        Table.from_csv = self.from_csv
+
+    @requests_mock.Mocker()
+    def test_successful_get_request(self, m):
+        m.get(f"{TEST_URI}/{TEST_CLIENT_ID}/{TEST_CSV_TYPE}.csv.gz", json=TEST_GET_RESPONSE)
+
+        assert self.com.get_request(resource=TEST_CSV_TYPE) == TEST_GET_RESPONSE
```
shaunagm, line 35: Let's add one more test for `get_response` - should be similar to your existing test, but will test that we successfully transform the request response into a parsons table.

shaunagm (changes requested):
looks pretty good, just a couple of changes requested (as we talked about on our call)

KasiaHinkson:
@shaunagm this is ready for re-review, and it includes a small change to `gcs.upload_table` that I think might technically be a breaking change, so I'd like your eyes on it for sure.

parsons/community/community.py
```diff
@@ -0,0 +1,101 @@
...
+       `Args:`
+            community_client_id: str
+                The Community provided Client ID. Not required if ``COMMUNITY_CLIENT_ID`` env
+                variable set.
+            community_access_token: str
+                The Community provided access token. Not required if ``COMMUNITY_ACCESS_TOKEN`` env
+                variable set.
+            community_uri: str
+                The URI to access the  API. Not required, default is
+                https://dl.community.com/download/v1/files/. You can set an ``COMMUNITY_URL`` env
```
shaunagm, line 24: Sorry I didn't catch this before. You can format external links using [this syntax](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages).

parsons/community/community.py
```diff
@@ -0,0 +1,101 @@
...
+                variable or use this URI parameter if a different endpoint is necessary.
+
+    API Documentation: https://developer.community.com/reference/data-export-api-downloading-data
```
shaunagm, line 27: Can you switch to using a hyperlink instead of the bare url?

parsons/google/google_cloud_storage.py
```diff
@@ -327,17 +323,18 @@ def upload_table(
             # CSVView. Once any transformations are made, the Table.table
             # becomes a different petl class
             if isinstance(table.table, petl.io.csv_py3.CSVView):
```
shaunagm, line 325: This change looks okay to me, I'm curious about the reasoning? That will help me understand how big the breaking change would be - right now this code is so similar that it's hard to tell how it might cause problems for downstream users.

shaunagm (changes requested):
We're getting very close! A bit confused about the gcs change, once we've got that figured out and the links in the docstrings updated we should be good to go.

parsons/google/google_cloud_storage.py
KasiaHinkson: The reasoning for the change is that something about how our Parsons operator is working is changing the attributes of the CSVView, and we just need to do the `to_csv` instead of getting the filename. It would presumably break for someone if they want it to fail when encountering an Attribute Error. I assume this is unlikely and that this probably wouldn't really break anything for anyone, but technically if someone is depending on getting an AttributeError in some case, this would mean they instead get an actual table. Not sure why that would be a problem, but trying to be proactive!

shaunagm: Okay gotcha, that makes sense. Yeah, it's hard to imagine someone depending on getting an error there - there's no reason to call the function at all if you error out there. It hasn't uploaded the data to the blob yet, and it's not saving anything on the object that you can access, so there's no reason to call the function and catch the attribute error unless you're checking if `table.table.source.filename` exists, doing something, then calling the function again. But that would be a very weird pattern, like you could literally just check if `table.table.source.filename` exists directly. So yes, this is technically a breaking change but I think the chances of anyone being bothered by it are very very small so we are okay to merge it into main instead of major release.

shaunagm (approved):

shaunagm:
This is my bad, I should've caught this on review - there's no docs for this connector, so it's not showing up on the docs sidebar and thus it's a secret undiscoverable connector. Can we get docs added to this, even bare minimum ones (full ones would have authentication info and a little quickstart blurb, bare minimum is a one sentence description of the connector and a link to the autodocs). More [here](https://www.parsonsproject.org/pub/build-a-new-connector#documentation)

shaunagm:
I can make an issue for this if needed

PR #1115, opened 2024-08-09
Gcs upload table csview fix

sharinetmc:
A quick fix because it seems like the attributes for a CSVView isn't consistent across ALL views, (i.e. I'v  been getting errors like `AttributeError: 'URLSource' object has no attribute 'filename'`, which is indeed true when I looked at the object. Treating it as a normal csv seemed to have done the trick!


sharinetmc:
@KasiaHinkson, seems to be affecting the ACTBLUE CSV API sync. Hoping that you can review this. Hoping to get a solid run of the parsons operator so that I can put up a PR for it!

KasiaHinkson:
I added this to my branch, so you can close this I think

PR #1121, opened 2024-09-05
Version bump for certain packages

anzelpwj:
Due to some packages found in security review, incrementing these packages in `requirements.txt`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,7 +19,7 @@ grpcio==1.62.2
 gspread==3.7.0
 httplib2==0.22.0
 joblib==1.2.0
-mysql-connector-python==8.0.18
+mysql-connector-python==8.0.19
 newmode==0.1.6
 oauth2client==4.1.3
 paramiko==3.4.0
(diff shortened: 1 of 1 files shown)


anzelpwj:
Note with mysql-connector-python at 8.3.0 we lose connectivity to older mysql versions (5.5 and 5.6) but does get official support for Python 3.12. See https://dev.mysql.com/doc/connector-python/en/connector-python-versions.html.

anzelpwj:
The `build` step failed due to a timeout for its installation - not sure what to do, since all the other steps passed.

https://app.circleci.com/pipelines/github/move-coop/parsons/5703/workflows/d9f6865f-585f-4d67-887a-dc53735de8b5/jobs/7289?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

bmos:
As it says in the logs, `INFO: This is taking longer than usual. You might need to provide the dependency resolver with stricter constraints to reduce runtime. See https://pip.pypa.io/warnings/backtracking for guidance.`

You probably just need to set google-api-core to a specific version or update some other packages that depend on it.

bmos:
I'm also looking at updating the circleci workflow so that it uses uv which should help avoid timeouts during package resolution.
https://github.com/move-coop/parsons/pull/1122

anzelpwj:
Sounds good @bmos - I tried pinning the google-api-core and it went ~30 minutes fussing around before timing out again. I'll await your PR getting merged in.

bmos:
It might also be good to post about these in the Slack if you are there.

shaunagm:
Here's the error I see on the docs-build, during the install dependencies step:
Screenshot from 2024-09-06 11-52-51

bmos:
curlify 2.2.1 will probably solve that as it removes ">= 3" shlex version (which does not exist)

anzelpwj:
I'll cut out the mysql upgrade for now so this isn't a breaking change. Will also pin curlify. Will that fix things? Guess we'll find out.

anzelpwj:
Issue for upgrading `mysql-connector-python`: #1125

shaunagm (approved):

PR #1122, opened 2024-09-06
circleci updates

bmos:
this updates circleci requirements and python version. it should make doc creation dependency installation faster so it doesn't time out as easily.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,9 +34,10 @@ jobs:
           # Note that we the circleci node image installs stuff with a user "circleci", rather
           # than root. So we need to tell npm where to install stuff.
           command: |
-            python3 -m venv venv
-            . venv/bin/activate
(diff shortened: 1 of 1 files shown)


shaunagm:
I'm a little concerned that only 4 tests are being run. In my experience that's a sign of some kind of formatting or configuration error, although I could be wrong!

anzelpwj:
@shaunagm that's because the other tests specify that they only run when certain files (which does not include the docs requirements.txt file) are changed.

```yaml
on:
  pull_request:
    paths:
      - .github/workflows/pip-install.yml
      - requirements.txt
      - setup.py
  push:
    paths:
      - .github/workflows/pip-install.yml
      - requirements.txt
      - setup.py
  workflow_dispatch:
  ```

shaunagm (approved):

Issue #1114, opened 2024-08-08
Test Coverage Assessment


jdw25:
HIGH LEVEL: we probably want to answer (at least):
- Are our current tests doing what we think they are doing?
- Do we have test gaps

bmos:
We have 78% coverage
https://github.com/move-coop/parsons/actions/runs/10742012333/job/29793717439

bmos:
Here is updated information about our test coverage:
https://htmlpreview.github.io/?https://github.com/move-coop/parsons/blob/python-coverage-comment-action-data/htmlcov/index.html

bmos:
parsons/utilities/dbt/* is entirely untested, via mock or live tests.
parsons/google/google_sheets.py has 14% coverage
parsons/aws/s3.py has 15% coverage
We have 14 files with <20% mock coverage.

bmos:
Since my last comment, things have improved but we still have a lot of room for improvement.
We have 70% overall coverage.

dbt now has decent coverage at 69%
google sheets remains at 14%
aws s3 still has 15% coverage
rs_table_utilities.py has 15% coverage
There are 6 more files with <20% coverage

PR #1126, opened 2024-09-06
Code coverage reporting

bmos:
Prompted by https://github.com/move-coop/parsons/issues/1114 I created coverage testing.

The main thing about this that I don't like is that it reports on each version which is probably pointless? Does anyone know if that is worth doing or not?

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,50 @@
+name: Pytest and ruff
+
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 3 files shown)


shaunagm:
Each version of what? Python? Parsons?

bmos:
Ah, sorry. Each version of Python and each testing platform.

shaunagm:
Yeah, feels like we should only have to run it once. It's not like the coverage would differ on those versions/platforms.

PR #1128, opened 2024-09-07
fix gmail test failing with invalid email address

bmos:
not sure why this didn't show up yet via github actions but I just downloaded the main branch and ran pytest locally and was getting a failed test here because ,com is not valid but the test expects it to validate...


bmos:
I'll have to dig into this more, I my local package versions must be different or something...

PR #1132, opened 2024-09-07
Github Actions Rebuild

bmos:
This is a total rebuild of how our action yml files are structured, with the python tests now occuring in a single multi-job action. This works [much better](https://github.com/move-coop/parsons/actions/runs/10753195270) in the GitHub web UI. This also allows the easy stuff (linting, formatting, etc) to finish much faster than the pytest runs. It also adds additional security checks and automatic dependency updates.

## What it does

- Run pytest on every version of macOS (since only pytest and pip install run on macOS it should use too much time -- linting, formatting, coverage, etc are only run once).
- Remove file name filters on action triggers so actions run on ever pushed commit / active pull request rather than trying to limit runs to when certain files are updated (to ensure nothing is missed and make it more obvious if actions fail to run for some reason)
- Move test.yml and pip-install.yml to shared python-checks.yml
- Add automatic weekly dependency pull requests with [dependabot](https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide) covering github actions, docker, and pip
- Add code coverage checking with [coverage](https://coverage.readthedocs.io) failing when the project has <75% coverage (for now) - important for #1114 
- Add code security checks with [bandit](https://bandit.readthedocs.io) (filtered to medium or higher risk and medium or higher confidence)
- Update security scorecard to pin hash of github/codeql-action/upload-sarif
- Update security scorecard to first harden the runner
- Update security scorecard to scan major-release branch along with main
- Update ruff, pytest, and similar

## What's needed to merge

- [x] Resolve the [4 issues](https://github.com/move-coop/parsons/actions/runs/10752486576/job/29820855435) found by bandit
  - [x] defusedxml
  - [x] md5


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,77 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+    push:
+        branches: [ "main", "marjor-release" ]
```
anzelpwj, line 16: I assume this should be `major-release`, not `marjor-release`

anzelpwj:
Just FYI to anyone doing a more proper review - the reason a large number of files are changed is because `isort` sorted the imports.

bmos:
I'd be happy to revert that commit if it makes it easier to view the changes and then re-commit it once we're happy with this. Just LMK if that is preferred 👍🏻

shaunagm (approved):
Looks good to me :)

bmos:
So the issue is that the gmail address validation is acting differently across different platforms and python versions.
Specifically, the test of validation behavior for email addresses of ",com" instead of ".com" is passing (mostly on older versions). I'm guessing there is a package we don't have pinned to a specific version and it's exhibiting different behavior based on what version is installed.

shaunagm:
I did a little investigating. The test failure is on the gmail connector, which inherits from the sendmail connector; the sendmail connector has a method [_validate_email_string](https://github.com/move-coop/parsons/blob/1163cc025a03c5da0f4c2cda52134f1eee5f40cf/parsons/notifications/sendmail.py#L191) which calls [validate_email](https://github.com/move-coop/parsons/blob/1163cc025a03c5da0f4c2cda52134f1eee5f40cf/parsons/notifications/sendmail.py#L198) which appears to be [this package](https://pypi.org/project/validate_email/). Our requirements.txt pins 1.3 which is the most recent version; unfortunately that most recent version was published [9 years ago](https://pypi.org/project/validate_email/). I'm not sure why the tests started failing just now (my guess is it's something to do with bumping the version of pytest) but regardless we probably don't want to be using a library that hasn't been updated in 9 years.

There is an [actively maintained fork](https://git.ksol.io/karolyi/py3-validate-email) but it says "Pypi removed newer versions due to license issues". So...maybe the thing to do is to migrate to a different package entirely? Josh Tauberer has an [email validation package](https://pypi.org/project/email-validator/) we could switch to.

bmos:
I did try reverting to the old version of pytest et al and it was still failing, but I agree it's probably a good idea to use better validation. (especially since the old solution allowed ",com" which is not valid afaik).

shaunagm:
I mentioned you in the other thread but I'll also add here that this failure does not seem to be specific to this PR. It's happening in all other recent PRs but not the most recent merged commit (which was 11:56am EDT yesterday Sep 17th). Given that there have been no updates to the github images between then and now, and the errors are happening at least on mac and ubuntu, it seems unlikely to be caused by a change to the runners. Maybe a change to uv? I don't think we pin our version of uv on install and there have been one, possibly two [releases of uv](https://pypi.org/project/uv/#history) since noon yesterday.

bmos:
Actually as of this PR we are using a pinned version of uv:
https://github.com/move-coop/parsons/pull/1132/files#diff-3e744f163053a3e58aeec8dede7e3711d8a87b2249a31fd94a884200b0aec1c7R31-R32

bmos:
This now works great on 3.12, but 3.8 is broken on all platforms and 3.9-3.11 only works on ubuntu...  Very strange behavior.
I'll try installing python 3.8 sometime soon and see if I can debug why that is failing.

shaunagm:
This is so weird. It's like whatever was causing the issue was fixed in Ubuntu for 3.9 but not in Mac/Windows until 3.12.

shaunagm:
For clarity, when you say "this now works great on 3.12" - do you mean that a validation error is raised as expected when given ",com"? 

I'm trying to wrap my head around where the problem could be coming from. Because until recently these same tests passed (on ubuntu and mac, at least) meaning that a validation error was raised, no? So something changed (either by parsons, uv, github, python, etc) that caused the validation error not to be raised. But now with this new package it's sometimes being raised.

bmos:
I think that one of our dependencies must have updated or something like that. Like if one of our dependencies has a subdependency that is not version pinned, it could potentially do something like this.

What I did to get the tests passing on 3.12 was change out the email validation library, but it still is passing ,con on certain configurations.

bmos:
We just recently updated some dependencies, so it could be related to that. But if I remember correctly, that pull request had all of the checks passed.

Unless maybe the checks didn't run because a particular file was being updated...

Probably time to jump into some actions logs and see what changed.

austinweisgrau:
I fixed the email parsing issue with #1146

shaunagm:
@austinweisgrau @bmos alas now all the tests are failing. FYI I did resolve the merge conflict between @austinweisgrau's PR and this one, perhaps incorrectly. But it also may be that the PR doesn't fully address the issue (I'm still unclear how a problem in Python 3.12 could cause our tests for 3.8-3.11 to fail, esp in the weird pattern they did (mostly but not entirely on Windows & Mac rather than Ubuntu)

austinweisgrau:
IMO this PR should be split up a bit. In particular the isort formatting should be its own PR. Because hundreds of files are changed with the isort formatting, it makes it very difficult to detect any other unwanted changes that are happening here, and also difficult to evaluate, for example, what changes in particular may be resulting in test failures.

shaunagm:
Yeah that's fair. We'd talked about moving the isort stuff to a separate PR a few weeks ago for precisely that reason but got derailed by the email address bug. @bmos if you'd be up for it can you split the PRs?

bmos:
I would be happy to, that should be pretty easy.

bmos:
I'm also pretty sure that none of these changes are creating test failures, but again, totally happy to break that out.

bmos:
Okay, that should take care of that.

shaunagm:
Looks good - thank you!

austinweisgrau:
hell yea good work

PR #1137, opened 2024-09-16
Postgres copy explicates columns and column order

austinweisgrau:
Without this change, the Postgres copy method only works if the Table has the same columns in the same order as the destination table. With this change, the source table can have a subset of columns in a different order but still successfully load to the destination table.


austinweisgrau:
Also now includes a couple other changes that make the Postgres query methods more robust:
- get_updated_rows() can now handle cutoff_value being None which happens sometimes
- a few Postgres query methods are able to handle case-sensitive column and table names, which happens sometimes

shaunagm (approved):
Looks good to me, sorry for the delay

PR #1141, opened 2024-09-19
updated README with correct instructions for installing from Github

lisamessier:


shaunagm (approved):
Looks good to me!

shaunagm:
Thanks so much for contributing @lisamessier! I don't think you're in our Slack - would you like to join? If so, email me at shaunagm@gmail.com and I'll add you :)

PR #1143, opened 2024-10-02
BigQuery copy method can convert dict column to JSON string

austinweisgrau:
By default without this change, a dict column will cause the copy method to fail.

This is a more robust followup to #1028 and #1068. At the time we decided to remove support for loading dict columns to BigQuery because there wasn't a super simple way to accomplish this. Instead we encouraged users to convert the dict columns to strings themselves if they desired to do so. This change makes that workaround happen by default, but it can be turned off with the `convert_dict_columns_to_json` flag to `BigQuery().copy()`

### DIFF
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+import json
 import pickle
 import random
 import uuid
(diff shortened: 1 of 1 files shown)


KasiaHinkson (approved):
I love it, I think this is an elegant solution that'll help a lot of people out.

Issue #1142, opened 2024-09-27
[Bug] van.apply_response raises unnecessary error 


lkesich:
## Detailed Description
The apply_response method from the VAN() class raises an error if the contact type is Phone, SMS etc and the phone number is missing. VAN made a change and no longer has phone number as a required field for bulk uploads with these contact types.  Would you please remove this exception if it no longer applies? 

From the source:

```
        if (
            contact_type_id == 1  # Phone
            or contact_type_id == 19  # Auto Dial
            or contact_type_id == 37  # SMS Text
            or contact_type_id == 67  # Phone Bank
            or contact_type_id == 68  # Consumer Phone
            or contact_type_id == 72  # Leader Phone
            or contact_type_id == 112  # Personal Phone
            or contact_type_id == 132  # Relational Text
            or contact_type_id == 143  # Distributed Text
            or contact_type_id == 147  # Bulk Text
            or contact_type_id == 149  # Paid SMS
        ):
            if not phone:
                raise Exception("A phone number must be provided if canvassed via phone or SMS")
```

## Your Environment

* Version of Parsons used (if you know): 3.1.0

## Priority
High

matthewkrausse:
I'll fix this.

lkesich:
@matthewkrausse I'm sorry, I noticed that this method is also used to apply canvass results from unsuccessful canvass attempts, and VAN does still require a phone number in that situation. It doesn't requires a phone number if you're applying survey responses/activist codes with a phone contact type. Sorry about that--I was having tunnel vision for my particular use case. I do still think the error should only be raised `if response is None` or `if resultid is not None` or with some other way of differentiating between successful and unsuccessful attempts.

bmos:
It seems like this may not have been fixed?

PR #1144, opened 2024-10-04
Fix: van.apply_response raises unnecessary error (Issue #1142 )

matthewkrausse:
This removes the error flag in Issue #1142. apparently VAN has removed the need to have a phone passed when a phone type contact type is used. Not sure why they did that but this makes this method in line with the api/


shaunagm (commented):
This looks good but is the addition of the optional campaignID parameter related? Just want to clarify it wasn't included by accident

shaunagm (approved):

PR #1146, opened 2024-10-09
email parse depends on python patch version

austinweisgrau:
The behavior of email.parseaddr depends on the python patch version.

See https://github.com/python/cpython/issues/102988 or associated changelogs, e.g. https://docs.python.org/3.8/whatsnew/changelog.html#python-3-8-20-final

### DIFF
--- a/test/test_gmail/test_gmail.py
+++ b/test/test_gmail/test_gmail.py
@@ -2,10 +2,10 @@ from parsons import Gmail
 import json
 import os
 import requests_mock
+import email
 import unittest
 import shutil
 import base64
-import email
 
 
 _dir = os.path.dirname(__file__)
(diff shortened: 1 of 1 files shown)


bmos:
Great work tracking down the changelog that describes this behavior!

shaunagm (approved):
Thanks so much for tracking this down! It was a tricky one. I like your solution to add a condition to the tests, rather than require folks to not pin to a specific python version.

shaunagm:
Although, question: how would this create the pattern of test failures we saw, where tests were failing on Mac and Windows up to 3.12 (but not 3.12) and Ubuntu on 3.8 (but not 3.9-3.12)?

austinweisgrau:
There isn't actually a breaking change in this PR, rather the change in behavior of email.parseaddr is with the python minor patch version which is very upstream of us.

I believe the the test failures were due to different times that the patches were merged into different python versions. The change in parseaddr was merged into each python major version in different releases at presumably different times.

shaunagm:
Sorry, that was a bit of inside baseball - I'm cutting the next release soon and I always check the PRs for breaking changes. When I see this PR, I'll remember to add a small note to the release notes calling out this fix in particular, just in case it's affected anyone else. We had a hard enough time detecting that it was due to a python patch version, so if a Parsons user runs into it I'd hope they could see it in the Parsons changelog notes instead of having to delve even further. 

tl;dr I won't actually list it as a breaking change.

shaunagm:
> I believe the different test failures were due to different times that the patches were merged into different python versions. The change in parseaddr was merged into each python major version in different releases at presumably different times.

Also this explanation makes sense, thank you!

PR #1148, opened 2024-10-10
Allow passing google auth Credentials directly to GCS connector

austinweisgrau:
Among other things, this enables authenticating the GCS connector using gcloud oauth from the parent shell, rather than requiring the use of service account credentials.


KasiaHinkson (approved):
Yep, LGTM!

PR #1149, opened 2024-10-10
SFTP client for CatalistMatch has 10 minute timeout

austinweisgrau:
This is important because sometimes the Catalist sftp server hangs indefinitely.


shaunagm (commented):
Ready to approve this, but just wanted to confirm that if the method hasn't returned in 10 minutes it's never going to return?

austinweisgrau:
> Ready to approve this, but just wanted to confirm that if the method hasn't returned in 10 minutes it's never going to return?

In my experience, that's right. The timeout doesn't actually check for the download to complete within that amount of time, but rather will trigger if the server hasn't sent back any new packets in that amount of time.

PR #1191, opened 2024-11-18
Kasiah/empower connector

KasiaHinkson:
Branched off @jburchard 's Empower connector PR. 

Adds a single method to return the full export unparsed, for pure ELT.


docs/empower.rst
```diff
@@ -0,0 +1,47 @@
...
+=======
+
+********
+Overview
+********
+
+The Empower class allows you to interact with the Empower API. Documentation for the Empower API can be found
+in their `GitHub <https://github.com/getempower/api-documentation/blob/master/README.md>`_ repo.
+
+.. note::
```
shaunagm, line 11: Can you change this note to include information about authentication, and then move this note into the overview? Just to maintain consistency with how other connectors docs work.

docs/empower.rst
```diff
@@ -0,0 +1,47 @@
...
+    The Empower API only has a single endpoint to access all account data. As such, it has a very high overhead. This
+    connector employs caching in order to allow the user to specify the tables to extract without additional API calls.
+    You can disable caching as an argument when instantiating the class.
+
+==========
+Quickstart
+==========
+
```
shaunagm, line 19: Typos: missing "as" in "as an environment variable"; also some confusion of singular key vs plural

parsons/empower/empower.py
```diff
@@ -0,0 +1,261 @@
...
+        self.api_key = check_env.check("EMPOWER_API_KEY", api_key)
+        self.empower_uri = (
+            check_env.check("EMPOWER_URI", empower_uri, optional=True) or EMPOWER_API_ENDPOINT
+        )
+        self.headers = {"accept": "application/json", "secret-token": self.api_key}
+        self.client = APIConnector(
+            self.empower_uri,
+            headers=self.headers,
+        )
+        self.data = None
```
shaunagm, line 40: this line seems unnecessary as you're overwriting it on the next line?

parsons/empower/empower.py
```diff
@@ -0,0 +1,261 @@
...
+        self.data = self._get_data(cache)
+
+    def _get_data(self, cache):
+        """
+        Gets fresh data from Empower API based on cache setting.
+        """
+
```
shaunagm, line 47: Oh if self.data was set to None so that this method works, it's probably better to move the logic inside the method. You could also do something like `self.get("data", None)` where it will look for data and return None if nothing's there instead of raising a KeyError

parsons/empower/empower.py
```diff
@@ -0,0 +1,261 @@
...
+        if not cache or self.data is None:
+            r = self.client.get_request(self.empower_uri)
+            logger.info("Empower data downloaded.")
+            return r
+
+        else:
+            return self.data
+
+    def _unix_convert(self, ts):
```
shaunagm, line 56: Do none of the utilities [here](https://github.com/move-coop/parsons/blob/main/parsons/utilities/datetime.py) cover this? If not, maybe we should add a util and reuse? (I don't have strong feelings about this, just theoretically that's what the utils are for)

parsons/empower/empower.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+        Converts UNIX timestamps to readable timestamps.
+        """
+
+        ts = datetime.utcfromtimestamp(int(ts) / 1000)
+        ts = ts.strftime("%Y-%m-%d %H:%M:%S UTC")
+        return ts
+
+    def _empty_obj(self, obj_name):
```
shaunagm, line 65: It feels a little weird to pull this out into its own method since it's only used once? But if we think it's likely we'll want re-use the helper method again might as well keep it separated out.

parsons/empower/empower.py
```diff
@@ -0,0 +1,261 @@
...
+
+        `Returns:`
+            Parsons Table
+                See :ref:`parsons-table` for output options.
+        """
+
+        tbl = Table(self.data["profiles"])
+        for col in ["createdMts", "lastUsedEmpowerMts", "updatedMts"]:
+            tbl.convert_column(col, lambda x: self._unix_convert(x))
+        tbl.remove_column("activeCtaIds")  # Get as a method via get_profiles_active_ctas
```
shaunagm, line 87: What is this line doing? I don't understand the code comment

parsons/empower/empower.py
```diff
@@ -0,0 +1,261 @@
...
+    def get_profiles_active_ctas(self):
+        """
+        Get active ctas assigned to Empower profiles.
+
+        `Returns:`
+            Parsons Table
+                See :ref:`parsons-table` for output options.
+        """
+
+        tbl = Table(self.data["profiles"]).long_table("eid", "activeCtaIds")
```
shaunagm, line 99: This is pure curiosity, what is a "long table"? Guess that's a part of Parsons core I've never had to touch before

parsons/empower/empower.py
```diff
@@ -0,0 +1,261 @@
...
+    def get_cta_results(self):
+        """
+        Get Empower call to action results.
+
+        `Returns:`
+            Parsons Table
+                See :ref:`parsons-table` for output options.
+        """
+
+        tbl = Table(self.data["ctaResults"])
```
shaunagm, line 124: There's enough data restructuring happening in this method that I would appreciate some code comments to help readers understand what's happening (ie so for instance one doesn't have to go read the docstring for `unpack_nested_columns_as_rows` to know the precise behavior - could just say `# unpacks answerIdsByPromptId into standalone rows`)

parsons/empower/empower.py
```diff
@@ -0,0 +1,261 @@
...
+        tbl.remove_column("answers")  # Per docs, this is deprecated.
+        return tbl
+
+    def _split_ctas(self):
+        """
+        Internal method to split CTA objects into tables.
+        """
+
+        ctas = Table(self.data["ctas"])
+        for col in [
```
shaunagm, line 142: formatting looks weird here but if it passes our linter/formatter I guess that's fine

parsons/empower/empower.py
```diff
@@ -0,0 +1,261 @@
...
+        ctas.remove_column("shareables")  # Get as a table via get_cta_shareables()
+        ctas.remove_column("prioritizations")  # Get as a table via get_cta_prioritizations()
+        ctas.remove_column("questions")  # This column has been deprecated.
+        cta_prompts = ctas.long_table("id", "prompts", prepend=False, retain_original=False)
+        cta_prompts.remove_column("ctaId")
+        cta_prompt_answers = cta_prompts.long_table("id", "answers", prepend=False)
+
+        return [ctas, cta_prompts, cta_prompt_answers]
+
+    def get_ctas(self):
```
shaunagm, line 159: Enough data restructuring is happening in these methods (or the private/utility methods they're calling) that it would be useful for each of the dcostrings for the public methods to describe the format of the data being returned.

parsons/empower/empower.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+        Get Empower calls to action.
+
+        `Returns:`
+            Parsons Table
+                See :ref:`parsons-table` for output options.
+        """
+
+        return self._split_ctas()[0]
```
shaunagm, line 168: If it's not too annoying, can we return a NamedTuple and then call by name instead of an index? I'm willing to be convinced this is too much boilerplate but it feels like it would help readability.

parsons/empower/empower.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def get_cta_prompts(self):
```
shaunagm, line 170: Same comments as previous method.

parsons/empower/empower.py
```diff
@@ -0,0 +1,261 @@
...
+        Get Empower calls to action prompts.
+
+        `Returns:`
+            Parsons Table
+                See :ref:`parsons-table` for output options.
+        """
+
+        return self._split_ctas()[1]
+
+    def get_cta_prompt_answers(self):
```
shaunagm, line 181: Same comment as previous previous method.

parsons/empower/empower.py
```diff
@@ -0,0 +1,261 @@
...
+            return Table([])
+
+        tbl = Table(self.data["outreachEntries"])
+        for col in [
+            "outreachCreatedMts",
+            "outreachSnoozeUntilMts",
+            "outreachScheduledFollowUpMts",
+        ]:
+            tbl.convert_column(col, lambda x: self._unix_convert(x))
+            logger.info(f"Unable to find column {col}")
```
shaunagm, line 247: Why is it logging 'unable to find {col}'?

shaunagm (changes requested):
Lots of comments but nothing major, excited to get this merged soon

parsons/empower/empower.py
KasiaHinkson: I think the comment means that if you want this column, you would get it using get_profiles_active_ctas

parsons/empower/empower.py
KasiaHinkson: You have your identifier columns (the values that tell you what row the data would belong to), then all your other variables are melting into `variable` and `value` columns. So for person A with variables X, Y, Z in a normal table, the long table will have 3 rows: 

```
Person     Variable     Value
A              X                x_value
A              Y                y_value
A              Z                z_value
```

shaunagm: Thanks for the explanation!

parsons/utilities/datetime.py
```diff
@@ -25,6 +25,17 @@ def date_to_timestamp(value, tzinfo=datetime.timezone.utc):
     return int(parsed_date.timestamp())
 
 
+def unix_convert(ts):
```
shaunagm, line 28: can we change this to `convert_unix_to_readable` or similar? just to make the name slightly more specific and descriptive?

shaunagm (approved):
One small additional change requested (re: the unix timestamp) but otherwise this looks good enough. I do wish there was a bit more documentation of how the methods are constructed and what they're doing to the data but I understand since you didn't write the code yourself that's hard to provide.

Issue #1138, opened 2024-09-17
Remove md5 security workaround after dropping 3.8


bmos:
Resolving one of bandit's flagged security issues required [adding an if/else statement](https://github.com/move-coop/parsons/commit/01ef28b856ce9126a16e06b486764d70d6ab1ff3) that can be reverted once we drop support for python 3.8.

DCRepublic:
I'd like to work on this, however the commit you linked doesn't seem to exist.

bmos:
This bug is actually relating to code that has not been merged yet, but I was asked to post it here. Python 3.8 should be dropped in December or early next year.

Issue #1066, opened 2024-05-29
Sunset support for Python 3.8


jdw25:
3.8 looks to be EOL as of ~Oct 2024 -- see [3.8 Lifespan](https://peps.python.org/pep-0569/#lifespan).
Also: Looks like some dependencies (such as Civis) have started dropping support for 3.8 in their latest versions.

PR #1192, opened 2024-11-21
Remove support for Python 3.8 (& fix failing MacOs 3.9 tests)

shaunagm:
Not sure what to do about the test failures of 3.9 and higher, but we can treat them separately.

For the record, there appear to be two distinct issues, a slightly older one where the combo of Mac, Python 3.9, and limited dependencies leads to [a psycopg2 install error](https://github.com/move-coop/parsons/actions/runs/11957548537/job/33334963382?pr=1192#step:5:24) and then one that's an issue for all limited dependency installs related to [this pydantic release](https://github.com/pydantic/pydantic/issues/10910) - the latter issue will hopefully have an easy fix soon but I'm not sure about the former.

### DIFF
--- a/.github/workflows/python-checks.yml
+++ b/.github/workflows/python-checks.yml
@@ -32,6 +32,10 @@ jobs:
             - name: Install uv
               uses: install-pinned/uv@de03c60d508703a83d3f8f49afcf1249590ecda1  # 0.4.12
 
+            - name: Patch install error when using Python 3.9, limited dependencies, and MacOS
+              if: ${{ matrix.limited-dependencies }} == True and ${{ matrix.os }} == "macos-latest" and ${{ matrix.python-version }} == "3.9"
+              run: uv pip install --system psycopg2-binary==2.9.9
+
             - name: Install dependencies
               env:
                   PARSONS_LIMITED_DEPENDENCIES: ${{ matrix.limited-dependencies }}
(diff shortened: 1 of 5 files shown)


shaunagm:
This should close #1138 and #1066 

Would love review on [this workaround](https://github.com/move-coop/parsons/pull/1192/files#diff-3e744f163053a3e58aeec8dede7e3711d8a87b2249a31fd94a884200b0aec1c7L34) - the tests are passing now but I'm not an installation expert and want to double check I'm not missing anything

parsons/etl/etl.py
```diff
@@ -659,20 +658,12 @@ def unpack_nested_columns_as_rows(self, column, key="id", expand_original=False)
...
-                        lambda row: hashlib.md5(  # nosec B324
-                            str.encode("".join([str(x) for x in row]))
-                        ).hexdigest(),
-                    )
+                orig.add_column(
+                    "uid",
+                    lambda row: hashlib.md5(
+                        str.encode("".join([str(x) for x in row])), usedforsecurity=False
+                    ).hexdigest(),
+                )
```
austinweisgrau, line 666: IDK what this section of code is, but these changes seem out of scope of this PR?

.github/workflows/python-checks.yml
```diff
@@ -160,6 +164,10 @@
                   python-version: ${{ matrix.python-version }}
                   cache: pip
 
+            - name: Patch install error when using Python 3.9, limited dependencies, and MacOS
+              if: ${{ matrix.limited-dependencies }} == True and ${{ matrix.os }} == "macos-latest" and ${{ matrix.python-version }} == "3.9"
+              run: pip install --system psycopg2-binary==2.9.9
+
```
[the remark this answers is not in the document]
austinweisgrau: this seems solid and makes sense to me

parsons/etl/etl.py
austinweisgrau: ohh I see, it's checking for the python version being 3.9 or higher

austinweisgrau (approved):
i like it

PR #1212, opened 2024-12-12
Allow ActionNetwork upsert_person to not change mobile opt-in status

austinweisgrau:
Prior behavior opted-in all phone numbers by default, and the only alternative would be to opt-out phone numbers sent to the API. This update allows for not changing the opt-in status of the number, and changes the default from opt-in to no change. This is a more reasonable and legally compliant default behavior than opting-in.

This causes a change in behavior! If users are currently relying on the default opt-in behavior of this connector's upsert_person method, this change will break their workflow. However, it is generally not legally compliant to opt-in by default, and more likely is that many users are unknowingly setting opt-in status to "subscribed" by default in a legally non-compliant way. The new configuration will require users to explicitly pass opt-in status if they want that status to be changed in ActionNetwork.

### DIFF
--- a/parsons/action_network/action_network.py
+++ b/parsons/action_network/action_network.py
@@ -2,7 +2,7 @@ import json
 import logging
 import re
 import warnings
-from typing import Dict, List, Union
+from typing import Dict, List, Union, Literal
 from parsons import Table
 from parsons.utilities import check_env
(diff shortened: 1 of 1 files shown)


matthewkrausse (approved):
This all looks good to me.

shaunagm (approved):
approving on behalf of MK

PR #1213, opened 2024-12-13
add method and test webinar report

sharinetmc:
Add get_webinar_report method to zoom connector

### DIFF
--- a/parsons/zoom/zoom.py
+++ b/parsons/zoom/zoom.py
@@ -271,9 +271,10 @@ class Zoom:
                 See :ref:`parsons-table` for output options.
         """
 
-        tbl = self._get_request(f"report/webinars/{webinar_id}")
-        logger.info(f"Retrieved {tbl.num_rows} webinar report.")
-        return tbl
+        dic = self._get_request(f"report/webinars/{webinar_id}")
+        if dic:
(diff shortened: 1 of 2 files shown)


parsons/zoom/zoom.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, account_id=None, client_id=None, client_secret=None):
             authorization_kwargs={"account_id": self.account_id},
         )
 
-    def _get_request(self, endpoint, data_key, params=None, **kwargs):
+    def _get_request(self, endpoint, data_key=None, params=None, **kwargs):
```
IanRFerguson, line 43: what's the rationale behind defaulting this to `None` type?

KasiaHinkson (approved):
LGTM

IanRFerguson (approved):
It looks like there's a formatting check with `ruff` that's failing - I would recommend looking into it to get it resolved but not existential imo

parsons/zoom/zoom.py
sharinetmc: This method that I have returns data that doesn't need to be parsed further using a key.

sharinetmc: But I guess it doesn't necessarily need to be a default value

PR #1227, opened 2024-12-20
Add newmode v2

sharinetmc:
Creates a separate class for NewMode's v2 API since the existing v1 (and the python wrapper the existing NewMode sync is based on will be sunset nest February)

@shaunagm, definitely hoping to discuss Parsons deprecation process with you! Also, wondering if you think v2 should just go in its own folder?
Discussed with @willyraedy some options to go about this. This PR currently follows the 2nd option

1) so one way to do this is with versions so we publish a new version and people have to upgrade to the version of the package to get the new and only the new NewMode connector
2) the other way to do it is have latest have both and then do a deprecation process where like:
NewMode and NewModeV2 are both available and people can update their code
All users swap to NewModeV2
Then you switch NewMode to be V2 as well so both NewMode and NewModeV2 are the v2 version
All users swap back to  NewMode
Delete NewModeV2


parsons/newmode/newmode.py
```diff
@@ -22,6 +22,7 @@ def __init__(self, api_user=None, api_password=None, api_version=None):
         Returns:
             Newmode class
         """
+        logger.warning("Newmode V1 API will be sunset in Feburary 2025.")
```
sharinetmc, line 25: definitely need more robust language here, but am hoping to discuss with Shauna along with the Parsons deprecation/versioning process

shaunagm:
Will the NewMode v1 API stop working in February?

parsons/newmode/newmode.py
shaunagm: This sounds pretty good, actually, but once we figure out the approach we might want to provide guidance here.

shaunagm:
Rather than creating a new connector with a different name that would have to be edited in lots of different scripts, can we set some kind of variable, either on the connector or at the global config level, that directs which connector to use under the hood?

I'm imagining something like:

```
class OldNewMode:
   ...

class NewNewMode:
  ...

if os.environ("UPDATE_NEWMODE_VERSION", "False"):
  class NewMode = NewNewMode
else:
  class NewMode = OldNewMode
```

So then people wouldn't need to change their scripts at all, just use an environmental variable, and then come Feb we just get rid of the OldNewMode entirely. Unless the old version will still be workable for a while...

I'm not 100% sure the above approach will work but it seems like it should?

parsons/__init__.py
```diff
@@ -73,6 +73,7 @@
     ("parsons.mobilize_america.ma", "MobilizeAmerica"),
     ("parsons.nation_builder.nation_builder", "NationBuilder"),
     ("parsons.newmode.newmode", "Newmode"),
+    ("parsons.newmode.newmode_v2", "NewmodeV2"),
```
shaunagm, line 76: See my comment in the main thread for thoughts on overall design of the deprecation process

parsons/newmode/newmode_v2.py
```diff
@@ -0,0 +1,237 @@
...
+        self.client.validate_response(response)
+        if response.status_code in success_codes:
+            if self.client.json_check(response):
+                response_json = response.json()
+                return response_json[data_key] if data_key else response_json
+            else:
+                return response.status_code
+        return response
+
+    def converted_request(
```
shaunagm, line 66: This method needs a docstring, I'll wait for you to add it before asking questions about the method as I imagine some of my questions will be answered by the docstring

parsons/newmode/newmode_v2.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def get_campaign(self, campaign_id, params={}):
+        """
+        Retrieve a specific campaign by ID.
+
+        In v2, a campaign is equivalent to Tools or Actions in V1.
+        `Args:`
+            campaign_id: str
+                The ID of the campaign to retrieve.
+            params: dict
```
shaunagm, line 111: If this is easy/straightforward to do it would be great to include a link here to newmode's docs to show what params are allowed and how to format them (same for all methods that accept `params`)

parsons/newmode/newmode_v2.py
```diff
@@ -0,0 +1,237 @@
...
+            organization_id: str
+                ID of organization
+            params: dict
+                Query parameters to include in the request.
+        `Returns:`
+            List containing all campaign ids.
+        """
+        self.base_url = API_CAMPAIGNS_URL
+        self.api_version = "jsonapi"
+        self.headers = {
```
shaunagm, line 138: It's a little surprising to see headers defined in a get method

shaunagm (changes requested):
Not sure if you wanted me to review the whole thing but I took a quick look, obviously the main thing is the deprecation process - happy to talk about this on a call if you'd prefer

sharinetmc:
> Will the NewMode v1 API stop working in February?

parsons/newmode/newmode_v2.py
sharinetmc: yes, this particular endpoint isn't actually available in the v2 API yet, but is available via another url, which requires unique header definitions. If you have any suggestions to define them in a better way, please lmk!

parsons/newmode/newmode_v2.py
sharinetmc: The docs for V2 aren't officially published yet and aren't entirely complete unfortunately, but will mark that as a TODO once the docs are published!

sharinetmc:
That sounds like an excellent idea, Shauna! Will make those changes!

parsons/newmode/newmode_v2.py
sharinetmc: done!

sharinetmc:
Marked this as ready for review @shaunagm, since I was able to get that last endpoint working! Happy to discuss versioning stuff further, though (specifically for how you think the best way to share with parsons community will be)!

parsons/newmode/newmode.py
```diff
@@ -323,3 +327,270 @@ def get_outreach(self, outreach_id, params={}):
...
+        method,
+        supports_version=True,
+        data=None,
+        json=None,
+        params={},
+        convert_to_table=True,
+        data_key=None,
+    ):
+        """Internal method to make a call to the Newmode API and convert the result to a Parsons table."""
+        self.client = OAuth2APIConnector(
```
willyraedy, line 400: this looks pretty slick

parsons/newmode/newmode.py
```diff
@@ -323,3 +327,270 @@ def get_outreach(self, outreach_id, params={}):
...
+        `Returns:`
+            Parsons Table containing campaign data.
+        """
+        endpoint = f"/campaign/{campaign_id}/form"
+        response = self.converted_request(
+            endpoint=endpoint,
+            method="GET",
+            params=params,
+        )
+        return response
```
willyraedy, line 445: nit: by calling the variable "response", I assume this is the full response. I'd call it "data" or even just do 
```suggestion
        return self.converted_request(
            endpoint=endpoint,
            method="GET",
            params=params,
        )
```

willyraedy (commented):
Great start! Love all the doc strings and there is clearly a lot of detail you've worked through on formatting the requests correctly. 

Big things to fix are (1) changing the state in the "get_campaign_ids" method and (2) a consistent return signature from the "converted_request" method. I'm gonna go look at the canales PR now to see how you may be handling the different return values in that script

parsons/newmode/newmode.py
```diff
@@ -323,3 +327,270 @@ def get_outreach(self, outreach_id, params={}):
...
+        else:
+            table = Table([data])
+
+        return table
+
+    def base_request(self, method, url, data=None, json=None, data_key=None, params={}):
+        """
+        Internal method to make a call to Newmode API and validate the response
+        """
+        response = None
```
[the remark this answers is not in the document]
sharinetmc: Without this line, if the api call fails for some reason, response will be left undefined

parsons/newmode/newmode.py
```diff
@@ -323,3 +327,270 @@ def get_outreach(self, outreach_id, params={}):
...
+        `Returns:`
+            Parsons Table containing submit data.
+        """
+
+        response = self.converted_request(
+            endpoint=f"campaign/{campaign_id}/submit",
+            method="POST",
+            data=data,
+            json=json,
+            params=params,
```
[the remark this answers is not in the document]
sharinetmc: I think we can do this once the Newmode docs are more robust. I currently don't know what the params are either.

parsons/newmode/newmode.py
```diff
@@ -323,3 +327,270 @@ def get_outreach(self, outreach_id, params={}):
...
+            "street_address": street_address,
+            "city": city,
+            "postal_code": postal_code,
+            "region": region,
+        }
+        if all(x is None for x in address_params.values()):
+            logger.error("Please specify a street address, city, postal code, and/or region.")
+            raise Exception("Incomplete Request")
+
+        params = {f"address[value][{key}]": value for key, value in address_params.items() if value}
```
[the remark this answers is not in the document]
sharinetmc: yes

parsons/newmode/newmode.py
```diff
@@ -323,3 +327,270 @@ def get_outreach(self, outreach_id, params={}):
...
+        self.headers = {
+            "content-type": "application/vnd.api+json",
+            "accept": "application/vnd.api+json",
+            "authorization": "Bearer 1234567890",
+        }
+        endpoint = "node/action"
+        response = self.converted_request(
+            endpoint=endpoint, method="GET", params=params, data_key="data"
+        )
+        return response["id"]
```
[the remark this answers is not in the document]
sharinetmc: This is a weird endpoint. It's one that isn't exposed in the V2 API, but is availible in the newmode in-house api. Without this index, the actual returned value contains a lot of otherwise not very useful information that is not very fun to wrangle.

mirabuck:
> Will the NewMode v1 API stop working in February?

We expect to shut it down completely on February 28th, 2025.

parsons/newmode/newmode.py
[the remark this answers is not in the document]
willyraedy: @mirabuck would y'all be able to put up a follow up PR adding this once the accepted parameters are set?

parsons/newmode/newmode.py
```diff
@@ -350,40 +356,57 @@ def __init__(
...
-        return table
+        self.client = OAuth2APIConnector(
+            uri=base_uri,
+            auto_refresh_url=V2_API_AUTH_URL,
+            client_id=self.client_id,
+            client_secret=self.client_secret,
+            headers=headers,
+            token_url=V2_API_AUTH_URL,
+            grant_type="client_credentials",
+        )
```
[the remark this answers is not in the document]
willyraedy: we could even call it `self.default_client` if that would be clearer. and then in this function it would say something like

`client = client if client else self.default_client`

jburchard:
> I by red d

sharinetmc:
@shaunagm, i think i'm still waiting on your approval for this!

parsons/utilities/api_connector.py
```diff
@@ -306,3 +307,13 @@ def json_check(self, resp):
             return True
         except JSONDecodeError:
             return False
+
+    def convert_to_table(self, data):
+        """Internal method to create a Parsons table from a data element."""
+        table = None
+        if type(data) is list:
```
IanRFerguson, line 314: I think `isinstance(data, list)` is more conventional

parsons/newmode/newmode.py
```diff
@@ -22,14 +33,12 @@ def __init__(self, api_user=None, api_password=None, api_version=None):
         Returns:
             Newmode class
         """
+        logger.warning(
+            "Newmode V1 API will be sunset in Feburary 2025. To use V2, set api_version=v2.1"
+        )
```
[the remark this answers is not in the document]
IanRFerguson: I see the comment above, we're still ~a month out

IanRFerguson (commented):
minimal feedback from me! I think once @shaunagm signs off on this we can ship it

parsons/newmode/newmode.py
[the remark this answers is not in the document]
sharinetmc: oh yes, forgot to update this comment after that update.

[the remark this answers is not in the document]
sharinetmc: I put Feburary first bc I didn't know the exact date.

shaunagm (approved):
Looks good to me, sorry for the delay!

IanRFerguson (approved):
Nice work!

PR #1242, opened 2025-01-19
Consolidate and DRY up tests on DB Sync utility

austinweisgrau:
Consolidates duplicative code into a standardized TestDBSync abstract base class, which can then be inherited and run the same tests on the FakeDatabase, Postgres, and Redshift connectors. Technically this new setup extends the test surface a bit, especially since a full suite of tests now runs on the FakeDatabase in the CICD environment (without LIVE_TEST=1). Previously only a very limited set of tests ran on the FakeDatabase in the CICD environment.


test/test_databases/test_dbsync.py
```diff
@@ -9,332 +12,192 @@
...
+    def initialize_db_connections(self) -> None:
+        self.source_db = self.db()
+        self.destination_db = self.db()
 
-        teardown_sql = f"""
-                       DROP SCHEMA IF EXISTS {self.temp_schema} CASCADE;
-                       """
-        self.db.query(teardown_sql)
+    def tearDown(self):
+        if self.teardown_sql:
```
shaunagm, line 61: I'm curious when you *wouldn't* want to tear down the sql during a test? Making this conditional and not the default (since `teardown_sql` defaults to None) makes it likely that it will be run unintentionally - is the idea that it's better to accidentally not clean up after a test when you should than vice versa?  I don't do a ton of testing of databases, so I'm maybe missing something basic.

shaunagm: Okay nevermind, I see that the sql is only getting run in the live tests where setup_sql and teardown_sql are set as attributes.

shaunagm (commented):
Took a quick look and happy to approve whenever @cmdelrio says it's ready (or anyone else with appropriate knowledge - basically, don't let me be a blocker)

cmdelrio:
@shaunagm I don't think I'm well equipped to review this, unfortunately!

shaunagm:
Hmm @austinweisgrau do you have a sense of who would be a good person to tag in? Maybe @IanRFerguson? I just don't know enough about how the db syncs work to feel comfortable being the lead/only reviewer.

test/test_databases/test_dbsync.py
austinweisgrau: Ya, there are circumstances where it's not necessary to define SQL to run on setup & teardown, so that's why the conditional. e.g. the SQLite connector creates an entirely new database on setup, and doesn't have a concept of schemas, so there's no need for SQL to run on setup or teardown.

parsons/databases/db_sync.py
```diff
@@ -155,7 +155,13 @@ def table_sync_incremental(
                 "Destination tables %s does not exist, running a full sync",
                 destination_table,
             )
-            self.table_sync_full(source_table, destination_table, order_by=primary_key, **kwargs)
+            self.table_sync_full(
+                source_table,
+                destination_table,
+                order_by=primary_key,
+                verify_row_count=verify_row_count,
```
IanRFerguson, line 162: funny, was this just not getting utilized in the function previously?

test/test_databases/fakes.py
```diff
@@ -1,14 +1,22 @@
 from parsons.etl.table import Table
 import logging
+from parsons.databases.database_connector import DatabaseConnector
+from typing import Union, Optional
 
 logger = logging.getLogger(__name__)
 
 
-class FakeDatabase:
+class FakeDatabase(DatabaseConnector):
```
IanRFerguson, line 9: this is inheriting the abstract class, am I following?

test/test_databases/test_dbsync.py
```diff
@@ -9,332 +12,192 @@
...
         # Load dummy data to parsons tables
         self.table1 = Table.from_csv(f"{_dir}/test_data/sample_table_1.csv")
         self.table2 = Table.from_csv(f"{_dir}/test_data/sample_table_2.csv")
 
+        self.source_table = (
+            f"{self.temp_schema}.source_table" if self.temp_schema else "source_table"
+        )
+        self.destination_table = (
+            f"{self.temp_schema}.destination_table" if self.temp_schema else "destination_table"
+        )
```
IanRFerguson, line 45: dig it

test/test_databases/test_dbsync.py
```diff
@@ -9,332 +12,192 @@
...
-        self.fake_destination.setup_table("destination", destination_data)
-
-        dbsync.table_sync_incremental("source", "destination", "id")
-
-        destination = self.fake_destination.table("destination")
-
-        # Make sure the rest of the data came through
-        assert_matching_tables(source_data, destination.data)
+class TestFakeDBSync(TestDBSync):
+    db = FakeDatabase
```
IanRFerguson, line 142: I really like the object-oriented approach to mocking here, that's slick

IanRFerguson (approved):
To no one's surprise, this is excellent

Really appreciate the object-oriented approach to mocking, gonna have to dig into this more to see how all the tests interact but looks great

test/test_databases/fakes.py
austinweisgrau: yessir

austinweisgrau: which means now this FakeDatabase thingy gets the whole range of generic tests on it as well

parsons/databases/db_sync.py
austinweisgrau: I think I've made a few modifications to verify row count over time, it seems like there are a lot of places this flag modifies behavior and it's a bit inconsistent to what extent it does or doesn't get passed down the stack

austinweisgrau: In general mostly creating problems when there are 0 rows copied for a few reasons, including that the destination table may or may not exist and the verify row count implementation doesn't handle that well. alternative strategy could be to make that implementation more robust to comparing 0 row counts

PR #1243, opened 2025-01-19
Use a temporary docker postgres container for postgres DB sync tests

austinweisgrau:
Uses the testcontainers package to easily generate and use a temporary postgres docker container for tests. Allows for running tests on postgres without running a persistent server separately, which should enable us to run tests on postgres in the CICD environment.


bmos:
this is such a great idea!!
we might also be able to use this approach for other database packages...

bmos:
If you could run `ruff format` and `ruff check --fix` and push the changes, it'll re-run the tests. I'd be happy to help get it working.

austinweisgrau:
The MacOS test runner does not allow for running docker containers - I made a change in 9d64d41f1ee814a714d6f01aa9243192bbdef2ff to skip the tests involving a docker container if docker is not available in the environment

```
    try:
        import docker

        docker.from_env().ping()
    except Exception:
        pytest.skip("Docker not available")
```

austinweisgrau:
Turns out you can't run docker in a Windows github actions environment either, so I took a slightly more robust skipping approach in 651cb39b04387473b2041b3ac6599fdfde6d0ce2

```
    container = PostgresContainer("postgres:9.5")
    try:
        container.start()
    except Exception as e:
        pytest.skip(f"Could not start Postgres container: {e}")
```

requirements-dev.txt
```diff
@@ -11,6 +11,7 @@ requests-mock==1.12.1
 ruff==0.13.0
 testfixtures==8.3.0;python_version<'3.11'
 testfixtures==9.1.0;python_version>='3.11'
+testcontainers==4.9.0
```
bmos, line 14:
```suggestion
testcontainers==4.14.1
```

might as well use the latest version here

parsons/databases/postgres/postgres.py
```diff
@@ -32,14 +32,22 @@ class Postgres(PostgresCore, Alchemy, DatabaseConnector):
...
+        timeout: int = 10,
+    ) -> None:
         super().__init__()
 
         self.username = username or os.environ.get("PGUSER")
         self.password = password or os.environ.get("PGPASSWORD")
         self.host = host or os.environ.get("PGHOST")
         self.db = db or os.environ.get("PGDATABASE")
-        self.port = port or os.environ.get("PGPORT")
+        self.port = port or os.environ.get("PGPORT") or 5432
```
bmos, line 50: should 5432 be a string like in test_postgres.py?

PR #1244, opened 2025-01-19
Create a Sqlite connector utility

austinweisgrau:
Should follow #1242

Works with query(), copy(), and the DB sync utilities

### DIFF
--- a/docs/databases.rst
+++ b/docs/databases.rst
@@ -62,7 +62,7 @@ Quick Start
 Postgres
 ********
 
-Postgres is popular open source SQL database dialect. The Parsons class leverages the `mysql <https://www.psycopg.org/>`_ python package.
+Postgres is popular open source SQL database dialect. The Parsons class leverages the `psycopg2 <https://www.psycopg.org/>`_ python package.
 
 ===========
 Quick Start
--- a/docs/databases.rst
+++ b/docs/databases.rst
@@ -104,3 +104,37 @@ Redshift
 
 See :ref:`Redshift <redshift>` for documentation.
 
+.. _sqlite:
+
+********
+Sqlite
+********
+
+SQLite is a performant flat-file database that's often touted as the "zero-config" database. The Parsons class uses the python3 built-in sqlite3 connector.
+
+===========
+Quick Start
+===========
+
+**Authentication**
+
+.. code-block:: python
+
+   from parsons import Sqlite
+
+   # Instantiate Sqlite from passed variables
+   sqlite = Sqlite(db_path='local.db')
+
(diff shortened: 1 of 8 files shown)


austinweisgrau:
Yes, the tests in this PR use the refactored code in #1242 so this should only be merged after #1242 is approved and merged.

The tests are failing because I haven't figured out how to get sqlite properly installed and accessible in the github actions test environment, so that's a work in progress. (Tests should be passing on a local setup, it's a github actions specific issue. I think.)

bmos:
So this is to connect to a local SQL instance and SQL is not installed within the GitHub Actions environment?

I'd suggest an actions step of installing SQL via apt on Linux and winget on Windows. Not. Sure about macOS.

bmos:
That being said, you might not actually need SQL installed to test the connector properly. There might be a Pytest SQL plugin or something which acts like an SQL instance.

austinweisgrau:
The test actually is appropriately failing, I realized. Even though sqlite3 comes with python as part of the stdlib, on windows machines and probably other misc architectures, the sqlite3 cli utility isn't available by default.

One of the methods in the connector attempts to use the shell utility through a subprocess, but that won't work if the shell utility isn't installed and on path.

I'm going to modify that behavior to check if the shell utility is available first, and use a fallback behavior if it is not.

bmos:
You should only need to install sqlite for the pytest job, not ruff, bandit, et al

austinweisgrau:
yay

austinweisgrau:
cool security check @bmos , very helpful

shaunagm (changes requested):
This looks good to me although it might make sense to have someone who uses the database connectors more often (like @IanRFerguson or @sharinetmc or @KasiaHinkson) take a quick look to.

The only thing missing from my perspective is docs. The database utilities/connectors are documented in this file: https://github.com/move-coop/parsons/blob/main/docs/databases.rst

You can see what this looks like from the reader's perspective here:
https://move-coop.github.io/parsons/html/stable/databases.html

austinweisgrau:
> This looks good to me although it might make sense to have someone who uses the database connectors more often (like @IanRFerguson or @sharinetmc or @KasiaHinkson) take a quick look to.
> 
> The only thing missing from my perspective is docs. The database utilities/connectors are documented in this file: https://github.com/move-coop/parsons/blob/main/docs/databases.rst
> 
> You can see what this looks like from the reader's perspective here: https://move-coop.github.io/parsons/html/stable/databases.html

Done!

shaunagm (approved):
LGTM!

PR #1251, opened 2025-01-24
Add bigquery export table + get job methods

sharinetmc:
This adds the export_table_to_gcs and get_job methods to the Bigquery Connector.

### DIFF
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -396,7 +396,7 @@ class GoogleBigQuery(DatabaseConnector):
             compression=compression,
             field_delimiter=field_delimiter,
         )
-        self.client.extract_table(
+        return self.client.extract_table(
             source=table_name,
             destination_uris=destination_uris,
             location=location,
@@ -417,7 +417,7 @@ class GoogleBigQuery(DatabaseConnector):
                 Other arguments to pass to the underlying get_job
                 call on the BigQuery client.
         """
-        self.client.get_job(job_id=job_id, **job_kwargs)
+        return self.client.get_job(job_id=job_id, **job_kwargs)
 
     def copy_from_gcs(
         self,
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -350,61 +350,6 @@ class GoogleBigQuery(DatabaseConnector):
         """
         self.query(sql=queries_wrapped, parameters=parameters, return_values=False)
 
-    def export_table_to_gcs(
-        self,
-        table_name: str,
-        destination_uris: str,
-        location: str = "US",
-        destination_file_format: str = "CSV",
-        field_delimiter: str = ",",
-        compression: str = None,
-        job_config: dict = {},
-        **export_kwargs,
-    ) -> job.ExtractJob:
-        """
-        Export a bigquery table to GCS
-
-        `Args:`
-            table_name: str
-                The table name to export.
-            destination_uris: str
-                Uri glob indicating where the table will be exported to in GCS
-                (If the table exceeds 1 GB, the table will be split among multiple files)
-            destination_file_format: str
-                The file format/type of the resulting file in GCS.
-                Accepts "CSV", "NEWLINE_DELIMITED_JSON", "PARQUET", or "AVRO".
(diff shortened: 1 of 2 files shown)


parsons/google/google_bigquery.py
```diff
@@ -349,6 +350,75 @@ def query_with_transaction(self, queries, parameters=None):
         """
         self.query(sql=queries_wrapped, parameters=parameters, return_values=False)
 
+    def export_table_to_gcs(
+        self,
+        table_name: str,
+        destination_uris: str,
```
willyraedy, line 356: why is this plural but not accepting a list?

parsons/google/google_bigquery.py
```diff
@@ -349,6 +350,75 @@ def query_with_transaction(self, queries, parameters=None):
...
+        **export_kwargs,
+    ):
+        """
+        Export a bigquery table to GCS
+
+        `Args:`
+            table_name: str
+                The table name to export.
+            destination_uris: str
+                Uri indicating where the table will be exported to in GCS
```
willyraedy, line 371: sometimes a wildcard is required if a file is big right? we should put that in the docs

willyraedy (commented):
Just some nits and documentation thoughts!

parsons/google/google_bigquery.py
sharinetmc: yes, it's a string because it's a glob. that's how the param is defined in the source method as well

sharinetmc: can add that in the docs

parsons/google/google_bigquery.py
```diff
@@ -349,6 +350,78 @@ def query_with_transaction(self, queries, parameters=None):
...
+                job_config values are preferred.
+            **export_kwargs: kwargs
+                Other arguments to pass to the underlying extract_table
+                call on the BigQuery client.
+        """
+        job_config_default = ExtractJobConfig(
+            destination_format=destination_file_format,
+            compression=compression,
+            field_delimiter=field_delimiter,
+        )
```
IanRFerguson, line 399: this is a totally valid approach, just curious why you opted to create a new variable rather than overwrite the reference if it's not provided?

```
if not job_config:
  logger.info("Using default job config as none was provided...")
  job_config = ExtractJobConfig(...)
```

IanRFerguson: that simplifies the logic a bit on L404 too

parsons/google/google_bigquery.py
```diff
@@ -349,6 +350,78 @@ def query_with_transaction(self, queries, parameters=None):
...
+            location=location,
+            job_config=job_config if job_config else job_config_default,
+            **export_kwargs,
+        )
+
+    def get_job(
+        self, job_id: str, **job_kwargs
+    ) -> Union[job.LoadJob, job.CopyJob, job.ExtractJob, job.QueryJob, job.UnknownJob]:
+        """
+        Fetch a job
```
IanRFerguson, line 412: can you call out what the function returns here too? I think I'm following but it would be helpful to summarize what could possibly be returned from the `Union` that you've helpfully provided here

IanRFerguson (approved):
No substantial changes from me! Nice work

parsons/google/google_bigquery.py
sharinetmc: It could be any one of those depending on what kind of job the job_id is associated with

sharinetmc: This is how it appears in the official Google docs as well

IanRFerguson: I understand that, it would just be nice to have them written in plain English in the docstring if possible

bmos:
I find this very helpful to avoid having to go back and re-format your code when the test doesn't pass:

#### Pre-commit
To avoid submissions of code that doesn't pass our checks, we recommend you set your
computer to automatically run checks when committing code.
We use [pre-commit](https://pre-commit.com/) to automate these checks.
To configure this on your computer, run the following commands:
`pip install pre-commit`
`pre-commit install`
To ensure this is working, run the following commands:
`pre-commit run --all-files`

PR #1255, opened 2025-01-28
Use environment markers to handle more recent versions of python-civis

shaunagm:
Basically python-civis has much newer versions, which require newer joblib versions, but they don't run on python 3.9. This PR uses [environment markers](https://peps.python.org/pep-0508/#environment-markers) to specify the older versions on python 3.9 and newer versions on 3.10+

Address #1195 and #1221


shaunagm:
Okay, the markers appear to work. Although not when we do the limited dependencies approach? But those don't cause the tests to break? Well hopefully this will all get fixed/changed soon anyway.

PR #1258, opened 2025-01-28
Prefer float over int for bigquery type coercion

crystaljgong:
We encountered a situation where a parsons table had a column with `int`, `float` and `NoneType` values. Bigquery kept complaining about there being `float`s in an `int` column, so it seemed like the schema generation wasn't behaving properly. This fix specifically ensures that the schema chooses `float` when both `float` and `int` are present in the petl_types.


austinweisgrau (approved):
this works and is a very helpful bugfix :D

shaunagm:
yes, thank you @crystaljgong!

PR #1262, opened 2025-01-30
Zoom Connector Poll Endpoints Version 2

KasiaHinkson:
- Adds a version_2 flag to poll based endpoints the defaults to False for backwards compatibility
- Version 2 skips handling json columns, to preserve a pure ELT practice and allow TMC to get all available data and handle the nested JSON in BigQuery

### DIFF
--- a/parsons/zoom/zoom.py
+++ b/parsons/zoom/zoom.py
@@ -421,7 +421,7 @@ class Zoom:
 
         logger.info(f"Retrieved {tbl.num_rows} polls for meeting ID {meeting_id}")
 
-        return self.__handle_nested_json(table=tbl, column="questions")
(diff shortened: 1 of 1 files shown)


shaunagm:
Looks like the test failures are a deeper Parsons issues since all tests with parsons_limited_dependencies set to true are failing, and none without

bmos:
Since limited dependencies installs with the latest versions of most packages, it's a little hard to track this down.
I have noticed this on my own branches as well because I'm doing some work right now to add support for Python 3.13.

bmos:
See #1263

bmos:
The issue has been resolved, the braintree library was updated

PR #1263, opened 2025-01-31
Fix broken limited_dependencies with braintree 4.33.0

bmos:
The CI pytest workflows have been failing lately because braintree [released 4.33](https://github.com/braintree/braintree_python/compare/ad42db3bd1c68ddcf708c1fbdc91f701fb2befdd...c3886eab20fdd807c2c843509927a73e28a6a3b5) of the Braintree Python library which doesn't work with parsons. Until the braintree parsons integration gets updated for support, I've limited the install versions to v4-v4.32.X which should allow for the intended flexibility of LIMITED_DEPENDENCIES without allowing use of this known incompatible version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ def main():
             "alchemer": ["surveygizmo"],
             "azure": ["azure-storage-blob"],
             "box": ["boxsdk"],
-            "braintree": ["braintree"],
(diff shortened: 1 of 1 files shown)


bmos:
No longer needed!
https://github.com/braintree/braintree_python/blob/master/CHANGELOG.md#4331

PR #1264, opened 2025-01-31
Python 3.13 Support

### LINKED ISSUE #1167: Python 3.13 Support
Python 3.13 is now out!
https://docs.python.org/3.13/whatsnew/3.13.html

bmos:
Closes #1167

### DIFF
--- a/.github/workflows/python-checks.yml
+++ b/.github/workflows/python-checks.yml
@@ -2,9 +2,7 @@ name: Python checks
 
 on:
     push:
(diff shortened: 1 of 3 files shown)


bmos:
Re 3aa2332, it's unclear whether I should blacklist this specific version or use a max version cap (as I have done here)

shaunagm (approved):

shaunagm:
@bmos I just realized this should have included an update to our README which still says we only support "Python 3.9-12"

PR #1267, opened 2025-02-01
Add import sorting

bmos:
As per last year's [discussion](https://github.com/move-coop/parsons/pull/1132#issuecomment-2405570655) in #1132, here is a discrete PR implementing import sorting in the Github Actions check and pre-commit config.

See PEP #8 > Imports
https://peps.python.org/pep-0008/#imports

### DIFF
--- a/parsons/actblue/actblue.py
+++ b/parsons/actblue/actblue.py
@@ -1,8 +1,9 @@
 import logging
 import time
-from parsons.utilities.api_connector import APIConnector
-from parsons.utilities import check_env
+
 from parsons.etl import Table
+from parsons.utilities import check_env
+from parsons.utilities.api_connector import APIConnector
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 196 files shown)


austinweisgrau (changes requested):
I think we should use the ruff import sorter rather than bring isort in. Since we're already using ruff and ruff comes with an isort-like import sorter.

Ruff documentation on its import sorter: https://docs.astral.sh/ruff/faq/#how-does-ruffs-import-sorting-compare-to-isort

Ruff documentation on how to use its import sorter: https://docs.astral.sh/ruff/formatter/#sorting-imports

I don't see any conflicts between ruff and isort in the repo as-is, and can't find any description of this, but I know isort and black and have conflicting opinions about how to format imports, and I imagine that could be true for isort and ruff as well.

bmos:
How interesting, I had no idea that Ruff offered that.
I completely agree and I will rework this pull request this afternoon.

austinweisgrau (approved):
How elegant

bmos:
If we go ahead with merging this, I will issue one more commit to accept the re-sorted imports.

shaunagm (approved):
looks great!

PR #1268, opened 2025-02-01
Dev tool updates

bmos:
Notes:
ruff format now targets python 3.9+ since this is the lowest version we support.
ruff changes required some reformatting of files.
pycodestyle was in requirements-dev.txt but not used.
python-checks.yml workflow now runs on commits pushed to any branches, not just main and major-release.
pre-commit updated with new ruff version to match CI workflow (as this does not pull from requirements-dev.txt/ installed version)

### DIFF
--- a/.github/workflows/python-checks.yml
+++ b/.github/workflows/python-checks.yml
@@ -2,9 +2,7 @@ name: Python checks
 
 on:
     push:
-        branches: [ "main", "major-release" ]
     pull_request:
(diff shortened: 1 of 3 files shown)


bmos:
If we go ahead with merging this, I will issue one more commit, which accepts the reformatted multi-line F-strings from ruff 0.9.0+

shaunagm (approved):
looks great - and thanks for fixing my version naming error :)

PR #1279, opened 2025-02-03
Catalist + SFTP: Export Files from Server in Chunks

IanRFerguson:
This PR adds logic to optionally chunk the exporting of a flat file from Catalist's SFTP server to the local filesystem (though technically this will provide chunking functionality to any connection using the Parsons `SFTP` class)

### DIFF
--- a/parsons/sftp/sftp.py
+++ b/parsons/sftp/sftp.py
@@ -12,6 +12,9 @@ from parsons.utilities import files as file_utilities
 
 logger = logging.getLogger(__name__)
 
+# Default byte size to export under the hood via Paramiko
+DEFAULT_EXPORT_BYTE_SIZE = 1024 * 1024 * 50
+
 
 class SFTP(object):
     """
@@ -145,7 +148,7 @@ class SFTP(object):
         remote_path,
         local_path=None,
         connection=None,
-        export_chunk_size: Optional[int] = None,
+        export_chunk_size: int = DEFAULT_EXPORT_BYTE_SIZE,
     ):
         """
         Download a file from the SFTP server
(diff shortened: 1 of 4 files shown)


parsons/sftp/sftp.py
```diff
@@ -162,13 +174,75 @@ def get_file(self, remote_path, local_path=None, connection=None):
...
-                connection.get(remote_path, local_path)
+                if export_chunk_size:
+                    self.__get_file_in_chunks(
+                        remote_path=remote_path,
+                        local_path=local_path,
+                        connection=connection,
+                        export_chunk_size=export_chunk_size,
+                    )
+                else:
+                    connection.get(remote_path, local_path)
```
IanRFerguson, line 197: not a super sophisticated solution here ... could maybe envision a higher order function applying here but would love some thought partnership

parsons/sftp/sftp.py
```diff
@@ -162,13 +174,75 @@ def get_file(self, remote_path, local_path=None, connection=None):
...
+        """
+
+        logger.info(
+            f"Reading from {remote_path} to {local_path} in {export_chunk_size}B chunks"
+        )
+
+        with connection.open(remote_path, "rb") as _remote_file:
+            with open(local_path, "wb") as _local_file:
+                # This disables paramiko's prefetching behavior
+                _remote_file.set_pipelined(False)
```
IanRFerguson, line 230: thoughts on parameterizing this? my thought is that if we're attempting to download a table large enough to require chunking the export, it's probably worth disabling prefetching by default

in case it's helpful, prefetching in this case stores a queue of the requests that will be sent to the server in memory ... by disabling it before reading a chunk of data, we're only going to request one chunk at a time which should keep the memory footprint a bit lighter

IanRFerguson:
@austinweisgrau makes a great point about usability in a discussion offline - do folks think it makes sense to chunk the exports by default rather than utilizing the `connection.get_file()` approach at all? 

in other words, this PR assumes that the user is aware that their table is too large to export normally ... if we chunk by default they won't have to consider it

parsons/sftp/sftp.py
austinweisgrau: i think this looks good

parsons/sftp/sftp.py
austinweisgrau: i think this is reasonable, to disable prefetching by default when setting the chunk size. both are about improving the reliability of a sftp server struggling with oversized requests, so i think they go together well.

if in the future someone comes up with a use case where it matters to separate these two, would be easy to add that parameterization at that time, but I think given the known use-case we have, that is currently YAGNI

austinweisgrau (approved):
good to go 👍🏻

austinweisgrau:
beautimous

PR #1281, opened 2025-02-04
Kasiah/zoom single row tables

KasiaHinkson:
- If the results are a dict, it's not an empty table, it's a 1 row table
- _handle_nested_json was unpacking the first dict in a list of dicts, leaving the others unnested. Added a version 2 flag to avoid this behavior

### DIFF
--- a/parsons/zoom/zoom.py
+++ b/parsons/zoom/zoom.py
@@ -79,7 +79,9 @@ class Zoom:
                 data.extend(self.client.data_parse(r))
             return Table(data)
 
-    def __handle_nested_json(self, table: Table, column: str) -> Table:
+    def __handle_nested_json(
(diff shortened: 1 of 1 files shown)


IanRFerguson (commented):
I think `version: int = 1` as the default and then `if version == 2` as the logic check is a little cleaner

and is version the API version? or something else? not totally following

KasiaHinkson:
We're just trying to be backwards compatible and  after talking with Shauna and Sharine, we had decided that what made the most sense was a version 2 flag that would allow for different capability, and in the future we could deprecate the "version 1" code that isn't the behavior we want long term. I was really hoping to get this out the door, but clearly that isn't going to happen today, I'll update the member and refactor tomorrow morning. I'll update in Asana as well.

PR #1282, opened 2025-02-06
Kasiah/zoom single row tables

KasiaHinkson:
- Corrects handling of single rows responses
- Adds a version flag to allow users to skip the weird JSON logic that unpacks a single dict in a list but not the others

### DIFF
--- a/parsons/zoom/zoom.py
+++ b/parsons/zoom/zoom.py
@@ -79,7 +79,9 @@ class Zoom:
                 data.extend(self.client.data_parse(r))
             return Table(data)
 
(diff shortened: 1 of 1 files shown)


willyraedy (approved):
Interesting. I don't think I'm following all the bespoke data formatting. Might be a more substantial rewrite here but this looks fine as a work around

KasiaHinkson:
Yeah, it's a bit wonky because we're really trying not to introduce breaking changes. This connector absolutely needs a substantial rewrite, but there's no way it wouldn't be breaking for anyone currently using it.

PR #1284, opened 2025-02-10
Support kwargs for from_Csv

Carl4:
For some reason, from_csv encoding defaults to 'charmap' encoding on my machine, which fails when importing ActBlue data due to unrecognized characters. This quick patch allows me to specify encoding='utf-8'.


shaunagm:
Thanks so much for this PR @Carl4! The failing test is our ruff formatter complaining that in your docstring, `**csvargs:` has a trailing space so it's actually `**csvargs: `. Sorry to bother you for such a tiny thing, but would you mind updating the PR to remove that space? If not I can do it. And then we can merge this. :)

shaunagm:
Thanks so much for the contribution @Carl4 !

PR #1293, opened 2025-02-13
Use TableSchema in BigQuery Column List

IanRFerguson:
This fix patches an inefficiency in the `GoogleBigQuery.get_columns_list()` function ... writing a `SELECT * FROM {{ TABLE }} LIMIT 1` in BigQuery isn't recommended as the entire table will be scanned in the process. 

This PR updates the logic to leverage the underlying `schema` definition, rather than querying against the table.

### DIFF
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -1194,9 +1194,9 @@ class GoogleBigQuery(DatabaseConnector):
             A list of column names
         """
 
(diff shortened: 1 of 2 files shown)


IanRFerguson:
not sure why the AirMeet test isn't passing the `ruff` format check ... it wasn't passing locally either but appears to be now

image

austinweisgrau (approved):
nice catch

PR #1294, opened 2025-02-13
convert both list and dict columns to json

austinweisgrau:
In #1143, an update was made that enabled the BigQuery connector to automatically load dict columns as JSON strings rather than fail to load and error out.

List columns need the same fix - they cannot load by default into BigQuery and need to be converted to a JSON string. This PR makes that update.

### DIFF
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -794,7 +794,7 @@ class GoogleBigQuery(DatabaseConnector):
         quote: Optional[str] = None,
         schema: Optional[List[dict]] = None,
         max_timeout: int = 21600,
-        convert_dict_columns_to_json: bool = True,
+        convert_dict_list_columns_to_json: bool = True,
         **load_kwargs,
     ):
         """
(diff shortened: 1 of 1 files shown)


parsons/google/google_bigquery.py
```diff
@@ -794,7 +794,7 @@ def copy(
         quote: Optional[str] = None,
         schema: Optional[List[dict]] = None,
         max_timeout: int = 21600,
-        convert_dict_columns_to_json: bool = True,
```
shaunagm, line 797: This is a breaking change because it renames a parameter. I don't know that it will impact many people because the parameter was only added four months ago - it might be that you're the only one using it - and if we are going to play fast and loose with the API I guess now's the time to do it. But alternatively you could just add a second parameter.

shaunagm: All right I'm going to go ahead and merge this as is, hopefully it's not an issue for folks.

shaunagm (approved):

Issue #1180, opened 2024-10-30
parsons[VAN] : Add NGP Introspection Endpoint


engelhartrueben:
I recently have come across this tool https://docs.ngpvan.com/reference/introspection by NGP VAN that has helped us check both the connection and the key permissions. It has been helpful when checking if a key has access to MyCampaigns, but not MyVoters, for instance. 

## Detailed Description

Add Introspection call to Parsons API. This method would take in the username and API Key, handle the conversion to base64, and make the introspection call to VAN. The response would return the JSON object, regardless of success or failure. 

## Context

This is important because it allows for a quick check of connection and permissions within a given application.

## Possible Implementation

The script on the VAN Documentation suffices for my local setup (with some tweaks), but am not super familiar with the this repository.
```python
import requests

url = "https://api.securevan.com/v4/apiKeyProfiles"

headers = {
     "accept": "application/json",
     "authorization": "Basic ****some base64 string ==****"
}

response = requests.get(url, headers=headers)

print(response.text)
```
Response:
```JSON
{
    "items": [
        {
            "databaseName": "SmartVAN Massachusetts",
            "hasMyVoters": true,
            "hasMyCampaign": true,
            "committeeName": "People for Good",
            "apiKeyTypeName": "Custom Integration",
            "keyReference": "1234",
            "userFirstName": "peopleforgood",
            "userLastName": "api",
            "username": "peopleforgood.api",
          	"userId": 4321
        }
    ],
    "nextPageLink": null,
    "count": 1
}
```

## Priority

Low priority, but helpful none the less (:

austinweisgrau:
Oh nice yeah this would be cool

PR #1295, opened 2025-02-14
Add Introspection class for API key profiles with tests

matthewkrausse:
Introduce an Introspection class to handle API key profiles and implement corresponding unit tests to validate functionality.

Issue #1180

### DIFF
--- a/parsons/ngpvan/introspection.py
+++ b/parsons/ngpvan/introspection.py
@@ -1,0 +1,21 @@
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class Introspection(object):
+    def __init__(self, van_connection):
+        self.connection = van_connection
+
+    def get_apikeyprofiles(self):
+        """
+        Get API key profiles.
+
+        `Args:`
+            None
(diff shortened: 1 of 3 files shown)


shaunagm (commented):
Can you tell me a bit about why you wanted to make an introspection class for this, as opposed to just calling `.get_request("apiKeyProfiles")` on the connection itself? Also you import and instantiate the logger but it doesn't seem to be used?

matthewkrausse:
> Can you tell me a bit about why you wanted to make an introspection class for this, as opposed to just calling `.get_request("apiKeyProfiles")` on the connection itself? Also you import and instantiate the logger but it doesn't seem to be used?

I created a new class because every endpoint used in the VAN connector is organized in a separate class. The logger was imported in every other class so just when I copied another class it came with it. I can add a simple log statement that confirms that a response came through.

shaunagm (approved):
Thanks for the explanation, @matthewkrausse - I'd forgotten that fact about the NGPVan connector! And thanks for adding the logging.

PR #1309, opened 2025-02-21
Catalist Match - Expose polling parameter

IanRFerguson:
Gives us the option to parameterize the polling parameter in Catalist's match function.


austinweisgrau (approved):
🫡

PR #1323, opened 2025-03-01
Add ruff Implicit String Concatenation Rules

bmos:
Built on top of #1324, this PR adds flake8-implicit-string-concat rules which ask contributors to use implicit multi-line string concatenation rather than explicit multi-line concatenation.

It also contains code changes to resolve all resulting errors.
No changes to code behavior.


austinweisgrau (approved):
ohh i like this one

bmos:
One weird quirk about this pull request is that it seems GitHub's counts of lines of code counts explicitly concatenated strings as a single line and implicitly concatenated strings as multiple lines. That's why this pull request shows that multiple lines of code are being added when technically it's just a single line of code being added.

PR #1324, opened 2025-03-01
Enable flake8-2020 linting rules

bmos:
Enable [flake8-2020](https://docs.astral.sh/ruff/rules/#flake8-2020-ytt) linting rules

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,7 @@ select = [
     "Q", # flake8-quotes (Q)
     "TC", # flake8-type-checking (TC)
     "YTT", # flake8-2020 (YTT)
+    "ISC", # flake8-implicit-str-concat (ISC)
 ]
 ignore = [
     "E501", # line-too-long (E501)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ select = [
     "TID", # flake8-tidy-imports (TID)
     "ICN", # flake8-import-conventions (ICN)
     "Q", # flake8-quotes (Q)
+    "TC", # flake8-type-checking (TC)
 ]
 ignore = [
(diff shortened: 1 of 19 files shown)


pyproject.toml
austinweisgrau, line 1: these documenting comments are super helpful

test/test_utilities.py
```diff
@@ -20,7 +19,9 @@
         pytest.param("2018-12-13", 1544659200),
         pytest.param("2018-12-13T00:00:00-08:00", 1544688000),
         pytest.param("", None),
-        pytest.param("2018-12-13 PST", None, marks=[xfail_value_error]),
+        pytest.param(
+            "2018-12-13 PST", None, marks=[pytest.mark.xfail(raises=ValueError, strict=True)]
+        ),
```
austinweisgrau, line 24: this is a nice simplification. was it enforced by one of the new ruff rules? cool if so.

austinweisgrau (approved):
Are the new rules fully redundant with the ruff defaults, or did we just happen to follow all of them through luck / good style? The latter seems doubtful to me. Maybe they're redundant with black, if not ruff, and got taken care of with the old black formatting configs? IDK

test/test_utilities.py
bmos: If I remember correctly, the linter flagged something else about this code, and then when I was looking at it for simplification, I realized it just didn't need to exist.

bmos:
I think it's a mix of all three of those, actually.

shaunagm (approved):

shaunagm:
@bmos can you take a quick look at this and either resolve the merge conflict or tell me how I should resolve it?

bmos:
Updated title since all but one commit were included in #1323

PR #1325, opened 2025-03-03
Add ruff flake8-bugbear rules

bmos:
See here
https://docs.astral.sh/ruff/rules/#flake8-bugbear-b

Includes code changes to get tests passing

### DIFF
--- a/parsons/utilities/api_connector.py
+++ b/parsons/utilities/api_connector.py
@@ -108,9 +108,7 @@ class APIConnector(object):
         else:
             raise RuntimeError(f"{return_format} is not a valid format, change to json or content")
 
-    def post_request(
-        self, url, params=None, data=None, json=None, success_codes=[200, 201, 202, 204]
-    ):
+    def post_request(self, url, params=None, data=None, json=None, success_codes=None):
         """
         Make a POST request.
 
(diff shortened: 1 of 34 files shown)


bmos:
I can't see the dockercloud logs to know why it failed...
It isn't failing on main and it was merging main into this branch that triggered it to fail...

parsons/utilities/api_connector.py
```diff
@@ -108,9 +108,7 @@ def get_request(self, url, params=None, return_format="json"):
         else:
             raise RuntimeError(f"{return_format} is not a valid format, change to json or content")
 
-    def post_request(
-        self, url, params=None, data=None, json=None, success_codes=[200, 201, 202, 204]
-    ):
```
shaunagm, line 113: It would be great if the specific success codes used were adding to the docstring

parsons/utilities/api_connector.py
```diff
@@ -142,7 +142,7 @@ def post_request(
             else:
                 return r.status_code
 
-    def delete_request(self, url, params=None, success_codes=[200, 201, 204]):
+    def delete_request(self, url, params=None, success_codes=None):
         """
         Make a DELETE request.
```
shaunagm, line 148: Same as above - adding the codes to the docstring

PR #1335, opened 2025-03-06
Requirements consolidation

### LINKED ISSUE #1118: Combine requirements-dev.txt and docs/requirements.txt
Currently we have three requirements files, the standard one in requirements.txt, a requirements-dev.txt in the top level, and another requirements.txt in the docs folder. The latter has requirements for building docs, and requirements-dev.txt has requirements for testing and linting. There's no reason not to combine the latter two (probably makes sense to remove the requirements.txt in the docs folder).

When doing this, make sure to check our CI workflows - presumably this changes how the docs are automatically built, for instance. We also want to check any of our documentation, like the Contributing Guide, and update instructions there as well.

bmos:
Closes #1118

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,3 +9,11 @@ pytest==8.3.4
 requests-mock==1.12.1
 ruff==0.9.4
 testfixtures==8.3.0
+
+# Docs Requirements
+Sphinx==8.1.3;python_version>='3.10'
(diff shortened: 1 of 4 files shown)


requirements-dev.txt
bmos, line 1: Sphinx and myst-parser versions we are using don't support python 3.9.
Allowing last version that did support them.

austinweisgrau (approved):
nice!

Issue #1341, opened 2025-03-11
[Bug] Pip sometimes finds an incorrect parsons build (0.18.1)


austinweisgrau:
Sometimes `pip install parsons` will result in pip attempting (and failing) to install a parsons package 0.18.1. Attempting to pin a parsons version above 0.18.1 will result in pip failing to find any matching parsons distribution.

I don't understand this issue and cannot reliably reproduce it, but have seen it come up sporadically. I don't understand pypi distributions so I'm not sure what's going on here.

## Detailed Description

As documented in an unrelated issue #1339, here is an example traceback of pip installing parsons
```
Run python -m pip install --upgrade pip
Requirement already satisfied: pip in /opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages (25.0.1)
Collecting prefect (from -r prefect/requirements.txt (line 1))
Downloading prefect-3.2.12-py3-none-any.whl.metadata (12 kB)
Collecting prefect-gcp (from -r prefect/requirements.txt (line 2))
Downloading prefect_gcp-0.6.4-py3-none-any.whl.metadata (2.7 kB)
Collecting parsons[google] (from -r prefect/requirements.txt (line 3))
Downloading parsons-0.18.1.tar.gz (273 kB)
Installing build dependencies: started
Installing build dependencies: finished with status 'done'
Getting requirements to build wheel: started
Getting requirements to build wheel: finished with status 'error'
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> [21 lines of output]
Traceback (most recent call last):
File "/opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 389, in
main()
File "/opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 373, in main
json_out["return_val"] = hook(**hook_input["kwargs"])
^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line [14](https://github.com/matthewkrausse/ggtx/actions/runs/13778503395/job/38532336420#step:4:15)3, in get_requires_for_build_wheel
[...]

PR #1342, opened 2025-03-11
Build + Release Workflow

bmos:
To get this set up, a new  GitHub Actions environment called "release" should be created.
This environment should be connected as a Trusted Publisher via pypi: https://docs.pypi.org/trusted-publishers/
I'm thinking re-implementing the build + pypi process should let us fix/render obsolete #1341


fgregg:
do you even want there to be wheel distributions? 

`setup.py` [has conditional logic to allow for modular imports if an environmental variables is set](#632). In order to take advantage of that, `setup.py` has to exist, and it won't in the wheel distributions.

fgregg:
this will not address #1341 because of the reasons given in https://github.com/move-coop/parsons/issues/1341#issuecomment-2758894132

bmos:
https://docs.astral.sh/uv/guides/package

Looks like maybe we can use uv for this?

shaunagm:
Okay, trusted publishing is set up and pointing at a currently-not-existent-in-main `release.yml` file. I think that means if we merge this, and then cut a new release, it will automatically get pushed to PyPI. Does that sound right? Are we missing anything?

Edit: also set up an environment named "release" for running release.yml in and configured it so a reviewer needs to approve it to run.

shaunagm (approved):

PR #1345, opened 2025-03-12
clean up signature for VAN initialization

fgregg:
This PR cleans up the signature and docstring for the VAN initialization code

* Remove arguments from docstring that should not be passed.
* I also had to adjust the .pre-commit.yaml and pyproject.toml files to get bandit to not complain about asserts in tests (where there should definitely be asserts)


parsons/ngpvan/van.py
```diff
@@ -54,19 +55,18 @@ class VAN(
...
     `Returns:`
         VAN object
     """
 
-    def __init__(self, api_key=None, auth_name="default", db=None, raise_for_status=True):
+    def __init__(
+        self,
+        api_key: Optional[str] = None,
+        db: Optional[Literal["MyVoters", "MyCampaign", "MyMembers", "EveryAction"]] = None,
+        **_,  # ignore any other keywords (previously this signature included other keywords)
```
fgregg, line 68: this is backwards compatible in that if people have code where they did set `auth_name` or `raise_for_status`, it won't throw an exception.

it could be better:
1. remove it entirely, and let the code throw an exception since if people were passing those arguments it wasn't doing anything, and they might want to be alerted to do that. it would also keep this signature clean.
2. not have a catchall here, but explicitly check for `auth_name` and `raise_for_status` and raise a warning if they are set. this would make the signature kind of messy, but it would avoid the catchall.

austinweisgrau: Personally I think I'd rather see that argument removed, I think it's more confusing and kind of a footgun to have an unused kwargs. I think losing backwards compatibility with arguments that were already fake is kind of OK

austinweisgrau: But yeah probably "best practice" would be to leave the arguments in place and issue a deprecationwarning if they're included, and then remove them in the next major release. 🤷🏻

austinweisgrau (changes requested):
This change is great, good catch, thank you! I'd love to see another solution besides an unused kwargs.

parsons/ngpvan/van.py
austinweisgrau: but yeah I vote option (1)

fgregg:
oh, just saw your comment. I'll just go with option 1 then.

austinweisgrau:
Can you explain the changes you're making to CICD / bandit? They're out of scope and should probably be their own PR, altho thats not super important if theyre relatively small changes

.pre-commit-config.yaml
```diff
@@ -1,10 +1,12 @@
...
     hooks:
       - id: ruff
       - id: ruff-format
   - repo: https://github.com/PyCQA/bandit
-    rev: '1.8.3'
+    rev: 1.8.3
     hooks:
-    - id: bandit
+      - id: bandit
+        args: ["-c", "pyproject.toml"]
```
fgregg, line 11: this and the next line are necessary for pre-commit to have bandid use pyproject.toml for config. see the comment on pyproject.toml for why we need it.

fgregg:
@austinweisgrau, comments added.

austinweisgrau (approved):

parsons/ngpvan/van.py
shaunagm: Do we know what the previous use of `auth_name` or `raise_for_status` was? What's the likelihood we'll catch people out on this?

@sharinetmc @KasiaHinkson any way you can take a quick look and see if this would cause issues within TMC?

fgregg: they were unused in the function body.

* `raise_for_status` hasn't done anything six years https://github.com/move-coop/parsons/pull/52/files#diff-ee3aa4f1f607ee3259fa9751ca3bf2b3ebdcced0c31d5722a09928744d36ee84R40
* and `auth_name` has never done anything (from the initial commit) https://github.com/move-coop/parsons/blame/fd9a816fdac0e3636cb0290da35322f4ae2925e7/parsons/ngpvan/van.py#L42

shaunagm: okay, in that case it seems unlikely we'll break anything by making this change (I'll still warn folks in the release notes though)

fgregg: for sure! if people were passing it because they thought it did something, then this PR will lead to the code throwing an exception!

shaunagm:
@fgregg can you resolve the conflicts in the pre-commit-config file? that's the only thing blocking merging, and I don't understand what the commands are doing well enough to resolve them myself

PR #1356, opened 2025-03-18
Add background_processing option to tagging create and delete methods

sjwmoveon:


parsons/action_network/action_network.py
```diff
@@ -1220,7 +1220,7 @@ def upsert_person(
                 status of the phone in ActionNetwork. If not included, won't update the status.
                 None by default, causes no updates to mobile number status. New numbers are set
                 to "unsubscribed" by default.
-            background_request: bool
```
shaunagm, line 1223: I don't love that we're changing the API here - is there any chance users would have `background_request` in their code and making this parameter name change would break things for them?

shaunagm (approved):

PR #1359, opened 2025-03-20
Add ruff flake8-use-pathlib rules

bmos:
https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -32,7 +32,7 @@ jobs:
             # DO NOT run actions/checkout here, for security reasons
             # For details, refer to https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
             - name: Post comment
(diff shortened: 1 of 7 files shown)


parsons/utilities/zip_archive.py
```diff
@@ -53,4 +53,4 @@ def unzip_archive(archive_path, destination=None):
     with zipfile.ZipFile(archive_path, "r") as z:
         file_name = z.namelist()[0]
         z.extractall(path=destination)
-        return os.path.join(destination, file_name)
+        return str(Path(destination) / file_name)
```
shaunagm, line 56: What's happening here? What the `/` doing? That errors for me when I try to recreate in my interpreter

shaunagm: Okay this is actually throughout the PR, this is just the first spot I noticed it. It looks like it specifies positional only arguments? https://peps.python.org/pep-0570/

But does str take arguments? When I try to use this syntax on my interpreter ie `str("cat" / "dog")` it fails

PR #1363, opened 2025-03-20
copy local data to bigquery directly, without staging

fgregg:
The BigQuery API supports uploading a local file directly without staging it to cloud storage. This can have the benefit of reducing the permissions needed by the IAM user and may be faster? <s>Under some circumstances, there may be advantages to a staged upload, though I can't find a clear discussion of what they may be.</s> According to @austinweisgrau, larger data will need to be staged through CloudStorage.

This PR add as `copy_direct` method that uploads local data directly to BigQuery. 

For the `copy` method, if none of the information needed for a staged upload is present, it adds a reference to the `copy_direct` method in existing error message.


austinweisgrau:
Using the cloud storage step is MUCH faster for any table above a certain relatively small threshold. Above a different threshold, it's probably impractical/impossible to load a table to BigQuery without that intermediary step. BigQuery is highly optimized to loading data from Google Cloud Storage.

I do think it's fine to enable an option to load directly without cloud storage step, but just FYI it's non-optimal in most cases.

fgregg:
@austinweisgrau swapped the code so that there is a new copy_direct method and the existing copy method functions unchanged except for a reference to the new copy_direct method in an exception.

however, if the staged copy is usually the better way to do things, then i think we should not increase the surface area of the API with a `copy_direct` method, and instead just expand on the documentation and explain why the staged approach is the right one to go with?

austinweisgrau:
That makes sense to me! Especially, it could make sense as a default behavior when tmp_gcs_bucket isn't set as you did at first. But yes, the documentation should explain that its not ideal to rely on that for larger datasets.

fgregg:
sorry, i'm not quite clear on your preference.

1. bring in `copy_direct`
2. make not changes to the `copy` except additional documentation on why the stages method is the right default
3. don't expose `copy_direct` as a public method but if the necessary settings do not exist for a staged upload fall back to direct upload along with a warning?

shaunagm:
It sounds like it makes sense to bring in `copy_direct` and reference it in exceptions in `copy`. I'd be open to changing the default behavior of `copy` when the gsc bucket isn't set but that's more of a change so happy to keep it as just a hint in the exception for now. 

@fgregg this PR changes the pyproject.toml and the pre-commit-config, is that intentional?

fgregg:
the pyproject.toml and .precommit-config.yaml file were changed to deal with local pre-commit issues that have since been resolved on main.

i think this PR is ready for your review.

shaunagm (approved):
Looks good to me! I appreciate that this also refactors out some helper methods - not a blocker to merging this PR but it might be nice to have some tests to cover those helper methods (for the stringify method at least it looks straightforward - sometimes the testing juice isn't worth the squeeze in parsons when methods rely on a ton on external connections)

PR #1371, opened 2025-03-23
Add Avro file support in Table class

matthewkrausse:
Introduce functionality for reading and writing Avro files in the Table class, including methods for basic operations, schema handling, and support for various compression codecs.

### DIFF
--- a/parsons/etl/tofrom.py
+++ b/parsons/etl/tofrom.py
@@ -98,6 +98,138 @@ class ToFrom(object):
 
         return local_path
 
+    def to_avro(
+        self, target, schema=None, sample=9, codec="deflate", compression_level=None, **avro_args
+    ):
+        """
+        Outputs table to an Avro file.
+
+        Write the table into a new avro file according to schema passed.
+
+        This method assume that each column has values with the same type
+        for all rows of the source `table`.
+
+        `Apache Avro`_ is a data
+        serialization framework. It is used in data serialization (especially in
+        Hadoop ecosystem), for dataexchange for databases (Redshift) and RPC
+        protocols (like in Kafka). It has libraries to support many languages and
+        generally is faster and safer than text formats like Json, XML or CSV.
+
+        `Args:`
+            target: str
+                the file path for creating the avro file.
+                Note that if a file already exists at the given location, it will be
+                overwritten.
+            schema: dict
+                defines the rows field structure of the file.
+                Check fastavro `documentation`_ and Avro schema `reference`_ for details.
+            sample: int, optional
+                defines how many rows are inspected
+                for discovering the field types and building a schema for the avro file
+                when the `schema` argument is not passed. Default is 9.
+            codec: str, optional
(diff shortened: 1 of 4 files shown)


matthewkrausse:
Looks like I am running into an issue with the limited dependencies. fastavro should be a dependency of the petl package but they are not being loaded in I guess. Wanted to chat through this before just adding fastavro to the limited dependency set which seems wrong. @austinweisgrau Have any thoughts here?

austinweisgrau:
For the limited dependencies setup, you need to add the dependency in the setup.py file. You'll want to think about if it should be an "extra" (so if people want to be able to use that functionality, they need to install parsons with `parsons[avro]` or if it should auto-bundled with parsons for everyone, even in the friendly dependencies setup.

By default, it should probably be the former (need to opt-in with parsons[avro]).

matthewkrausse:
Thanks. Made what I believe will fix it. Let's find out.

austinweisgrau:
> Thanks. Made what I believe will fix it. Let's find out.

yep that's it!

matthewkrausse:
@shaunagm This is ready for your review

parsons/etl/tofrom.py
```diff
@@ -98,6 +98,142 @@ def to_html(
...
+            ...     'namespace': 'test',
+            ...     'type': 'record',
+            ...     'fields': [
+            ...         {'name': 'name', 'type': 'string'},
+            ...         {'name': 'friends', 'type': 'int'},
+            ...         {'name': 'age', 'type': 'int'},
+            ...     ]
+            ... }
+            ...
+            >>> # now demonstrate what writing with toavro()
```
shaunagm, line 179: extra word "what" in there

parsons/etl/tofrom.py
```diff
@@ -98,6 +98,142 @@ def to_html(
...
+            >>> import parsons import Table
```
shaunagm, line 180: `import Parsons import Table` seems like it would be a syntax error? can you put them on two lines? I know these are just docs but people might copy and paste

parsons/etl/tofrom.py
```diff
@@ -98,6 +98,142 @@ def to_html(
...
+
+        In order to use this method, you must have the `fastavro` library installed.
+        If using limited dependencies, you can install it with `pip install parsons[avro]`.
+
+        Write the table into a new avro file according to schema passed.
+
+        This method assume that each column has values with the same type
+        for all rows of the source `table`.
+
+        `Apache Avro`_ is a data
```
shaunagm, line 115: I see a lot of links syntax with no corresponding links - can you add them? (Or are you using RST in a way I'm unfamiliar with?) 

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links

shaunagm (changes requested):
Couple minor docs changes requested but other than that this looks good

shaunagm (changes requested):
just updating the status to reflect that there's still a couple changes needed (a few links without urls still)

PR #1381, opened 2025-03-25
Add pre-commit test to ci

bmos:
Since there have been a few issues with our pre-commit workflow lately I wrote a CI test which should help catch them before they hit main in the future.

How it works:
Since there isn't a good way to test pre-commit independently of the tests it runs, I have configured this only to run when ruff check, ruff format, and bandit ci jobs all pass. If they pass, then the pre-commit workflow will run. If it fails, there must be an issue with pre-commit.


austinweisgrau (approved):
i like it

PR #1393, opened 2025-03-30
Replace-petl-with-pandas-dataframe

matthewkrausse:


shaunagm:
Is your goal to make it so people can use parsons with pandas *or* petl, or to swap out petl for pandas? If it's the latter we should have more of a discussion with the community as a whole before you get too far with doing this work

austinweisgrau:
This is a non-starter unless we can incorporate some pandas extension that allows for lazy loading, which I think as of [most recent discussion](https://tmc-parsons.slack.com/archives/C018CQD3G4E/p1765931732698739) we're not sure exists. The pandas API is more familiar to most of us and that would be nicer than petl to work with, but we need lazy loading and pandas doesn't offer that.

Lazy loading is a critical feature for Parsons as it allows for working with arbitrarily large files in the Table idiom. pandas dataframes are fully loaded into memory and that imposes some severe limits on how large the underlying data can be. I think we can close this PR unless there's a proposal here for how this would work.

matthewkrausse:
I am fine to close. I was really just exploring this as a possibility but I don't think we should do it.

Issue #1400, opened 2025-03-31
Clarify how GCS Temp Bucket is referenced in Google Bigquery Documenation


matthewkrausse:
The copy() relies on passing a gcs_temp_bucket but the copy() method wants tmp_gcs_bucket but we set gcs_temp_bucket as the variable name that is passed. This confused me when writing some other documentation. I am going to create a PR to standardize the variable in the docs.

```
dataset = 'parsons_dataset'
table = 'parsons_table'

# Table name should be project.dataset.table, or dataset.table, if
# working with the default project
table_name = f"`{project}.{dataset}.{table}`"

# Must be pre-existing bucket. Create via GoogleCloudStorage() or
# at https://console.cloud.google.com/storage/create-bucket. May be
# omitted if the name of the bucket is specified in environment
# variable GCS_TEMP_BUCKET.
gcs_temp_bucket = 'parsons_bucket'

# Create dataset if it doesn't already exist
bigquery.client.create_dataset(dataset=dataset, exists_ok=True)

parsons_table = Table([{'name':'Bob', 'party':'D'},
                       {'name':'Jane', 'party':'D'},
                       {'name':'Sue', 'party':'R'},
                       {'name':'Bill', 'party':'I'}])

# Copy table in to create new BigQuery table
bigquery.copy(
   table_obj=parsons_table,
   table_name=table_name,
   tmp_gcs_bucket=gcs_temp_bucket
)
```

PR #1401, opened 2025-03-31
Rename variable for consistency in documentation

matthewkrausse:
Rename the variable `gcs_temp_bucket` to `tmp_gcs_bucket` in the documentation to enhance consistency.

I would like to update the actual code but it would be a breaking change. We name the env var it expects one thing and the variable another but they are similar enough that it is an easy thing to miss.


matthewkrausse:
This handles #1400

PR #1410, opened 2025-04-07
Pass object as JSON, not params, to create message POST request

sjwmoveon:
The default second argument for [the API connector's POST request](https://github.com/move-coop/parsons/blob/main/parsons/utilities/api_connector.py#L111) is params, the request parameters.  We should be passing the message as part of the body of the POST, not as parameters.

### DIFF
--- a/parsons/action_network/action_network.py
+++ b/parsons/action_network/action_network.py
@@ -906,7 +906,7 @@ class ActionNetwork(object):
         `Documentation Reference`:
             https://actionnetwork.org/docs/v2/messages
         """
-        return self.api.post_request("messages", payload)
+        return self.api.post_request("messages", json=payload)
 
     def update_message(self, message_id, payload):
         """


technicalex (approved):
Edit: Oops, I reviewed the wrong pull request. Sorry about that! Going to try to remove this.

parsons/action_network/action_network.py
```diff
@@ -906,7 +906,7 @@ def create_message(self, payload):
         `Documentation Reference`:
             https://actionnetwork.org/docs/v2/messages
         """
-        return self.api.post_request("messages", payload)
+        return self.api.post_request("messages", json=payload)
```
shaunagm, line 909: I support this change but I just want to be sure we don't introduce errors for people. The docstring for this method says the payload should be a dict (presumably a Python dict, not a JSON dict) and I'm concerned that if someone passed in a Python dict as a payload it would cause an error.

Can you confirm that this doesn't break things for people passing in Python dicts? Or perhaps that was already broken? As long as this is not a breaking change I'm happy to approve + merge this.

shaunagm (changes requested):

parsons/action_network/action_network.py
sjwmoveon: I believe python dicts were already broken here.

shaunagm: Okay, so maybe the thing to do is update the docstring to specify that the payload being passed in should be a json dict, not a Python dict? Or we could check if it's a json dict and convert to json if it's just a python dict.

sjwmoveon: I think it's fine as a Python dict, actually.  See the [API connector documentation](https://github.com/move-coop/parsons/blob/9be3fd2488d3755579834af29bb0a1caa8cdf5e1/parsons/utilities/api_connector.py#L60):
> json: dict
> The payload of the request object. By using json, it will automatically serialize the dictionary

shaunagm: Okay great! Will approve and merge this and the other PR.

shaunagm (approved):

PR #1411, opened 2025-04-08
Kasiah/GitHub return value

KasiaHinkson:
This method didn't actually return the table, adding a return value.


IanRFerguson (approved):
funny, looks like a typo in the original source code. I'm surprised our unit tests didn't catch that

PR #1412, opened 2025-04-08
Update message API needs JSON payload

sjwmoveon:


parsons/action_network/action_network.py
```diff
@@ -926,7 +926,7 @@ def update_message(self, message_id, payload):
         `Documentation Reference`:
             https://actionnetwork.org/docs/v2/messages
         """
-        return self.api.put_request(f"messages/{message_id}", payload)
+        return self.api.put_request(f"messages/{message_id}", json=payload)
```
shaunagm, line 929: This looks like a similar change to the other PR, and so I have the same question - is there any chance that changing the param to JSON will cause things to break if people are passing in Python dicts instead of json like the docstring suggests

sjwmoveon: No, I believe Python dicts were already broken.

shaunagm (approved):

PR #1421, opened 2025-04-24
add client_credential auth method for salesforce

fgregg:
this PR adds the ability to connect with the [client_credential method](https://help.salesforce.com/s/articleView?id=sf.remoteaccess_oauth_client_credentials_flow.htm&language=en_US&type=5) to salesforce.

this is a much safer auth method than the password flow. 

there are a number of other auth methods, so we might want to take a beat and think about a design that can grow more gracefully.

### DIFF
--- a/parsons/salesforce/salesforce.py
+++ b/parsons/salesforce/salesforce.py
@@ -12,7 +12,7 @@ class Salesforce:
     """
     Instantiate the Salesforce class
 
-    Supports the passwor and client_credentials authentication_methods
+    Supports the password and client_credentials authentication_methods
 
     `Args:`
         username: str
@@ -261,6 +261,6 @@ class Salesforce:
                     domain=self.domain,
                 )
             else:
-                raise Exception("Shoudl not be possible to reach this code")
+                raise Exception("Should not be possible to reach this code")
 
         return self._client
--- a/parsons/salesforce/salesforce.py
+++ b/parsons/salesforce/salesforce.py
@@ -12,33 +12,72 @@ class Salesforce:
     """
     Instantiate the Salesforce class
 
+    Supports the passwor and client_credentials authentication_methods
+
     `Args:`
         username: str
             The Salesforce username (usually an email address). Not required if
-            ``SALESFORCE_USERNAME`` env variable is passed.
+            ``SALESFORCE_USERNAME`` env variable is passed. Used in the 'password' auth method.
         password: str
             The Salesforce password. Not required if ``SALESFORCE_PASSWORD`` env variable is
-            passed.
+            passed. Used in the 'password' auth method.
         security_token: str
             The Salesforce security token that can be acquired or reset in
             Settings > My Personal Information > Reset My Security Token.
-            Not required if ``SALESFORCE_SECURITY_TOKEN`` env variable is passed.
+            Not required if ``SALESFORCE_SECURITY_TOKEN`` env variable is passed. Used in the 'password' auth method.
         test_environment: bool
             If ``True`` the client will connect to a Salesforce sandbox instance. Not required if
             ``SALESFORCE_DOMAIN`` env variable is passed.
+        consumer_key: str
+            consumer key for a connected app. Used in the 'client_credentials' auth method.
+        consumer_secret: str
+            consumer secret for a connected app. Used in the 'client_credentials' auth method.
+        domain: str
+            url for the salesforce instance. Used in the 'client_credentials' auth method
+        authentication_method: str
+            the method to use for authentication. defaults to "password". Not required if ``SALESFORCE_AUTHENTICATION_METHOD`` env variable is passed.
+
     `Returns:`
         Salesforce class
     """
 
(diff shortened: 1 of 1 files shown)


fgregg:
the ruff failure is on code i did not touch. would you like me to resolve it anyway?

fgregg:
the ruff failure is unrelated to any changes i have made. do you want me to address it anyway?

parsons/salesforce/salesforce.py
```diff
@@ -12,33 +12,72 @@ class Salesforce:
     """
     Instantiate the Salesforce class
 
+    Supports the password and client_credentials authentication_methods
```
shaunagm, line 15: I think the underscore between "authentication" and "methods" is a typo?

Is there public-facing documentation about how to get `client_credentials` auth info? Like do you have to request a key? If so we should update the docs page for the salesforce connector and/or put a link to that page here.

parsons/salesforce/salesforce.py
```diff
@@ -12,33 +12,72 @@ class Salesforce:
...
+        security_token=None,
+        test_environment=False,
+        consumer_key=None,
+        consumer_secret=None,
+        domain=None,
+        authentication_method="password",
+    ):
+        try:
+            self.authentication_method = check_env.check("SALESFORCE_AUTHENTICATION_METHOD", None)
+        except KeyError:
```
shaunagm, line 57: Why might we get a keyerror here? I'm a little confused about how this block of code is doing what it's doing (I assume the goal is 'decide which method to use, based on env variable/input parameter and raise an error if neither is set')

parsons/salesforce/salesforce.py
```diff
@@ -12,33 +12,72 @@ class Salesforce:
...
+        elif self.authentication_method == "client_credentials":
+            self.consumer_key = check_env.check("SALESFORCE_CONSUMER_KEY", consumer_key)
+            self.consumer_secret = check_env.check("SALESFORCE_CONSUMER_SECRET", consumer_key)
+            self.domain = check_env.check("SALESFORCE_DOMAIN", domain)
 
-        if test_environment:
-            self.domain = check_env.check("SALESFORCE_DOMAIN", "test")
         else:
-            self.domain = None
+            raise ValueError(
```
shaunagm, line 78: If we're catching the `authentication_method = None` case here do we need the try/except above?

You might be able to just do `self.authentication_method = check_env.check("SALESFORCE_AUTHENTICATION_METHOD", authentication_method)` above without catching anything? But I could also be misunderstanding what you're trying to do

shaunagm (changes requested):
Overall this looks good, just have a minor docs request and then a bit of confusion over some of the code

parsons/salesforce/salesforce.py
```diff
@@ -12,33 +13,71 @@ class Salesforce:
     """
     Instantiate the Salesforce class
 
+    Supports the password and `client_credentials <https://help.salesforce.com/s/articleView?id=xcloud.connected_app_client_credentials_setup.htm&type=5>`_ authentication methods.
```
fgregg, line 16: added a link to the client credential docs in salesforce.

parsons/salesforce/salesforce.py
```diff
@@ -12,33 +13,71 @@ class Salesforce:
...
+        consumer_secret=None,
+        domain=None,
+        authentication_method=None,
+    ):
+        if authentication_method:
+            self.authentication_method = authentication_method
+        elif env_authentication_method := os.environ.get("SALESFORCE_AUTHENTICATION_METHOD"):
+            self.authentication_method = env_authentication_method
+        else:
+            self.authentication_method = "password"
```
fgregg, line 61: hopefully this logic is a lot clearer.

parsons/salesforce/salesforce.py
```diff
@@ -12,33 +13,71 @@ class Salesforce:
...
+        elif self.authentication_method == "client_credentials":
+            self.consumer_key = check_env.check("SALESFORCE_CONSUMER_KEY", consumer_key)
+            self.consumer_secret = check_env.check("SALESFORCE_CONSUMER_SECRET", consumer_key)
+            self.domain = check_env.check("SALESFORCE_DOMAIN", domain)
 
-        if test_environment:
-            self.domain = check_env.check("SALESFORCE_DOMAIN", "test")
         else:
-            self.domain = None
+            raise NotImplementedError(
```
fgregg, line 78: this will be raise when self.authentication_method is a string, but not `client_credentials` or `password`

parsons/salesforce/salesforce.py
fgregg: thanks! fixed type and added link!

parsons/salesforce/salesforce.py
fgregg: simplifed this a lot.

shaunagm (approved):

shaunagm:
Thanks so much @fgregg

PR #1431, opened 2025-04-30
Eventually Fail Actblue

codybraun:
For context: we've had some issues with ActBlue CSV exports for which the export runs indefinitely without any sort of failure message. I have requested to them that they eventually add a new status that will indicate when these have failed and we should give up, but in the meantime it would be nice not to have jobs that run infinitely long.

This PR:
1.) Makes it so that if AB do eventually add a new status like "error" we will respect it and quit retrying
2.) Adds an optional maximum number of retries to the ActBlue endpoint so that we can eventually abandon hope and raise an error (rather than having a process that runs forever)

Firsts Parsons PR for me, so let me know what I do wrong here! Would a PR template be helpful for me to add?

### DIFF
--- a/parsons/actblue/actblue.py
+++ b/parsons/actblue/actblue.py
@@ -27,7 +27,10 @@ class ActBlue(object):
                 https://secure.actblue.com/api/v1. You can set an ``ACTBLUE_URI`` env variable or
                 use this URI parameter if a different endpoint is necessary - for example, when
                 running this code in a test environment where you don't want to hit the actual API.
-
+            max_retries: int
+                The maximum number of times to poll the API for a download URL. Not required, default
+                is None, which means it will poll indefinitely until a download URL is returned.
+                ```ACTBLUE_MAX_RETRIES``` env variable can be set, which will override this parameter.
         For instructions on how to generate a Client UUID and Client Secret set,
         visit https://secure.actblue.com/docs/csv_api#authentication.
     """
@@ -134,7 +137,7 @@ class ActBlue(object):
             tries += 1
 
         if download_url is None:
-            raise TimeoutError("CSV generation timed out. Increase retries and try again.")
+            raise TimeoutError("CSV generation timed out. Increase max_retries and try again.")
 
         logger.info("Completed data generation.")
         logger.info("Beginning conversion to Parsons Table.")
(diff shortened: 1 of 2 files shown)


parsons/actblue/actblue.py
```diff
@@ -32,7 +32,13 @@ class ActBlue(object):
         visit https://secure.actblue.com/docs/csv_api#authentication.
     """
 
-    def __init__(self, actblue_client_uuid=None, actblue_client_secret=None, actblue_uri=None):
+    def __init__(
+        self,
+        actblue_client_uuid=None,
+        actblue_client_secret=None,
+        actblue_uri=None,
+        max_retries=None,
```
shaunagm, line 40: Can you add the new `max_retries` parameter to the docstring for the class?

parsons/actblue/actblue.py
```diff
@@ -46,6 +52,8 @@ def __init__(self, actblue_client_uuid=None, actblue_client_secret=None, actblue
             auth=(self.actblue_client_uuid, self.actblue_client_secret),
             headers=self.headers,
         )
+        self.max_retries = check_env.check("ACTBLUE_MAX_RETRIES", max_retries, optional=True)
```
shaunagm, line 55: In the docstring you're adding, worth mentioning that we'll check for an environmental variable `ACTBLUE_MAX_RETRIES` and if we find it set, it overrides the `max_retries` passed in.

parsons/actblue/actblue.py
```diff
@@ -117,9 +127,14 @@ def poll_for_download_url(self, csv_id):
...
         download_url = None
-        while download_url is None:
+        tries = 0
+        while download_url is None and (self.max_retries is None or tries < self.max_retries):
             download_url = self.get_download_url(csv_id)
             time.sleep(POLLING_DELAY)
+            tries += 1
+
+        if download_url is None:
+            raise TimeoutError("CSV generation timed out. Increase retries and try again.")
```
shaunagm, line 137: While you're editing things might as well swap "increase retries" to "increase max_retries" as that's the name of the input parameter

shaunagm (changes requested):
This looks great, thank you! A couple minor documentation changes but otherwise this looks good to merge

shaunagm:
And yes @codybraun a PR template would be lovely!

codybraun:
all good calls @shaunagm - will open a separate small PR with a PR template

shaunagm (approved):

shaunagm:
Looks great - thanks so much for the PR and welcome to Parsons! :D

PR #1439, opened 2025-05-06
Add a PR template

codybraun:
I always think this is helpful and encourages folks to leave better structured, richer explanations of their PRs. Happy to change this, are there are other items that should be included here so that folks remember to discuss them?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,5 +5,5 @@
 ## Considerations for discussion
 - (List out any significant design decisions that were made and why.)
 
-## How to test the changes
+## How to test the changes (if needed)
 - (How should a reviewer test this functionality)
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
+## What is this change?
+- (List out the changes.)
+- (Link to any relevant Github issues or Slack discussion)
+
+## Considerations for discussion
+- (List out any significant design decisions that were made and why.)
+
+## How to test the changes
+- (How should a reviewer test this functionality)
```
shaunagm, line 9: Sometimes tests are not needed (ie for most docs changes) so maybe add an 'if possible/needed' to that part of the template?

Other than that, this looks good!

shaunagm (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
shaunagm: looks great - thanks so much for doing this!

PR #1440, opened 2025-05-11
move project definitions to pyproject.toml

bmos:
My previous pass at this in #1015 got stuck because it removed the LIMITED_DEPENDENCIES / friendly dependencies feature and I was not confident enough to reimplement it, so I closed it.

Here is a more limited scope-approach that leaves the requirements.txt, requirements-dev.txt, and setup.py files handing dependencies, but moves project data and package selection to pyproject.toml

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,45 @@
+[build-system]
+requires = [ "setuptools" ]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "parsons"
+version = "4.0.0"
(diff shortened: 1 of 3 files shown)


shaunagm:
@bmos can you resolve the conflict, and then I can merge this PR? I'm 90% sure I know how to resolve the conflict correctly but not 100% so I'd rather have you do it.

bmos:
assuming the checks pass, this should be all set :)

PR #1445, opened 2025-05-14
Allow range of dependency versions

bmos:
## What's Changed
- Alphabetized dependencies
- Renamed test workflow job to pytest since other jobs used tool name (ruff, bandit, etc)
- Added pytest workflow matrix for installing oldest-supported versions of dependencies and integrated limited-deps testing into same matrix for more readable github action outputs
- Set min versions for some additional dependencies that wouldn't install (likely issues with upstreams being too permissive)
- Set all direct dependencies to allow a range of versions until the next major release
- Bumped ruff to 0.11.9
- Fixed a few miscellaneous typing issues
- Set coverage report to only post to console when pytest passes (to avoid having to scroll past it to see failures and errors)
- Changed all test_*.py files to import classes from parsons rather than from subfiles to avoid circular imports and test things as they will be used.

### DIFF
--- a/.github/workflows/python-checks.yml
+++ b/.github/workflows/python-checks.yml
@@ -14,9 +14,10 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
-                os: ["ubuntu-latest", "windows-latest", "macos-latest"]
                 limited-dependencies: ["", "TRUE"]
+                os: ["ubuntu-latest", "windows-latest", "macos-latest"]
+                python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+                resolution: ["highest", "lowest"]
 
         runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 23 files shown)


bmos:
I tend to agree (and would generally just set this to a recent version of each with the next minor version as the max), but wanted to demonstrate the diverse range of dependency versions which pass all tests. Most likely some connectors also remain broken in this PR, as some connectors are untested or only have live tests.

bmos:
Regarding the pip install times, uv installs it very quickly in a brand new venv.

bmos:
I took the changes that weren't necessary to the primary purpose of this PR and moved them to new PRs

austinweisgrau:
It seems like parsons limited dependencies/extras doesn't work with uv for some reason, which makes me more partial to this PR, although it's still a bit indirect. Seems like we need a comprehensive fix for dependency management. But yeah in the meantime this PR would be helpful for using uv with parsons.

bmos:
That is so strange because we do test limited dependencies in our GitHub Actions workflow...
What happens when you try and use it locally?

fgregg:
what about ditching the requirements file entirely and doing something like this in setup.py

```python
def main():

    install_requires = [
        "petl",
        ...,
        "simplejson",
    ]
    extras_require = {
        "airtable": ["pyairtable"],
        ...,
        "twilio": ["twilio"],
    }
    extras_require["all"] = sorted({lib for libs in extras_require.values() for lib in libs})
    
    limited_deps = os.environ.get("PARSONS_LIMITED_DEPENDENCIES", "")
    if limited_deps.strip().upper() not in ("1", "YES", "TRUE", "ON"):
        install_requires.extend(extras_require['all'])
        extras_require = {"all": []}

```

that way you are only needing to manage dependencies in one place?

bmos:
> what about ditching the requirements file entirely and doing something like this in setup.py
> [...](https://github.com/move-coop/parsons/pull/1445/#issuecomment-3046246244)

Personally, I'm very hesitant to use that approach.
There are two common approaches I see for requirements (pyproject.toml and requirements.txt) and that's neither of them.

I do like the functionality though!

fgregg:
Right now, the two ways to install *all* the dependencies will install different sets of dependencies.

`PARSONS_LIMITED_DEPENDENCIES pip install parsons[all]`

and

`pip install parsons`

will not install the same exact set of requirements, because `PARSONS_LIMITED_DEPENDENCIES pip install parsons[all]` will install the dependencies specified in the extras_required block of setup.py, and `pip install parsons` will install the dependencies specified in the requirements.txt file.

this does not seem good. 

if we are very, very careful to keep the requirements.txt file up and the setup.py extras_require up to date we may be able to have the two ways of install all the dependencies install the same, but that seems actually very hard.

I agree that our setup.py is strange, and I look forward to world where we don't need to support the dynamic behavior of PARSONS_LIMITED_DEPENDENCIES.  I see my suggested change as a step torwards that.

When `pip install parsons[all]` is the default, we will only have pyproject.toml and no setup.py or requirements.txt. We can't ditch setup.py yet because we need the dynamic behavior of PARSONS_LIMITED_DEPENDENCIES, but we can go ahead and ditch requirements.txt now.

austinweisgrau:
I do think it is more idiomatic for something like the PARSONS_LIMITED_DEPENDENCIES behavior to be the default, and users always need to specify extras or [all] if they want them. This can all be achieved in pyproject.toml rather than setup.py or requirements.txt. Here's an example: https://github.com/pypa/setuptools/discussions/3627#discussioncomment-6476654

PR #1460, opened 2025-05-27
Fix link in convert docstring

anzelpwj:
I found that original link for this was no longer working, returning a 404 error.


shaunagm (approved):
nice catch

Issue #1456, opened 2025-05-23
Design for APIConnector Clas


fgregg:
As per the discussion at the 2025-05-22 Parsons Contributor Meeting, I'm going to work on the APIConnector class. I will create a new class object that will sit along side the existing APIConnector Class, and we can switch the wrappers over to use new class to learn about the design. This will allow for freedom in class design. If successful, we will eventually want to go through and switch all the wrappers, but that's another discussion for another day.

One thing that came from that discussion was that there was two layers for what we would want and APIConnector to do.

1. Supporting the interactions with any **site**
2. Supporting the interactions with all or many **methods** of particular site.

For example, for any site, we want to make and receive HTTP requests, communicate errors, and retry on transient errors.

For a particular site, the pagination method for across endpoints will usually be the same, and it would be good to implement that in connector class for that site.

I would split out the list of the requested features in #736 as follows:

| All Site | All Methods in a Site |
|-|-|
| retry logic           | polling |
| default headers  | pagination |
| | unwrapping data |
| | snakecase to cameCase |

This all suggests a two level design, a request Session subclass for the "all sites" methods, and an APIConnnector that uses that subclass.

## Caveats
Of course, in these clean levels don't stay separated. It's not uncommon, for example for a site to return improper status codes, such as a 200 code with a message saying "no record found." This kind of issues can be handled by providing hooks in the Session subclass to rewrite the status code of a response based upon the body or header of the response.

shaunagm:
@fgregg it looks like your description cuts off? @ "The simplest thing that could work is start with preserving the"

I'm not sure I understand your distinction between "all site" and "all methods in site" - how are for example retries vs polling different here, in terms of how you're categorizing them?

Issue #736, opened 2022-08-30
[Refactor] Improve ulitity of the APIConnector class


Jason94:
The `APIConnector` class is a largely internal-to-Parsons class that is supposed to make writing connectors easier. Querying endpoints and parsing the results is a task most connectors need to do repeatedly. Making this easier allows the Parsons community to develop new connectors faster and maintain existing connectors with less effort. This is the goal of the `APIConnector` class, but it doesn't succeed in this yet.

## Detailed Description
The API Connector does not do much to actually make using API's easier. Here are a few examples:

If you create a connector that has a `data_key`, the APIConnector does not unwrap that data for you when you make a request:
``` python
client = APIConnector(my_url, data_key='data')

# Note how you still need to grab the data out of the 'data' key
data = client.get_request('orders')['data']
# There is a built in method in the client, but it's actually more typing to use
data = client.data_parse(client.get_request('orders'))
```

The `pagination_key` is also mostly unused.

If an API returns a payload like:
``` json
{
    data: { ... <data here> ... },
    next: <url you should call to get next 20 results>
}
```

then querying the API with your `APIConnector` like this:
``` python
client = APIConnector(my_url, pagination_key='next')
data = client.get_request('orders')
```
still only returns the first 20 results.

## Possible Implementation
([See notes for the 8/25/2022 contributor meeting](https://docs.google.com/document/d/1ovkdOJhBjpaC3sRzZwFFDvlfgYw1vuP_HsPiIdzYtPk/edit#)) Since `APIConnector` is used by most of the connectors in Parsons, we need to be careful to not break existing connectors while we refactor the class. This is the plan we came up with:

Create an initial PR doing the actual refactor:
1. Copy the existing class into a new `APIConnector` class in another module.
2. Refactor the new version of the class, leaving the old version in place. This way existing connectors won't be affected.
3. Submit for review and get feedback on the made changes and other potential improvements.

When this PR is merged, we can create more PR's migrating each of the existing connectors to the new `APIConnector.`

When all of the connectors are migrated, we can delete the old `APIConnector`.

## Priority
This is not time sensitive. But the sooner we update it, the fewer connectors will get written with the old version and need to be re-written.

shaunagm:
This is a pretty trivial note, but whoever refactors the class should probably be aware: [the APIConnector docs](https://github.com/move-coop/parsons/blob/main/parsons/utilities/api_connector.py#L23) say that `auth` should be a dictionary but it's actually a tuple.

[...]

PR #1481, opened 2025-06-19
WIP: API Connector Refactor

fgregg:
This is an experimental refactor of the API Connector class, applied to the NGP Van code.

It adds
- automatic retries
- header management

through using the [scraplib](https://github.com/jamesturk/scrapelib)'s scraper as a drop in replacement for the request session object.

this makes the API connector class much simpler. really just handling URI building and setting up headers. pagination and data unrolling are moved to the van connector class.

despite this PR not being marked as a draft, it should not be merged until more discussion. it is not in draft mode right now in order to allow for comments.

relates to #1456 and #736

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -33,6 +33,7 @@ PyGitHub==2.6.0
 python-dateutil==2.9.0.post0
 requests==2.32.3
 requests_oauthlib==2.0.0
+scrapelib==2.4.1
 setuptools==80.9.0
 simple-salesforce==1.12.6
 simplejson==3.20.1
@@ -51,4 +52,3 @@ jinja2>=3.0.2
 selenium==3.141.0
 us==3.2.0
 sshtunnel==0.4.0
-


parsons/etl/table.py
```diff
@@ -50,6 +51,10 @@ def __init__(
     ):
         self.table = None
 
+        #
+        if inspect.isgenerator(lst):
```
fgregg, line 55: it's very convenient to write pagination code as a generator as i've done in this PR, and imo parsons should support that. https://github.com/move-coop/parsons/issues/1529

parsons/ngpvan/van_connector.py
```diff
@@ -1,46 +1,32 @@
...
-from parsons.utilities import check_env
-from parsons.utilities.api_connector import APIConnector
+from parsons.utilities.api_connector_next import APIConnector
 
 logger = logging.getLogger(__name__)
 
-URI = "https://api.securevan.com/v4/"
-SOAP_URI = "https://api.securevan.com/Services/V3/ListService.asmx?WSDL"
 
+class VANConnector(APIConnector):
```
fgregg, line 9: the van connector class is the one that knows about pagination and data unboxing.

parsons/ngpvan/van.py
```diff
@@ -66,9 +68,20 @@ def __init__(
...
+        else:
+            raise KeyError(
+                "Invalid database type specified. Pick one of:"
+                " MyVoters, MyCampaign, MyMembers, EveryAction."
+            )
+
+        session = Scraper()
+        session.auth = ("default", api_key + "|" + str(self.db_code))
+
+        self.connection = VANConnector(session=session, uri="https://api.securevan.com/v4/")
```
fgregg, line 84: with this pattern we don't need to keep a handle of the api key in this class.

parsons/ngpvan/van.py
```diff
@@ -66,9 +68,20 @@ def __init__(
...
+            self.db_code = 0
+        elif db in ["MyMembers", "MyCampaign", "EveryAction"]:
+            self.db_code = 1
+        else:
+            raise KeyError(
+                "Invalid database type specified. Pick one of:"
+                " MyVoters, MyCampaign, MyMembers, EveryAction."
+            )
+
+        session = Scraper()
```
fgregg, line 81: i like the idea of late binding the sessions object, but i'm not sure this the right pattern.

fgregg:
this is ready for an initial look and discussion @shaunagm @KasiaHinkson @sharinef1

shaunagm:
Great, thanks @fgregg. @KasiaHinkson and @sharinef1 - maybe we could all try to take at least a quick look before the contributor meeting Thursday, so we can talk about it then?

austinweisgrau:
I don't love the idea of building the core API connector on top of scrapelib, scrapelib looks like a somewhat minor package without much maintenance, it feels like a somewhat significant liability for the most fundamental and core code in parsons without a lot of upside.

austinweisgrau:
I think it would be a better idea to take some inspiration from the scrapelib design and build those features directly into APIConnector rather than having the dependency

fgregg:
i agree. it's a very small library so we could could just vendor or take it in directly.

shaunagm:
Should I close this PR, or are you interested in exploring bringing scrapelib's features into parsons and potentially using this as a base/reference?

PR #1499, opened 2025-06-30
NewMode retries and pagination

codybraun:
## What is this change?
- We've been seeing a good number of 5xx statuses back from NewMode. They seem to usually just clear on a retry so let's save people the trouble of running a manual retry
- We are also not paginating- so only the first 50 submissions are being synced for example, when there are hundreds of them
- I also added type hints because I was having a hard time following the various transformations we're doing on the Parsons Table

- I don't know that you would want the retry behavior by default on create requests- the only POST I see here though is to submit a job, where I would guess this is okay. Happy to update those assumptions though

- `pytest test`
- or 
```
nm = Newmode(); subs = nm.get_submissions('4c18a593-75a3-4d61-aa87-200d7ed7a35f'); len(sub)
```
(obviously you'll need to add client id and secret to this and pick a campaign id that exists in that account)

### DIFF
--- a/parsons/newmode/newmode.py
+++ b/parsons/newmode/newmode.py
@@ -2,6 +2,7 @@ from typing import Optional, Dict, Any, List, Union
 import logging
 
 from Newmode import Client
+
 from parsons.etl import Table
 from parsons.utilities import check_env
 from parsons.utilities.oauth_api_connector import OAuth2APIConnector
@@ -62,7 +63,7 @@ class NewmodeV1:
 
         return table
 
-    def get_tools(self, params: Dict[str, Any] = {}) -> Table:
+    def get_tools(self, params: Dict[str, Any] = None) -> Table:
         """
         Get existing tools.
         Args:
@@ -71,6 +72,8 @@ class NewmodeV1:
         Returns:
             Tools information as table.
         """
+        if params is None:
+            params = {}
         tools = self.client.getTools(params=params)
         if tools:
             return self.convert_to_table(tools)
--- a/parsons/newmode/newmode.py
+++ b/parsons/newmode/newmode.py
@@ -1,4 +1,4 @@
-from typing import Optional, Dict, Any, List, Union
+from typing import Any, Dict, List, Optional, Union
 import logging
 
 from Newmode import Client
--- a/parsons/newmode/newmode.py
+++ b/parsons/newmode/newmode.py
@@ -1,5 +1,5 @@
-from typing import Any, Dict, List, Optional, Union
 import logging
+from typing import Any, Dict, List, Optional, Union
 
 from Newmode import Client
 
--- a/parsons/newmode/newmode.py
+++ b/parsons/newmode/newmode.py
@@ -1,7 +1,7 @@
+from typing import Optional, Dict, Any, List, Union
 import logging
 
 from Newmode import Client
-
 from parsons.etl import Table
 from parsons.utilities import check_env
 from parsons.utilities.oauth_api_connector import OAuth2APIConnector
@@ -18,9 +18,18 @@ V2_API_CAMPAIGNS_HEADERS = {
     "authorization": "Bearer 1234567890",
 }
 
+PAGINATION_NEXT = "next"
+RESPONSE_DATA_KEY = "data"
+RESPONSE_LINKS_KEY = "links"
+
 
 class NewmodeV1:
-    def __init__(self, api_user=None, api_password=None, api_version=None):
+    def __init__(
+        self,
+        api_user: Optional[str] = None,
+        api_password: Optional[str] = None,
+        api_version: Optional[str] = None,
+    ):
         """
         Args:
             api_user: str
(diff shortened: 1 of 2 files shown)


shaunagm:
@codybraun feel free to @ me once you're done adding things. also hope the linting and formatting checks aren't too burdensome - please let me know if they are!

codybraun:
> @codybraun feel free to @ me once you're done adding things. also hope the linting and formatting checks aren't too burdensome - please let me know if they are!

@shaunagm I'm done! The CI checks are all very reasonable, I've just been too lazy to configure VSC with the different linting setup for Parsons, which is on me! I'll do that if I'm committing code regularly here

parsons/newmode/newmode.py
```diff
@@ -403,54 +439,112 @@ def __init__(
...
+        client.validate_response(response)
+        if response.status_code in success_codes:
+            try:
+                if client.json_check(response):
+                    return response.json()
+            except Exception:
+                logger.error("Response is not in JSON format.")
+        raise ValueError(f"API request encountered an error. Response: {response}")
+
     def base_request(
```
shaunagm, line 456: did a couple of these params on base_request get removed (endpoint, supports_version, etc)? what's the reasoning there?

parsons/newmode/newmode.py
```diff
@@ -403,54 +439,112 @@ def __init__(
...
+    ) -> Optional[Union[Dict[str, Any], None]]:
         """
         Internal method to instantiate OAuth2APIConnector class,
-        make a call to Newmode API, and validate the response.
+        make a single call to Newmode API, and validate the response.
+        """
+        if params is None:
+            params = {}
+
+        for attempt in range(retries + 1):
```
shaunagm, line 473: Why retries +1? and not just retries?

parsons/newmode/newmode.py
```diff
@@ -403,54 +439,112 @@ def __init__(
...
+                return self.checked_response(response, client)
+            except Exception as e:
+                if attempt < retries:
+                    logger.warning(f"Request failed (attempt {attempt + 1}/{retries}). Retrying...")
+                else:
+                    logger.error(f"Request failed after {retries} retries.")
+                    raise e
+        raise Exception(f"Failed to retrieve data from {url} after {retries} attempts.")
+
+    def paginate_request(
```
shaunagm, line 487: Oh is this where those other params ended up? Can you briefly at a high level explain the overall design/refactoring happening here?

shaunagm (commented):
So I overall support this but I'm not sure I understand how base_request, paginated_request and converted_request relate to each other

parsons/newmode/newmode.py
codybraun: that just keeps this sort of semantically correct- so if you said retries=0, you probably still want it to attempt once and then fail. retries=1 should mean two total attempts, etc. - make sense?

parsons/newmode/newmode.py
codybraun: yeah we were previously building the full URL in base_request- that's fine, except now we need some method that can take the already-full-url returned by "next". So we need a base method that doesn't do any decorating of the URL- the same thing should still be happening it is just a level higher in the paginator method

parsons/newmode/newmode.py
codybraun: Yep, basically we need somewhat of a refactor to handle the full URL returned by the "next" pointer to handle iteration. That next pointer already includes the API version, the endpoint, etc. so we want some method that just takes the URL and makes the request, then does our status checks and returns if those look good.

That just means when we are first assembling the URL for an initial request, we just need to do that a level higher (the new pagination wrapper method). Does that make sense? Open to other suggestions for how to refactor this but I don't see where we could handle the pagination without some changes

codybraun: and converted_request I didn't really mess with in part because I don't feel like I have a great understanding of Parsons tables- but basically that wraps paginate_request which wraps base_request

parsons/newmode/newmode.py
shaunagm: yup, makes sense!

parsons/newmode/newmode.py
```diff
@@ -617,19 +717,21 @@ def get_submissions(self, campaign_id, params=None):
         if params is None:
             params = {}
         params = {"action": campaign_id}
-        response = self.converted_request(endpoint="submission", method="GET", params=params)
+        response = self.converted_request(
+            endpoint="submission", method="GET", params=params, data_key=RESPONSE_DATA_KEY
```
shaunagm, line 721: I am curious why some calls to `converted_request` pass in `data_key=RESPONSE_DATA_KEY` and some don't

parsons/newmode/newmode.py
shaunagm: No need to change - this makes perfect sense, I just couldn't quite follow the logic (the type hints while a nice addition made it a little harder to parse what substantively was being changed)

codybraun: yeah sorry, the typing added a bunch to the diff

parsons/newmode/newmode.py
codybraun: yeah I tried to avoid changing this as much as I could, but it looks like that's because in some places we're GETting campaigns/ which returns something like `{"data": [{"campaign_id":1....}]` and in other we're GETing /campaigns/1 which returns ` {"campaign_id":1....}` (note that this is _not_ nested)

shaunagm:
I'm adding the breaking change label just in case so I can flag it in the release notes, even though the signature we're changing is a private method and so it's not a big deal that we're breaking it

shaunagm (approved):

PR #1500, opened 2025-07-01
Cody/feature/fetch zoom webinar occurrences

codybraun:
## What is this change?
- A member wants access to the occurrences of a webinar- turns out a webinar is not a single session, it can be a recurring thing with many instances. We have a member who wants to track the occurrences over time, which makes sense 

- It doesn't appear that meetings also have occurrences, this seems to be just a webinar thing
- It is also a little weird that there is not an occurrence_id on things like webinar attendances... I guess you just have to use the timestamp to know which instance of webinar they attended.

- `pytest test`
- or in a REPL:
```
zoom.get_webinar_occurrences('89216100225')
zoom INFO Retrieved 17 webinar occurrences.
{'occurrence_id': '1753401600000', 'start_time': '2025-07-23T00:00:00Z', 'duration': 60, 'status': 'available'}
{'occurrence_id': '1756425600000', 'start_time': '2025-08-27T00:00:00Z', 'duration': 60, 'status': 'available'}
{'occurrence_id': '1758844800000', 'start_time': '2025-09-24T00:00:00Z', 'duration': 60, 'status': 'available'}
{'occurrence_id': '1761264000000', 'start_time': '2025-10-29T00:00:00Z', 'duration': 60, 'status': 'available'}
{'occurrence_id': '1764291600000', 'start_time': '2025-11-28T01:00:00Z', 'duration': 60, 'status': 'deleted'}
```

### DIFF
--- a/parsons/zoom/zoom.py
+++ b/parsons/zoom/zoom.py
@@ -18,7 +18,12 @@ ZOOM_AUTH_CALLBACK = "https://zoom.us/oauth/token"
 
 
 class ZoomV1:
-    def __init__(self, account_id=None, client_id=None, client_secret=None):
+    def __init__(
+        self,
+        account_id: Optional[str] = None,
+        client_id: Optional[str] = None,
+        client_secret: Optional[str] = None,
+    ):
         """
         Instantiate the Zoom class.
 
@@ -49,7 +54,13 @@ class ZoomV1:
             authorization_kwargs={"account_id": self.account_id},
         )
 
-    def _get_request(self, endpoint, data_key, params=None, **kwargs):
+    def _get_request(
+        self,
+        endpoint: str,
+        data_key: Optional[str],
+        params: Optional[Dict[str, str]] = None,
+        **kwargs,
+    ) -> Table:
         """
         TODO: Consider increasing default page size.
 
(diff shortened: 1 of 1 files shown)


parsons/zoom/zoom.py
```diff
@@ -639,12 +661,27 @@ def get_webinars(self, user_id: int):
         logger.info(f"Retrieved {tbl.num_rows} webinars.")
         return tbl
 
-    def get_user_webinars(self, user_id):
+    def get_webinar_occurrences(self, webinar_id: int) -> Table:
```
codybraun, line 664: This is the only _actual_ change here- I just added type hinting since I've been doing that elsewhere in Parsons but shout if you want me to quit it

IanRFerguson (approved):
nice - it's just an endpoint we can hit? glad we don't have to do any gnarly loops, thanks for knocking this out

parsons/zoom/zoom.py
shaunagm: the type hints are great! on this PR it isn't even an issue of making the PR harder to read because there's one commit with the changes and one commit with the type hints so I can just view them separately :)

PR #1424, opened 2025-04-25
Fix census tests

Charlie-Kramer:
Hi,fixed the census tests, thanks for the feedback!

Charlie

PR #1501, opened 2025-07-02
remove file that shouldn't be in repository

austinweisgrau:
This file looks like it was accidentally included in #1424


shaunagm (approved):
good catch

PR #1502, opened 2025-07-02
Refresh the NewMode token when needed

codybraun:
## What is this change?
- Refresh the NewMode oauth token when it is expired- this is happening a lot more often now that we're correctly paginating through all the results!

- I mean arguably the token refresh failure shouldn't count towards your retries which it currently will but this also keeps things simpler and makes it more certain we're not giong to get stuck in some token-refresh loop since we do count those

- Sort of headache to test since you have to wait for an expiration, added an automated test that I think we can feel good about
`pytest test -k test_token_refresh_on_expired_token`


willyraedy (approved):
LGTM

Issue #1383, opened 2025-03-27
[Bug] Pypi releases should be source-only or documentation should be adjusted about modular imports


fgregg:
#632 introduced modular imports, but it depends on setup.py being executed while a particular environmental variable is set.

However, when you install a wheel, you do *not* run setup.py, so users have no access to the modular imports. 

Until modular imports do not require executing setup.py, one of two things should happen.

1. *only* source-distributions should be uploaded to pypi. don't upload any wheel distributions because pip will prefer the wheels to the source distributions.
2. update [the documentation](https://move-coop.github.io/parsons/html/stable/index.html#integrating-parsons) to instruct the user who wants to use modular dependencies to install from github.

shaunagm:
Do you know if making change #1 - the switch to only source-distributions - could cause any knock-on effects for end users or the automated scripts/workflows people have made with parsons. #1 seems easier on end users generally but I want to make sure we're not causing other problems and I don't know enough about packaging to be confident we aren't.

(Also if you are available for the contributor meeting today, we can definitely talk this through there (plus your other issue about the APIConnector design))

fgregg:
closed by #1505

PR #1505, opened 2025-07-03
simplify distribution

fgregg:
## What is this change?
- The previous github release CI (1342) attempted to build wheels, but we don't need to do wheels since we only want to do a source distribution . This simplifies the release script accordingly (#1383)

- It may desirable to run tests before releasing. Let me know if you want that behavior.

## Testing
Unfortunately, there's no real way to test except for merging and triggering another run (either through manual dispatch or by cutting a new release)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,9 @@ name: Build & Publish Wheels + Source
 
 on:
   release:
-    types: [published]
+    types:
+      - published
+  workflow_dispatch:    
 
 permissions:
     contents: read


requirements-dev.txt
```diff
@@ -12,9 +12,9 @@ ruff==0.11.7
 testfixtures==8.3.0
 
 # Build and publish requirements
-cibuildwheel==2.23.0
```
fgregg, line 15: don't need this anymore.

requirements-dev.txt
```diff
@@ -12,9 +12,9 @@ ruff==0.11.7
...
 twine==6.1.0
-wheel==0.45.1
+build
```
fgregg, line 16: this is what was missing and causing the previous failure.

fgregg: @bmos, it think we should not pin here because build is an official pypa project and so i think we can safely just use the newest version.

shaunagm (approved):

PR #1507, opened 2025-07-07
Fix release (attempt 2)

shaunagm:
Let's just remove the conditional entirely


bmos:
Definitely the right call.
The OS check was only there from when there had been a job matrix.

Issue #1515, opened 2025-07-07
[Behavior Change] Do not set upper bound of Python version


fgregg:
It has been the practice for parsons to specify that it will only work with a minimum python version and a maximum python version, by setting the ["python_requires" argument in setup.py](
https://github.com/move-coop/parsons/blob/362749954f32dbad0c79b24c64e3ece39655fef3/setup.py#L107).

It makes sense to set a minimum version if we only support Python versions more recent than some version, however I think it does not make sense to set a maximum version.

When you set a maximum version, and there has not been a release on PyPi then a user with the most recent of Python cannot install the library. Python is fairly slow moving now, and usually things will work in a newer version.

This was a significant issue when the release process was cumbersome and infrequent. If we have a more frequent release process, and tie that to testing upcoming releases, then there will be no issue.

Thoughts @bmos @shaunagm 

Related discussion: https://github.com/move-coop/parsons/issues/1341#issuecomment-2758894132

shaunagm:
I think this makes sense - we can keep the language in our readme that it has only been tested on XYZ versions but allow folks to install on more recent ones.

bmos:
I have tried running Parsons on Python 3.14, and [it doesn't work](https://github.com/bmos/parsons/actions/runs/16157978799/job/45604154687).

fgregg:
okay! so maybe the behavior change should be keep the maximum version but add the preview version of the next release to our testing matrix, so that would be '3.14-dev' for now?

bmos:
That's a thought. There might be some way to not count that as a failure until it releases.

fgregg:
@bmos yes, we could do that.

https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow

or we could let it fail so that we could find issues early. i think either one would be fine.

maybe it depends on how close we are to the final release.

bmos:
> or we could let it fail so that we could find issues early. i think either one would be fine.

The issue is that most things that cause our incompatibility are outside our control because they are upstream issues from dependencies.

fgregg:
got it! so made a PR to add an upcoming version of python but which should not fail if it, and only it, fails.

fgregg:
but, if it's not under our control then that suggests that we shouldn't be setting a max version. it's the responsibility of the other packages to state what they support?

fgregg:
[...]

PR #1516, opened 2025-07-07
test dev version of python 3.14 but allow for failure

fgregg:
This adds the preview version of the next version of python to the testing matrix, but if it fails but the other versions succeed the whole job will succeed.

relates to #1515


fgregg:
there's not really an elegant way to do this. it still shows an x mark against the commit. 

maybe if https://github.com/actions/runner/issues/2347#issuecomment-2301892819 gets implemented, we will revisit.

PR #1517, opened 2025-07-07
Dont Refresh NewMode Oauth

codybraun:
## What is this change?
- Doesn't seem like NewMode's Oauth2 endpoints support automatic refresh (wherein you exchange your token for a new). Rather we should just use the refresh that exists in this NewMode client class. Currently we error when the token expires after about an error and the error doesn't bubble up to the try/catch block I added for this because we attempt to auto refresh in the Oauth session (which doesn't work)

- Bit hard to test this unfortunately

- Run some NewMode methods until your token expires

### DIFF
--- a/parsons/newmode/newmode.py
+++ b/parsons/newmode/newmode.py
@@ -549,7 +549,12 @@ class NewmodeV2:
             else:
                 results.append(response)
             # Check for pagination
-            url = response.get(RESPONSE_LINKS_KEY, {}).get(PAGINATION_NEXT) if response else None
+            url = None
+            if response:
+                url = response.get(RESPONSE_LINKS_KEY, {}).get(PAGINATION_NEXT, {})
+                if isinstance(url, dict):
+                    url = url.get("href")
+
         return results
 
     def converted_request(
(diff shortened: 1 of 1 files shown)


shaunagm:
@codybraun what is the behavior after this change? We *will* get an error and it *will* bubble up? You write "the error doesn't bubble up to the try/catch block I added for this" but I'm not sure if you mean in your own scripts or a try/catch you added in Parsons

codybraun:
> @codybraun what is the behavior after this change? We _will_ get an error and it _will_ bubble up? You write "the error doesn't bubble up to the try/catch block I added for this" but I'm not sure if you mean in your own scripts or a try/catch you added in Parsons

Yeah sorry, to clarify- the current behavior is: The Oauth client catches the TokenExpired error, and attempts to exchange the token for a fresh one. This fails because NewMode doesn't support that, and raises some InvalidRequest error (not a TokenExpired error) because they don't support that.

The new behavior I believe will be- the Oauth client raises the TokenExpired error because we don't send it a URL to try and refresh it from (because a functional one does not exist). We catch that, and handle it by instantiating a fresh client.

Make sense?

parsons/newmode/newmode.py
```diff
@@ -484,6 +485,7 @@ def base_request(
                 except TokenExpiredError as e:
                     logger.warning(f"Token expired: {e}. Refreshing it...")
                     self.default_client = self.get_default_oauth_client()
+                    client = self.default_client
```
codybraun: otherwise we just keep trying to use the expired one

codybraun:
Alright @shaunagm Kasia found some snags in this today, but we've got it working correctly now and I'll quit pushing changes to it

KasiaHinkson (approved):
Approving because we've now tested this quite thoroughly and I believe Cody answered @shaunagm 's question

PR #1521, opened 2025-07-12
Add Python 3.14 to workflows

bmos:
Per [the discussion](https://github.com/move-coop/parsons/issues/1515#issuecomment-3045724722) in #1515, I have added the latest build of python 3.14 to the testing matrix.
It is set to not fail when the tests fail, so it shouldn't confuse contributors but will allow us to keep an eye on what failures block support.

From running this, we can see some existing compatibility issues that will need to be addressed before supporting 3.14:
- lxml will not install o ubuntu [`lxml` (v6.0.0) was included because `parsons` (v5.2.0) depends on `simple-salesforce` (v1.12.6) which depends on `zeep` (v4.3.1) which depends on `lxml`]
- psycopg2-binary will not install on windows [`psycopg2-binary` (v2.9.10) was included because `parsons` (v5.2.0) depends on `psycopg2-binary`]

There may be other issues after these are resolved. Both packages are the latest versions, which indicates we are most likely dependent on upstream fixes.

### DIFF
--- a/.github/workflows/python-checks.yml
+++ b/.github/workflows/python-checks.yml
@@ -32,7 +32,7 @@ jobs:
                   python-version: ${{ matrix.python-version }}
 
             - name: Install uv
-              uses: install-pinned/uv@05389e8b8c79922b1376aee8c2d598b3219da230  # 0.4.12
+              uses: install-pinned/uv@3aec1379ab70bb5b1be041748d52f765e3a3dc74  # 0.4.12
 
             - name: Install dependencies
               continue-on-error: ${{ contains(matrix.python-version, '3.14') }}
@@ -69,7 +69,7 @@ jobs:
                   python-version: "3.13"
 
             - name: Install uv
-              uses: install-pinned/uv@05389e8b8c79922b1376aee8c2d598b3219da230  # 0.4.12
+              uses: install-pinned/uv@3aec1379ab70bb5b1be041748d52f765e3a3dc74  # 0.4.12
 
             - name: Install dependencies
               run: |
@@ -91,7 +91,7 @@ jobs:
                   python-version: "3.13"
 
             - name: Install uv
-              uses: install-pinned/uv@05389e8b8c79922b1376aee8c2d598b3219da230  # 0.4.12
+              uses: install-pinned/uv@3aec1379ab70bb5b1be041748d52f765e3a3dc74  # 0.4.12
 
             - name: Install dependencies
               run: |
@@ -113,7 +113,7 @@ jobs:
                   python-version: "3.13"
 
             - name: Install uv
-              uses: install-pinned/uv@05389e8b8c79922b1376aee8c2d598b3219da230  # 0.4.12
+              uses: install-pinned/uv@3aec1379ab70bb5b1be041748d52f765e3a3dc74  # 0.4.12
 
             - name: Install bandit
               run: |
@@ -135,7 +135,7 @@ jobs:
 
         steps:
             - name: Harden Runner
-              uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49  # v2.12.2
+              uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911  # v2.13.0
               with:
                   egress-policy: audit
 
@@ -148,7 +148,7 @@ jobs:
                   cache: pip
 
             - name: Install uv
-              uses: install-pinned/uv@05389e8b8c79922b1376aee8c2d598b3219da230  # 0.4.12
+              uses: install-pinned/uv@3aec1379ab70bb5b1be041748d52f765e3a3dc74  # 0.4.12
 
             - name: Install module and dependencies
               run: |
@@ -188,7 +188,7 @@ jobs:
 
             - name: Coverage comment
               id: coverage_comment
-              uses: py-cov-action/python-coverage-comment-action@970a227e0c16ef4589a99a9970ab0ceb8c53059a  # v3.34
+              uses: py-cov-action/python-coverage-comment-action@91910686861e4e1d8282a267a896eb39d46240fb  # v3.35
               with:
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
                   MERGE_COVERAGE_FILES: true
@@ -212,7 +212,7 @@ jobs:
 
         steps:
             - name: Harden Runner
-              uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
+              uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911 # v2.13.0
               with:
                   egress-policy: audit
 
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@ from setuptools import find_packages
 
 def main():
     limited_deps = os.environ.get("PARSONS_LIMITED_DEPENDENCIES", "")
+    install_requires = []
     if limited_deps.strip().upper() in ("1", "YES", "TRUE", "ON"):
         install_requires = [
             "petl",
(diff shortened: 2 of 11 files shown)


fgregg:
nice! if the continue-on-error is on the step [instead of on the job](https://github.com/move-coop/parsons/pull/1516/files) you get a green check. good work figuring that out!

.github/workflows/python-checks.yml
```diff
@@ -27,27 +27,30 @@ jobs:
             - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
             - name: Set up Python ${{ matrix.python-version }}
-              uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
```
fgregg, line 30: what about letting this float, and then we can have the the dest python be "3.14-dev" which will translate to whatever prerelease is most recently supported by actions/setup-python?

.github/workflows/python-checks.yml
```diff
@@ -27,27 +27,30 @@ jobs:
...
+              uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065
               with:
                   python-version: ${{ matrix.python-version }}
 
             - name: Install uv
               uses: install-pinned/uv@51bb057ca8857db7531f188d414ba32b7bb16d2b  # 0.4.12
 
             - name: Install dependencies
+              continue-on-error: ${{ contains(matrix.python-version, '3.14.0-') }}
```
fgregg: if the python version was python 3.14-dev this would be simplified.

setup.py
```diff
@@ -103,8 +103,9 @@ def main():
             "Programming Language :: Python :: 3.11",
             "Programming Language :: Python :: 3.12",
             "Programming Language :: Python :: 3.13",
+            "Programming Language :: Python :: 3.14",
         ],
-        python_requires=">=3.9,<3.14",
+        python_requires=">=3.9,<3.15",
```
fgregg: we probably don't wan to make this change until python 3.14 is green, right?

fgregg (changes requested):
A real improvement over what we have now! 

Not a perfect solution since it is showing the individual 3.14 jobs as green, when they should be red, but much better than [my attempt](https://github.com/move-coop/parsons/pull/1516) which showed the whole job as red, when it should be green. I think this the best we can get until something like ["allow_failure" gets implemented by github](https://github.com/actions/runner/issues/2347#issuecomment-2301892819).

.github/workflows/python-checks.yml
bmos: I am against unpinning actions.
I might be open to compromising to using the tags, like V5, but hash pinning is recommended for GitHub Actions.

setup.py
bmos: Correct. None of the workflow jobs, except for pip-install, actually depend on this change.

.github/workflows/python-checks.yml
fgregg: great, actions/setup-python@v5 would be perfect!

setup.py
fgregg: yes, but we don't want this particular change checked in because we may want to cut a release in the next week or so and the python_requires would incorrectly say that parsons supports 3.14.

fgregg: ah. but there is a catch 22, which until we update this then we can test the new version of python. hmmm…

bmos: Exactly.

bmos: Not actually an issue yet since the workflow can't even get past installing dependencies 🤷🏻‍♀️

.github/workflows/python-checks.yml
```diff
@@ -27,27 +27,30 @@
             - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
             - name: Set up Python ${{ matrix.python-version }}
-              uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
+              uses: actions/setup-python@v5
```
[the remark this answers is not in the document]
bmos: see https://github.com/move-coop/parsons/pull/1521#discussion_r2202829770

.github/workflows/python-checks.yml
```diff
@@ -61,7 +64,7 @@
             - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
             - name: Set up Python 3.13
-              uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
+              uses: actions/setup-python@v5
```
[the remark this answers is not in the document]
bmos: see https://github.com/move-coop/parsons/pull/1521#discussion_r2202829770

.github/workflows/python-checks.yml
```diff
@@ -83,7 +86,7 @@
...
```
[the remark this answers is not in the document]
bmos: see https://github.com/move-coop/parsons/pull/1521#discussion_r2202829770

.github/workflows/python-checks.yml
```diff
@@ -139,7 +142,7 @@
             - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
 
             - name: Set up Python 3.13
-              uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55  # v5.5.0
+              uses: actions/setup-python@v5
```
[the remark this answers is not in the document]
bmos: see https://github.com/move-coop/parsons/pull/1521#discussion_r2202829770

.github/workflows/python-checks.yml
```diff
@@ -105,7 +108,7 @@
             - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
             - name: Set up Python 3.13
-              uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
+              uses: actions/setup-python@v5
```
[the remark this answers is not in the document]
bmos: see https://github.com/move-coop/parsons/pull/1521#discussion_r2202829770

.github/workflows/python-checks.yml
```diff
@@ -216,12 +219,13 @@
             - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
             - name: Set up Python ${{ matrix.python-version }}
-              uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
+              uses: actions/setup-python@v5
```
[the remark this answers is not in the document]
bmos: see https://github.com/move-coop/parsons/pull/1521#discussion_r2202829770

fgregg:
@bmos if we change `uv pip install --system -e .` to `uv pip install --system -e . --ignore-requires-python` will allow us to test 3.14 without loosening the requirements in setup.py before we are ready.

fgregg:
@bmos, reading the code more for setup-actions, i don't think a hashed-commit pin is will also allows us to just use "3.14-dev" and have that get is the most recent prelease version. happy to have that previous behavior reverted.

setup.py
```diff
@@ -103,6 +103,7 @@ def main():
             "Programming Language :: Python :: 3.11",
             "Programming Language :: Python :: 3.12",
             "Programming Language :: Python :: 3.13",
+            "Programming Language :: Python :: 3.14",
```
fgregg: maybe remove this until it's true?

bmos:
> @bmos if we change `uv pip install --system -e .` to `uv pip install --system -e . --ignore-requires-python` will allow us to test 3.14 without loosening the requirements in setup.py before we are ready.

https://github.com/move-coop/parsons/actions/runs/16249114550/job/45876284578#step:5:12

fgregg:
> > @bmos if we change `uv pip install --system -e .` to `uv pip install --system -e . --ignore-requires-python` will allow us to test 3.14 without loosening the requirements in setup.py before we are ready.
> 
> https://github.com/move-coop/parsons/actions/runs/16249114550/job/45876284549#step:5:1

it works locally (but without uv) i’ll see if i can get it work with uv

fgregg:
okay, this indeed a behavior difference between uv and pip. 

however uv does not respect upper bounds for python requires anyway, so i think no action is really needed here. maybe a comment next to the matrix about that fact could be helpful,

https://github.com/astral-sh/uv/issues/8374#issuecomment-2424246009

Issue #1475, opened 2025-06-15
[Feature/Addition] Update documentation to indicate what table columns you get after running various methods


anzelpwj:
## Detailed Description

Something that's been kind of annoying for me is that I have to run Parsons code and then inspect the generated PETL table to know what sort of data I'm expecting to get. If it's pretty clear ahead of time what data we're getting, why don't we make our docstrings more informative about what we get?

For example, for `parsons/ngpvan/email.py:get_email_stats` we just have the docstring:

```
        `Returns:`
            Parsons Table
                See :ref:`parsons-table` for output options.
```

What if we instead did something like:

```
        `Returns:`
            Parsons Table
                See :ref:`parsons-table` for output options. Returned columns are: name, createdBy,
                dateCreated, dateModified, dateScheduled, foreignMessageId, recipientCount,
                bounceCount, contributionCount, contributionTotal, formSubmissionCount,
                linksClickedCount, machineOpenCount, openCount, unsubscribeCount, subject
```

## Context

This is mostly a quality of life improvement. If I can browse the docs as I'm looking for methods, this can help me be more clear if a method is returning the data I expect it to.

## Possible Implementation

Suggested form is above, but if there is a preferred way to do it I'm happy to follow whatever convention we have.

## Priority

Nice to have, not make-or-break. Work on this can probably be done piecemeal.

bmos:
Alternately, what if the Table class was extended to allow defining expected columns?
Then IDEs could actually pick up on it too, instead of it just being in the docstring?
It would still have to be filled in manually, of course.

PR #1522, opened 2025-07-12
Update some docs with column names

anzelpwj:
## What is this change?
- Adding some documentation about the columns you get back for methods in `actblue/actblue.py` and `ngpvan/email.py`.
- An initial effort on #1475 - these are the modules I ingest data with, so I decided to document the outputs here first.

- We discussed a bit about how we wanted to document columns during a prior Parsons meeting and decided that following whatever we had in #1475 was fine for now. Maybe at some point we'd want to make some kind of type hint that specifies the columns in a Parsons table, but for now docs are plenty. There's a lot to start filling out.

- Do the docs build? Do they look okay?

### DIFF
--- a/parsons/actblue/actblue.py
+++ b/parsons/actblue/actblue.py
@@ -169,7 +169,32 @@ class ActBlue(object):
                 Any additional arguments will be passed to Table.from_csv as keyword arguments.
 
         `Returns:`
-            Contents of the generated contribution CSV as a Parsons table.
+            Parsons Table
+                Contents of the generated contribution CSV. List of columns:
+
+                Receipt ID, Date, Amount, Recurring Total Months, Recurrence Number, Recipient,
+                Fundraising Page, Fundraising Partner, Reference Code 2, Reference Code,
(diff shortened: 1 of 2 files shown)


parsons/actblue/actblue.py
```diff
@@ -169,7 +169,32 @@ def get_contributions(self, csv_type, date_range_start, date_range_end, **csvarg
                 Any additional arguments will be passed to Table.from_csv as keyword arguments.
 
         `Returns:`
-            Contents of the generated contribution CSV as a Parsons table.
+            Parsons Table
+                Contents of the generated contribution CSV. List of columns:
+
+                Receipt ID, Date, Amount, Recurring Total Months, Recurrence Number, Recipient,
```
shaunagm, line 175: wow that's a lot of columns

shaunagm (approved):

parsons/actblue/actblue.py
anzelpwj: It is so many...

Issue #1530, opened 2025-07-17
[Bug] pip install parsons breaks because requirements.txt is not included in the distribution


fgregg:
easiest immediate solution would be to include requirements.txt in a MANIFEST.in

https://packaging.python.org/en/latest/guides/using-manifest-in/

cc @bmos @shaunagm 

```python
(.venv) 15:42:00 ~/Developer/temp $ pip install parsons
Collecting parsons
  Downloading parsons-5.1.0.tar.gz (406 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error

  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [24 lines of output]
      Traceback (most recent call last):
        File "/Users/forestgregg/Developer/temp/.venv/lib/python3.13/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 389, in <module>
          main()
          ~~~~^^
        File "/Users/forestgregg/Developer/temp/.venv/lib/python3.13/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 373, in main
          json_out["return_val"] = hook(**hook_input["kwargs"])
                                   ~~~~^^^^^^^^^^^^^^^^^^^^^^^^
        File "/Users/forestgregg/Developer/temp/.venv/lib/python3.13/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 143, in get_requires_for_build_wheel
          return hook(config_settings)
        File "/private/var/folders/g2/qym228p92js6bv6kv1zqwv9h0000gn/T/pip-build-env-n25r6ga9/overlay/lib/python3.13/site-packages/setuptools/build_meta.py", line 331, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
                 ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/private/var/folders/g2/qym228p92js6bv6kv1zqwv9h0000gn/T/pip-build-env-n25r6ga9/overlay/lib/python3.13/site-packages/setuptools/build_meta.py", line 301, in _get_build_requires
          self.run_setup()
          ~~~~~~~~~~~~~~^^
[...]

PR #1531, opened 2025-07-17
Release fixes

bmos:
## What is this change?
- Separate the release action to build, then test on each platform, and then publish if it could install on all
- Add requirements.txt to the release files #1530

- Should we be testing install on all python versions, too?
- Should the build/test jobs be incorporated into the python-check script to run on each PR?
- Should we include other files like docs or tests?

See workflow log:
https://github.com/bmos/parsons/actions/runs/16357543246

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@ from setuptools import find_packages
 
 def main():
     limited_deps = os.environ.get("PARSONS_LIMITED_DEPENDENCIES", "")
+    install_requires = []
     if limited_deps.strip().upper() in ("1", "YES", "TRUE", "ON"):
         install_requires = [
             "petl",


fgregg:
thanks @bmos. 

here are my views:

1. **should we be testing install on all python versions, too?** I don't think we need to do this. our normal python testing script is ensuring that the setup.py works across a many different versions. these additional tests are really just to ensure that we have a full manifest.
2. **Should the build/test jobs be incorporated into the python-check script to run on each PR?** neutral on this
3. **Should we include other files like docs or tests?** No. the only way for people to access those would be go go into there `site-packages` folder and poke around, and that's not a normal thing to do. Adding docs and tests are just unhelpful weight for installers.

.github/workflows/release.yml
```diff
@@ -46,9 +37,57 @@ jobs:
...
+          path: dist/
+
+  test:
+    needs: build
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+
+    runs-on: ${{ matrix.os }}
```
fgregg, line 53: i don't thing we really need a matrix here, just ubuntu should be fine. but it's up to you.

fgregg (approved):
looks good!

.github/workflows/release.yml
bmos: I'm not personally familiar enough with Python packaging to know whether there are realistic scenarios in which a package would install on one operating system and not another, so I included it out of an abundance of caution.

shaunagm: We don't cut releases a ton so I think running the extra matrix jobs should not be a big deal.

shaunagm:
This looks good to me - what (if anything) needs to be decided before we merge this?

fgregg:
> This looks good to me - what (if anything) needs to be decided before we merge this?

i think it’s good to go. but defer to @bmos

bmos:
I don't think there's anything else we need to do here.

shaunagm (approved):
lgtm (per slack convo)

PR #1555, opened 2025-08-14
Remove data type support value error in copy from gcs function

sharinetmc:
## What is this change?
- Remove data type support value error in copy from gcs function since it now supports other data_types (not just csv and json)

- (How should a reviewer test this functionality)

### DIFF
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -1565,8 +1565,6 @@ class GoogleBigQuery(DatabaseConnector):
                 f"Unexpected value for if_exists: {if_exists}, must be one of "
                 '"append", "drop", "truncate", or "fail"'
             )
-        if data_type not in ["csv", "json"]:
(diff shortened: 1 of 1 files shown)


parsons/google/google_bigquery.py
```diff
@@ -1559,13 +1561,15 @@ def _fetch_query_results(self, cursor) -> Table:
...
+    def _validate_copy_inputs(
+        self, if_exists: str, data_type: str, override_data_type_check: bool = False
+    ):
         if if_exists not in ["fail", "truncate", "append", "drop"]:
             raise ValueError(
                 f"Unexpected value for if_exists: {if_exists}, must be one of "
                 '"append", "drop", "truncate", or "fail"'
             )
-        if data_type not in ["csv", "json"]:
+        if data_type not in ["csv", "json"] and not override_data_type_check:
```
KasiaHinkson: Is there a data type we can just add to the list that we now accept?

KasiaHinkson (approved):
I love this!

PR #1646, opened 2025-11-27
Add survey methods to ActionNetwork class

matthewkrausse:
- Introduces methods for managing surveys in the ActionNetwork class, including retrieving, creating, and updating surveys.

### DIFF
--- a/parsons/action_network/action_network.py
+++ b/parsons/action_network/action_network.py
@@ -1805,7 +1805,8 @@ class ActionNetwork:
                 When None, no filter is applied.
 
         `Returns:`
-            A JSON with all of the survey entries
+            Parsons Table
+                A Table with all of the survey entries
         `Documentation Reference`:
             https://actionnetwork.org/docs/v2/surveys
         """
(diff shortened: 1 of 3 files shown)


parsons/action_network/action_network.py
```diff
@@ -1788,6 +1788,198 @@ def update_submission(self, form_id, submission_id, data):
...
+        if browser_url is not None:
+            payload["browser_url"] = browser_url
+        if featured_image_url is not None:
+            payload["featured_image_url"] = featured_image_url
+
+        # Add any additional fields from kwargs
+        payload.update(kwargs)
+
+        if background_processing:
+            payload["action_network:background_processing"] = background_processing
```
NirTatcher, line 1933: Hi @matthewkrausse, are we sure that Surveys support background processing? Also we usually add the param as a query param to the URL:
```
        if background_processing:
            url = f"{url}?background_processing=true"

        response = self.api.post_request(url, data=json.dumps(data))
```
You sure what you wrote is accurate? Did you test that? Or maybe we can just remove the background processing param and logic all together from Surveys?
image

bmos: @matthewkrausse just wanted to bump this question for you

matthewkrausse:
Yeah, I think we are good to remove. Must've gotten confused on it it was
supported.

bmos:
in that case this seems like it can be closed :)
i don't think there is anything left in this that isn't on main after #1647

PR #1818, opened 2026-04-08
Catalist: Patch Exception Handling

IanRFerguson:
## What is this change?
Adds an exception for clarity in the `CatalistMatch` class if there is a failure to save the zipped file from the vendor's SFTP server. The current active exception is raising on `zf.extractall(path=temp_dir)`, which tells us there's a problem with the file but not much else.

## Breaking changes (if needed)
This is not a breaking change.

N/A

## Labels
Please add exactly one of the following labels to this PR:
- `breaking-change` — This PR introduces a breaking change to the public API
- `non-breaking-change` — This PR does not break any existing functionality

The label check CI job will fail until one of these labels is applied.


parsons/catalist/catalist.py
```diff
@@ -371,11 +371,22 @@ def load_matches(self, id: str) -> Table:
...
         remote_filename = [filename for filename in remote_filepaths if id in filename][0]
         remote_filepath = "/myDownloads/" + remote_filename
+
         temp_file_zip = self.sftp.get_file(
             remote_path=remote_filepath, export_chunk_size=DEFAULT_EXPORT_CHUNK_SIZE
         )
-        temp_dir = tempfile.mkdtemp()
+        logger.debug(
+            f"Download complete for job {id} (local size: {temp_file_zip.stat().st_size} bytes)."
+        )
```
bmos: as the information in the debug log message is already in the error message, perhaps it should be after the exception?

bmos (approved):

parsons/catalist/catalist.py
IanRFerguson: oh yeah good catch!

bmos:
In case it helps, here are rebuilt tests that actually use real files.
https://github.com/move-coop/parsons/compare/ian/patch/catalist-zip-operation...bmos:parsons:catalist-testing

IanRFerguson:
> In case it helps, here are rebuilt tests that actually use real files. [ian/patch/catalist-zip-operation...bmos:parsons:catalist-testing](https://github.com/move-coop/parsons/compare/ian/patch/catalist-zip-operation...bmos:parsons:catalist-testing)

awesome, appreciate you Ramona! looks like everything besides the label check was passing (and I've since fixed that)

should I merge your branch into this one? or what's the standard operating procedure?

bmos:
Merging the tests before merging this PR is probably better in theory because it shows that the tests pass with the existing code, but I'm fine with either and will be doing PR reviews with @shaunagm this afternoon if it's preferred to handle them separately.

shaunagm (approved):
approved and happy to merge once the tests pass (I assume they're not failing from this and the PR just needs to be updated)

bmos:
They won't pass, the test suite needs to be changed.
~~There's a PR for @IanRFerguson to merge into this.~~
I'm going to post the PR against main since the test improvements are technically separate.

shaunagm:
oh gotcha. i'll wait then!

bmos:
@IanRFerguson The tests have been merged into main. If you update this branch from main I think they'll all pass.

sharinetmc (approved):

IanRFerguson:
hi @bmos - just want to make sure I'm tracking your changes correctly, should these checks be passing?

bmos:
Hm, idk why it's not passing now.
I'll double check later this week.

parsons/catalist/catalist.py
```diff
@@ -371,11 +371,22 @@ def load_matches(self, id: str) -> Table:
         remote_filepaths = self.sftp.list_directory("/myDownloads/")
         remote_filename = [filename for filename in remote_filepaths if id in filename][0]
         remote_filepath = "/myDownloads/" + remote_filename
+
         temp_file_zip = self.sftp.get_file(
             remote_path=remote_filepath, export_chunk_size=DEFAULT_EXPORT_CHUNK_SIZE
         )
```
bmos: This will fix your test failure, not sure what happened here.

```suggestion
        temp_file_zip = Path(
            self.sftp.get_file(
                remote_path=remote_filepath,
                export_chunk_size=DEFAULT_EXPORT_CHUNK_SIZE,
            )
        )
```

parsons/catalist/catalist.py
```diff
@@ -371,11 +371,22 @@ def load_matches(self, id: str) -> Table:
...
+        if not is_zipfile(temp_file_zip):
+            raise RuntimeError(
+                f"Downloaded file for job {id} is not a valid zip file "
+                f"(size: {temp_file_zip.stat().st_size} bytes, remote path: {remote_filepath}). "
+                "The SFTP download may be corrupt or incomplete."
+            )
+
+        logger.debug(
+            f"Download complete for job {id} (local size: {temp_file_zip.stat().st_size} bytes)."
+        )
```
bmos, line 391: On an unrelated note, you might also want to use lazy formatting in your logger call, since f-strings are resolved even when the logger is not in debug mode.
```suggestion
        logger.debug(
            "Download complete for job %s (local size: %s bytes).", id, temp_file_zip.stat().st_size
        )
```

PR #1820, opened 2026-04-08
Comment out requirements instructions in requirements.txt

bmos:
dependabot doesn't like that we have a requirements.txt file with invalid syntax.

## What is this change?

- Comment out contents of requirements.txt file.

- Should we just delete the file? We left it hoping that the error people would get when trying to install it would alert people to the new install method.

## Breaking Changes

<details open>

</details>


shaunagm (approved):
Let's leave this for a while and then remove it in six months or so.

PR #1826, opened 2026-04-10
Refactor and Document check_env.check()

bmos:
## What is this change?

- Adjust a few nonstandard imports of check directly rather than via importing check_env.
- Add comprehensive documentation to `check_env.check()`.
- Convert `optional` parameter to keyword-only.
- Rename `field` parameter to `value`.
- Improve tests for check_env.check
- Add deprecation handler to avoid `optional` and `field` breaking changes.

- Despite being a core function used in many connectors, `check_env.check()` was minimally documented.
- Allowing users to provide boolean values for positional parameters is not recommended as their intention can be confusing.

- CI tests should be adequate

## Breaking Changes

<details open>

</details>


bmos:
Do we consider these breaking changes?
I think this function is only used within connectors?

parsons/utilities/check_env.py
```diff
@@ -1,18 +1,106 @@
 import os
+import warnings
+from typing import Literal, TypeVar, overload
 
+T = TypeVar("T")
 
-def check(env: str, field: str | None, optional: bool | None = False) -> str | None:
+
+@overload
```
shaunagm, line 8: What are these @overload decorators doing?

parsons/utilities/check_env.py
```diff
@@ -1,18 +1,106 @@
...
-    and the passed field or arguments have not been passed, then raise an
-    error.
+    Check if an environment variable has been set or value has been provided.
+
+    Args:
+        env:
+            Name of environment variable to check.
+        value:
+            If provided, ignore environment variable and return this.
+        opt:
```
shaunagm, line 57: Where is the opt parameter coming from? I get that it's being listed as deprecated, but I'm not seeing where it's used at all.

parsons/utilities/check_env.py
bmos: Overload is a type hinting feature. 
It tells static type checkers that if you provide a certain type as an input you get a certain type as an output.

parsons/utilities/check_env.py
bmos: optional (the former 3rd positional argument) is now a keyword-only argument. So I added "opt" as the 3rd positional argument with the deprecation warning. That way it's not a breaking change if people try to provide the optional parameter as a positional argument.

parsons/utilities/check_env.py
shaunagm: Gotcha. Why are there three of them? It feels more verbose than is ideal (but maybe if we can use a comment to signal to the reader that these are type hints with no impact on the code itself, the verbosity is less of an issue)

parsons/utilities/check_env.py
shaunagm: Gotcha. What's your reasoning on moving it from a 3rd positional argument to a keyword-only argument?

parsons/utilities/check_env.py
bmos: That's just how the syntax works, unfortunately.
That's why overload is only used situationally. I think it's worth it here since check is used in a ton of connectors so having accurate typing is extra useful.

parsons/utilities/check_env.py
bmos: https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/

parsons/utilities/check_env.py
bmos: Happy to chat through how this works sometime, I just keep being busy on Thursdays 😅
It's a very helpful feature.

shaunagm: You've got some other PRs I could review too so yeah let's schedule a call.

shaunagm (approved):

PR #1827, opened 2026-04-11
Fix BigQuery API (and others) not showing in Parsons Documentation

bmos:
#1721 broke our documentation build, which needs access to all our dependencies.

## What is this change?

- Install parsons with all extras during build job of documentation workflow
- Improve type hints
- Improve docstrings
- Remove unused variables
- Pass `**load_kwargs` to load_table_from_file() as described in docstring
- Mark static methods with @staticmethod

- CI tests should be adequate

## Breaking Changes

<details open>

</details>

### Details (if needed)

-

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -51,6 +51,7 @@ jobs:
           --upgrade
           --no-default-groups
           --group docs
+          --all-extras
 
       - name: configure git
         run: |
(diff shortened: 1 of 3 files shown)


shaunagm:
@bmos can you confirm this should be a non-breaking change?

bmos:
I changed the names of functions that are not intended for public use. It would be breaking if someone was using those utility functions. If desired, I can add deprecation aliases to make it 100% not breaking even in that edge case.

shaunagm:
Yes please - or separate them out into two PRs so we can fix the broken docs quickly and wait until the next major release for the (admittedly very minor) breaking change.

bmos:
That's probably the better approach, they're not really in scope and I didn't think of them as breaking when I made them

shaunagm (approved):

shaunagm:
Thanks for making the changes @bmos!

PR #1832, opened 2026-04-14
Move dependencies to pyproject.toml and eliminate setup.py

bmos:
Thanks to self-referencing dependencies, we can create parsons[all] within pyproject.toml. See [here](https://discuss.python.org/t/aliased-or-aggregated-optional-dependencies/20488), [here](https://hatch.pypa.io/dev/config/dependency/#self-referential), and [here](https://github.com/pypa/pip/issues/11296#issue-1315774941) (poorly documented) for more about self-referencing dependencies.

## What is this change?

- Moved dependencies to pyproject.toml and eliminated setup.py
- Blacklist mysql-connector-python 9.7.0 as it was released incorrectly and doesn't install (needed to get CI working).

- Ability to self-reference extras was [added in pip 21.2](https://github.com/pypa/setuptools/issues/1260), so older pip versions are not supported. As python 3.10 is our oldest supported python version and was released in Oct 2021, our users shouldn't be on versions older than that.

- CI tests should be adequate.

## Breaking Changes

<details open>

</details>

### Details (if needed)


bmos:
@shaunagm done!

shaunagm (approved):

PR #1853, opened 2026-04-24
Add catalist match download retry logic

sharinetmc:
## What is this change?

- Add catalist match download retry logic

## Breaking Changes

<details open>

</details>

### Details (if needed)


parsons/catalist/catalist.py
```diff
@@ -339,7 +339,45 @@ def await_completion(
...
+            try:
+                # 1. Attempt the download
+                temp_file_zip = self.sftp.get_file(
+                    remote_path=remote_path, export_chunk_size=DEFAULT_EXPORT_CHUNK_SIZE
+                )
+
+                # 2. Validate the integrity of the file
+                if not is_zipfile(temp_file_zip):
+                    size = temp_file_zip.stat().st_size
+                    raise RuntimeError(f"Corrupt zip (Size: {size} bytes)")
```
IanRFerguson, line 357: curious why we don't just decorate this function with `@backfoff`, the retry logic feels kind of manual

IanRFerguson (commented):
I have a code question and a context question

in terms of the code, it feels like we could wrap the function with `@backoff.expo` and achieve a similar outcome

in terms of context, this seems fine I'm just wondering if we have reason to believe that a retry is going to solve this Catalist issue. probably can't hurt I just want to understand the intent here

parsons/catalist/catalist.py
```diff
@@ -339,7 +339,45 @@ def await_completion(
...
+        self, remote_path: str, job_id: str, max_retries: int = 3, backoff_factor: int = 2
+    ):
+        last_exception = None
+
+        for attempt in range(1, max_retries + 1):
+            try:
+                # 1. Attempt the download
+                temp_file_zip = self.sftp.get_file(
+                    remote_path=remote_path, export_chunk_size=DEFAULT_EXPORT_CHUNK_SIZE
+                )
```
bmos, line 352: This will solve your test failure
```suggestion
                temp_file_zip = Path(
                    self.sftp.get_file(
                        remote_path=remote_path,
                        export_chunk_size=DEFAULT_EXPORT_CHUNK_SIZE
                    )
                )
```

parsons/catalist/catalist.py
austinweisgrau: agree

PR #1854, opened 2026-04-28
Blacklist broken mysql-connector-python v9.7.0

bmos:
They didn't publish this version correctly, so it's only compatible with some versions of python. It also seems to only contain documentation updates, so it's very weird they pushed it as a minor version change.

## What is this change?

- Fix CI + installing from main by blacklisting broken dependency

- Here is [what changed](https://github.com/mysql/mysql-connector-python/compare/9.6.0...9.7.0) between v9.6 and v9.7.
- Here are [the files they published to pypi](https://pypi.org/project/mysql-connector-python/9.7.0/#files) for 9.7.0. You'll see it only supports 3.10 and 3.11.
- Here are [the files they published to pypi](https://pypi.org/project/mysql-connector-python/9.6.0/#files) for 9.6.0. You'll see there are many more files.

- CI tests are adequate

## Breaking Changes

<details open>

</details>

### Details (if needed)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ EXTRA_DEPENDENCIES = {
         "validate-email >= 1",
     ],
     "mysql": [
-        "mysql-connector-python >= 7",
+        "mysql-connector-python >= 7, != 9.7.0",
         "sqlalchemy >= 1.4",
     ],
     "newmode": ["newmode >= 0.1.6"],


bmos:
@shaunagm FTFY :) Unfortunately this is the downside of having such permissive dependency version requirements.

I think #1833 will help avoid this in the future, but the idea of including a lockfile and how we should manage that would benefit from discussion.

shaunagm (approved):

PR #1857, opened 2026-05-03
APIConnector Typing / Documentation

bmos:
## What is this change?

- Improve documentation for [APIConnector](https://move-coop.github.io/parsons/stable/utilities.html#module-parsons.utilities.api_connector) and [OAuth2APIConnector](https://move-coop.github.io/parsons/stable/utilities.html#module-parsons.utilities.oauth_api_connector)
- Add type hints for APIConnector and OAuth2APIConnector
- Call get_request with keyword args within parsons codebase
- Authenticate APIConnector with HTTPBasicAuth instead of tuple within parsons codebase
- Implement `raise_on_error` param for `APIConnector.get_request` which was documented but not implemented
- Add `raise_on_error` param to the other APIConnector *request methods
- Pass kwargs to requests.request from all APIConnector *request methods for advanced use-cases

- Because of quirks of python type hint syntax, I had to make most parameters of `APIConnector.get_request` keyword-only. For consistency, I did this for all the `*request` methods.

- CI tests should be adequate.

## Breaking Changes

<details open>

- [x] label: Breaking change — This PR introduces one or more breaking changes.

</details>

### Details (if needed)

- Switching to keyword-only args could be a breaking change, if users are using [APIConnector](https://move-coop.github.io/parsons/stable/utilities.html) directly. Do we consider APIConnector a public API? I wasn't sure because, while it is in our documentation, it's described as just being used by other connectors.


shaunagm (approved):
This is only a breaking change if people are using Parsons internals, which they shouldn't be. But I'll make sure it's highlighted in the release notes just in case.

PR #1877, opened 2026-05-31
Resolve CI concurrency skipping jobs when it shouldn't

bmos:
## What is this change?

- Adjust concurrency groups so that job-specific configs include job names
- Apply concurrency to entire documentation workflow instead of each job
- Fix naming of a workflow step in release.yml

- Concurrency conflicts are arising because the concurrency groups are not taking into account job names when configured within a job. This means that when the linkcheck and docs-test jobs are initialized, one of them is skipped every time.
- This didn't show up in testing because linkcheck doesn't run on PRs.

- It worked on my fork. You can ensure that CI workflows pass on this PR, but unless you push multiple commits quickly you won't see a difference.

## Breaking Changes

<details open>

</details>

### Details (if needed)

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -24,15 +24,15 @@ env:
     *.github.com:443
     *.githubusercontent.com:443
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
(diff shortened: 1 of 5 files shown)


bmos:
@shaunagm this should fix the test failures you've been getting on main

shaunagm (approved):

PR #1882, opened 2026-06-06
Consistent Logging

bmos:
## What is this change?

- Ensure all logging calls are sent through a named logger (not directly to logging)
- Refactor all logger calls to use lazy string concatenation

- logging.exception includes the traceback, logging.error generally does not
- using f-strings in logger calls [is not performant](https://docs.astral.sh/ruff/rules/logging-f-string/)

- CI tests, although some code has no coverage. Because they're logging calls, I am not concerned.

## Breaking Changes

<details open>

</details>

### Details (if needed)


shaunagm:
@bmos could this potentially break people's logging pipelines? It's not a breaking change to the API but I'm curious about how folks are using our logging (if at all) and if we should be concerned about breaking that. @sharinetmc @IanRFerguson any thoughts?

bmos:
I think switching from error to exception potentially could but it's a real edge case.
It's also not that important to include so I could revert that.

bmos:
I think that will have resolved the concern.

bmos:
> @bmos could this potentially break people's logging pipelines?

I believe I have updated this PR so that it does not carry this risk.

docs/training_guides/etl_best_practices.rst
```diff
@@ -532,9 +532,7 @@ We start by pulling our Mobilize data out of the Redshift table where it's been
    sql_query = 'select * from mobilize_schema.mobilize_users_to_sync limit 5;'
    new_mobilize_users = my_rs_warehouse.query(sql_query)
 
-   logger.info(
-      f"There are {new_mobilize_users.num_rows} new mobilize users that need to be synced to Action Network."
-   )
+   logger.info("There are %s new mobilize users that need to be synced to Action Network.",new_mobilize_users.num_rows)
```
shaunagm: Is there a reason all the f-strings are swapped out?

PR #1884, opened 2026-06-16
ngpvan: don't re-append params when following pagination links

fgregg:
VANConnector.get_request re-passed the caller's **kwargs (including `params`) to each `nextPageLink`. But `nextPageLink` is already has all the params, so we were accumulating params, breaking pagination.

## What is this change?

- Fixes pagination in the VANConnector

I added a unit test for this

## Breaking Changes

No breaking changes. Just a bugfix.


shaunagm (approved):

shaunagm:
Thanks so much @fgregg!

Issue #364, opened 2020-09-04
Convenience Function to get Distkey & Sortkey


elyse-weiss:
Given the new warnings in the Redshift class, I would find it super useful to have a convenience function that gets you distkey and sortkey.

shaunagm:
Is this something we still want? If so, can you provide more information, @elyse-weiss?

elyse-weiss:
Yes! Distkey and Sortkeys are very important and Redshift, and I do think this would be a used convenience function - especially if you need to drop and re-create a table.

shaunagm:
Okay, bumping up the priority. Does TMC have a redshift sandbox that a contributor could access while working on this?

elyse-weiss:
Upon request, I could definitely create a Redshift user and give them access to our `communal_data_assets` schema, which just includes publicly available data, so they can play around there!

HandcartCactus:
This seems to be stalled because it requires Redshift (a paid service). I've got an AWS account. I could adopt this ticket, if it's worthwhile.

shaunagm:
@HandcartCactus that would be amazing!

HandcartCactus:
Sounds good. This is my first issue so it might take a while to get spun up.

This issue doesn't state what the actual warning is or when it comes up in the Redshift class, so I'll try to see if I can get it to appear first.

From my research online, a table's distkey and sortkey can be found in svv_table_info and/or pg_table_def if the table is in the user's search_path.

HandcartCactus:
Huh. I'm trying to use the Redshift class for the first time and it appears to be "invisible" to my language server/mypy. I've tried this both installed directly from github and the distribution from pypi. I wish the DX was better. 

<img width="366" height="78" alt="Image" src="https://github.com/user-attachments/assets/03963383-ec13-4b70-aea6-61aaadd59e7f" />

This appears to be caused by imports coming from the `__init__.py` `_CONNECTORS` list as strings, instead of as regular import statements. Here's the imports:
<img width="665" height="700" alt="Image" src="https://github.com/user-attachments/assets/ab6b3d21-4f5e-441c-ab9c-6d66a33c34fc" />
Here's how they get imported, which is quite interesting.
<img width="727" height="485" alt="Image" src="https://github.com/user-attachments/assets/7261ac9e-abe1-4871-bc4b-080a1b27377b" />

[...]

PR #1887, opened 2026-06-22
Add Redshift distkey & sortkey helper methods

HandcartCactus:
## What is this change?

Adds methods requested in #364 plus some other helper functions for search path management.

+ `get_search_path`
+ `set_search_path`
+ `add_schema_to_search_path`
+ `get_distkey`
+ `get_sortkey`

+ If the table could not be found in the internals, sortkey and distkey methods return `None` instead of a blank table. Setting `errors='raise'` will force it to raise a ValueError instead of return None. I'm not 100% sure "None instead of Table" was the right call but it seemed like it might confuse users to get a blank table. None (or some kind of error) is a little more expected.
+ I've gotten some mixed signals between the docs and the actual testing if having the schema in the `search_path` is actually required to the the distkey and sortkey. Kept those methods in just in case, because I used them while troubleshooting and they have other uses as well.

1. set up redshift test environment
2. 
```bash
pytest test/test_redshift/test_redshift.py::TestRedshiftDB --live
```

## Breaking Changes

<details open>

</details>

### Details (if needed)

### DIFF
--- a/parsons/databases/redshift/redshift.py
+++ b/parsons/databases/redshift/redshift.py
@@ -1236,6 +1236,11 @@ class Redshift(
         return RedshiftTable(self, table_name)
 
     def get_search_path(self):
+        """Returns the schema search_path for the current user.
+
+        Returns:
+            Table|None: The user's "search_path".
+        """
         sql_searchpath = """SHOW search_path;"""
 
         with self.connection() as connection:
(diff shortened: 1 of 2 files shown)


shaunagm (approved):
This looks great - thanks for submitting! I am not really able to review the SQL code you're using to query for the sortkeys and distkeys but the tests pass and I assume you've inspected the results when testing locally.

HandcartCactus:
Yes, I inspected the results when testing locally and added live tests to ensure they are working properly. Thanks for looking this over!

PR #1888, opened 2026-06-23
Add connection timeout to MySQL connector (#452)

### LINKED ISSUE #452: Add Timeout Arg to MySQL Connector
I think it's as simple as modifying init and connection along the following lines:
```
def __init__(self, host=None, username=None, password=None, db=None, port=3306, timeout=1200):

    self.username = check_env.check('MYSQL_USERNAME', username)
    self.password = check_env.check('MYSQL_PASSWORD', password)
    self.host = check_env.check('MYSQL_HOST', host)
    self.db = check_env.check('MYSQL_DB', db)
    self.port = port or os.environ.get('MYSQL_PORT')
    self.timeout = os.environ.get('MYSQL_TIMEOUT') or timeout

@contextmanager
def connection(self):

    # Create a mysql connection and cursor
    connection = mysql.connect(host=self.host,
                               user=self.username,
                               passwd=self.password,
                               database=self.db,
                               port=self.port,
                               connect_timeout=self.timeout
                              )

    try:
        yield connection
    except mysql.Error:
        connection.rollback()
        raise
    else:
        connection.commit()
    finally:
        connection.close()
```

Musikavanhu:
Closes #452.

The MySQL connector had no way to set a connection timeout, unlike the Postgres and Redshift connectors, which both accept `timeout=10` and forward it to the driver. This adds a `timeout` argument to `MySQL` and forwards it to `mysql.connector.connect` as `connection_timeout`, matching that convention.

- Matches the existing 10s default of the Postgres and Redshift connectors (previously, no timeout was explicitly set, defaulting to OS TCP defaults). If users on very slow links require the previous behavior, they can pass an explicit/larger `timeout` value.
- Added `TestMySQLConnectionTimeout` covering the default, an explicit value, and that the resolved timeout is forwarded to the connect call (mocked, no live DB).
- `ruff check` / `ruff format` clean.

- [ ] label: Breaking change
- [x] label: Non-breaking change

### DIFF
--- a/parsons/databases/mysql/mysql.py
+++ b/parsons/databases/mysql/mysql.py
@@ -38,10 +38,12 @@ class MySQL(DatabaseConnector, MySQLCreateTable, Alchemy):
         port: int
             If omitted or ``None``, uses ``MYSQL_PORT`` when set, otherwise 3306. If passed
             (including ``3306``), the argument takes precedence over ``MYSQL_PORT``.
+        timeout: int
+            Seconds to timeout if connection not established. Defaults to 10.
 
     """
 
-    def __init__(self, host=None, username=None, password=None, db=None, port=None):
+    def __init__(self, host=None, username=None, password=None, db=None, port=None, timeout=10):
         super().__init__()
 
         self.username = check_env.check("MYSQL_USERNAME", username)
@@ -53,6 +55,8 @@ class MySQL(DatabaseConnector, MySQLCreateTable, Alchemy):
         else:
             env_port = check_env.check("MYSQL_PORT", None, optional=True)
             self.port = int(env_port) if env_port is not None else 3306
+
+        self.timeout = timeout
 
     @contextmanager
     def connection(self):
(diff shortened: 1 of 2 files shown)


shaunagm (approved):
Thank you, this looks great! I'll flag it in the release notes in case anyone was somehow relying on a longer timeout, but overall this seems like a sensible approach.

(BTW, are you in the Parsons Slack? If not, do you want to be?)

Musikavanhu:
> Thank you, this looks great! I'll flag it in the release notes in case anyone was somehow relying on a longer timeout, but overall this seems like a sensible approach.
> 
> (BTW, are you in the Parsons Slack? If not, do you want to be?)

Thanks! And yes, I'd love to be! appreciate the invite. My email is musikavanhu1@gmail.com if that's easiest, otherwise happy to use an invite link.

shaunagm:
You should have an invite! Looking forward to having you join us :)

PR #1891, opened 2026-07-01
Remove unused variable in census_geocoder

bmos:
## What is this change?

- Remove an unused variable in census_geocoder.py

- This variable doesn't do anything and has been in the codebase since its earliest commit.
- I recommend merging #1892 first so that this will not have 0% test coverage.

- CI tests do not cover this, but you can see it's not used.

## Breaking Changes

<details open>

</details>

### Details (if needed)


shaunagm (approved):
goodbye `batch_count` we hardly knew you

PR #1892, opened 2026-07-01
Run all Census Geocoder tests without network access (no more live test requirement!)

bmos:
## What is this change?

- Refactor censusgeocode tests to use pytest fixture instead of unittest framework
- Always run any censusgeocode tests that don't require network access
- Add [pytest-recording](https://pypi.org/project/pytest-recording/) to dev dependencies
- Use pytest-recording to always run the live test without requiring network access (re-using response data I captured locally)
- Update expected response from API as census data has changed since the tests were added and ran (another reason to use recorded responses)

- Census Geocode tests didn't run because it was all marked as a live test.

- Run `pytest .\test\test_geocode`
- View CI logs

## Breaking Changes

<details open>

</details>

### Details (if needed)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -216,6 +216,7 @@ test = [
     "pytest~=9.0.3",
     "pytest-datadir~=1.8.0",
     "pytest-mock~=3.15.1",
+    "pytest-recording~=0.13.4",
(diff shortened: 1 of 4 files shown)


shaunagm (approved):
Great catch - thank you for fixing this!

PR #1899, opened 2026-07-14
Bigquery get_table_ref doesn't ignore project_id

austinweisgrau:
prior behavior would ignore a project_id in a table ref, defaulting to the project_id of the whole BigQuery connector. This update ensures that the project_id included in a table ref will not be ignored

## What is this change?

In parsons as-is, this operation: 
```
BigQuery(project_id='project_A').copy(tbl, 'project_B.some_dataset.some_table')
```
would load into `project_A.some_dataset.some_table` - or more likely, simply fail because `project_A.some_dataset` doesn't exist

This bug occurs because `get_table_ref` ignores the project_id. This PR fixes that so that the project_id is used if specified.

## Breaking Changes

<details open>

</details>

### Details (if needed)

### DIFF
--- a/parsons/google/google_bigquery.py
+++ b/parsons/google/google_bigquery.py
@@ -11,10 +11,10 @@ from typing import Literal
 import google
 import petl
 from google.api_core import exceptions
+from google.auth.credentials import Credentials
 from google.cloud import bigquery
 from google.cloud.bigquery import ExtractJob, dbapi, job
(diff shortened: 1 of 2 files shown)


shaunagm (approved):
i'll wait for willy to approve but this looks good to me

shaunagm:
This PR has been sitting for a while, @willyraedy did you want to review? If I don't hear from you in a few days I'll just merge since it's a small, sensible and hopefully non-breaking change.

PR #1907, opened 2026-07-30
Test suite: standard + migrations + coverage/mutation parity tooling

matthewkrausse:
# Test suite: one standard, safe incremental migration, and coverage/mutation parity tooling

This PR establishes a single, documented testing standard for Parsons connectors and migrates the existing suites onto it — **without weakening a single test**. It also adds tooling that objectively proves that claim (per-connector coverage *and* mutation-score baselines), wired into CI.

It is large because it touches many connectors, but it is built to be reviewed and landed **one connector at a time**: every connector is an independent change, and every connector keeps its old tests running side-by-side until the new ones are proven equivalent.

---

## Problem

Parsons' tests had drifted into several incompatible styles, and a lot of them weren't testing what they appeared to test:

- **Three different mocking approaches** with no rule for which to use — HTTP (`requests_mock`), SDK clients (`unittest.mock`), and hand-rolled fakes — often mixed within one file.
- **Tests that mock the connector's own methods.** For example `test_action_kit` replaced `self.conn` (the real `requests.Session`) with a `MagicMock`; `test_geocode` mocked the census client correctly but the *entire class* was marked `@pytest.mark.live`, so it never ran in CI. These suites pass even when the connector is broken.
- **Hollow assertions.** No-op checks like `assert self.test_survey_id, table["survey_id"]` (asserts a truthy string, never looks at the result), `for i in range(0, 1)` loops that check only the first row, and `len(copy_call_args[0]) == 3` (counts a dict's keys, not the calls).
- **`unittest.TestCase` everywhere**, with `setUp`/`tearDown`, inline megabyte payloads, and `*_responses.py` data modules.
- **No safety net for rewrites.** There was no way to know whether "modernizing" a suite quietly dropped coverage or assertions — green tests don't prove equivalence.

The net effect: high line-coverage numbers that masked low real coverage, and a test suite that was hard to trust and harder to extend consistently.

## Solution

Three pieces, designed to work together:

1. **One standard** (`docs/write_tests.rst`) — a single decision rule ("mock the outermost boundary you don't own; never mock the connector's own methods"), three canonical patterns (HTTP → `requests_mock`; SDK → `mocker`; protocol/DB → a `fakes.py` fake), a fixed directory layout, and one convention for test data (`data/*.json` loaded via a shared `load` fixture).

2. **A safe, incremental migration** using **expand → bake → contract**:
   - *Expand* — write the new pytest suite next to the old one, which is renamed `test_<connector>_legacy.py` and marked `@pytest.mark.legacy`.
   - *Bake* — **both suites run in CI.** Because they exercise the same source, any source change that breaks one but not the other exposes a gap — the drift signal a frozen snapshot can't give.
   - *Contract* — once parity is proven, delete the legacy suite.

3. **Objective parity tooling** (`test/tools/parity.py`) — for each connector it records a committed baseline of **line %, branch %, and mutation score**, then fails if a migration regresses against it. Mutation testing (via cosmic-ray) is the key: it catches gutted assertions that line coverage cannot. Its `survivors` command lists exactly which behaviors no test checks, turning "is this test any good?" into a concrete to-do list. CI runs the coverage half on every PR and mutation on the connectors a PR touches (advisory today; a hard gate once the bake completes).

## Why this is the best solution

- **It makes weakening a test impossible to do by accident.** The mutation-score parity gate is the crux: you cannot "simplify" a suite into passing-but-empty without the score dropping and the gate catching it. Coverage alone would not — several suites here had ~90% line coverage and near-0% mutation score.
- **The bake removes the biggest risk of any test rewrite** — silent loss of behavior — by running old and new together against the same code instead of trusting a one-time diff.
- **It's incremental and independently reviewable.** Each connector stands alone: reviewers can approve `copper` without understanding `bigquery`. Nothing forces a big-bang cutover.
- **It leaves the suite measurably better, not just different.** The `survivors` output is used to *add* the assertions the old tests were missing, so migrations routinely raise the real (mutation) score — not merely preserve line coverage.
- **It's honest about its own state.** The tracker (`test/MIGRATION.md`) is regenerated from the tree, not hand-maintained, and calls out where a "done" directory still hides an unmigrated sub-file.

## Mutation testing, briefly (for anyone new to it)

Line coverage answers "did the tests *run* this line?" Mutation testing answers the harder question: "would a test actually *fail* if this line were **wrong**?"

The idea: a tool makes tiny, deliberate breakages in the source — one at a time — and re-runs the connector's tests against each broken version. Each broken version is a **mutant**. For example, given this bit of a connector:

```python
resp = self.conn.post(url, data=json.dumps(payload))
if resp.status_code != 201:
    raise Exception(...)
```

the tool will try mutants like flipping `!= 201` to `== 201`, changing `201` to `200`, or deleting the `raise`. For each mutant, one of two things happens:

- **Killed** — at least one test fails. Good: a test noticed the change.
- **Survived** — every test still passes. Bad: **no test checks this behavior.** That exact change could ship as a real bug and CI would stay green.

The **mutation score** is `killed / total`. A suite with 95% line coverage but a 25% mutation score is *running* almost all the code while *asserting* almost nothing about it — precisely the "hollow assertions" problem in the [Problem](#problem) section. That's why we baseline on mutation score, not just coverage.

**How we use it** (all through `test/tools/parity.py`, backed by [cosmic-ray](https://cosmic-ray.readthedocs.io/)):

1. **Baseline.** Before migrating a connector, `parity.py capture <connector>` records its killed / survived / score into `test/baselines/<connector>.json` and commits it — a frozen "this is how strong the tests were before I touched them."
2. **Guardrail.** After migrating, `parity.py compare <connector>` re-measures and **fails if the score dropped**. This is the mechanism that makes silently weakening a suite impossible: rewrite it into passing-but-empty and the score falls, so the gate trips. (A small ±2-point tolerance absorbs timing noise — cosmic-ray counts a mutant that hangs past its 60s timeout as "killed," and whether a slow mutant trips that varies run to run.)
3. **To-do list.** `parity.py survivors <connector>` prints the exact source lines whose mutants survived — i.e. the behaviors no test checks yet. We read that list and *add* the missing assertions. That's how these migrations **raised** scores (e.g. `alchemer` 25% → 88%) rather than merely preserving them: the survivor list literally tells you what to test next.

Two caveats worth knowing:

- **It's slow.** Every mutant re-runs the whole suite, and there can be hundreds per connector — so it's scoped to one connector at a time, and CI only runs it on the connectors a given PR changes.
- **Not every survivor is a real gap.** Some are *equivalent mutants* — changes that can't actually be observed (e.g. `x == "GET"` vs `x is "GET"` when the value is always the same interned string literal). Those aren't bugs waiting to happen, so we don't chase 100%; we chase the survivors that represent real, checkable behavior.

## What we've done

- **A written standard** (`docs/write_tests.rst`) and reference exemplars for each pattern: `test_airtable` (HTTP), `test_salesforce` (SDK), `test_smtp` (protocol fake).
- **Parity tooling + CI** (`test/tools/`, `.github/workflows/python-checks.yml`): `parity.py` (capture / compare / survivors), `connector_map.py`, `changed_connectors.py`, and **63 committed baselines**.
- **46 connector directories migrated** to the standard, each with a `_legacy` sibling running the old suite through the bake — **validated against their pre-migration baselines with zero regressions**.
- **Merged up to date with `main`** (0 commits behind): reconciled with upstream's concurrent work — `catalist` and `geocode` were migrated upstream in parallel (kept the stronger version), the `gmail`/`newmode` connector changes are reflected in the tests, and the removed `crowdtangle` connector was dropped. Upstream's added coverage (an ngpvan events pagination regression test, a `check_env` optional case) was ported into the new suites.
- **~150 `data/*.json`** payloads extracted from inline dicts and `*_responses.py` modules.
- **A single `load` fixture** in the root `test/conftest.py`, replacing 16 duplicated per-connector fixtures and 3 divergent module-level helpers.

Several suites were materially **strengthened**, not just moved (mutation score before → after):

| Connector | Mutation | Note |
|---|---|---|
| `ssh` | 0% → **100%** | mocked tests were dead; asserted cleanup + error path |
| `geocode` | 0% → **76%** | entire suite was `@pytest.mark.live`; revived + branch coverage |
| `alchemer` | 25% → **88%** | fixed a no-op assert; added pagination + column-transform checks |
| `sftp` | 21% → **55%** | added a `FakeSFTP` so operations run in CI, not just live |
| `twilio` | 81% → **91%** | exercised the real table conversion, not just call routing |
| `copper` | 75% → **83%** | survivor-driven request/pagination assertions |
| `controlshift` | 24% → **100%** | (earlier) hostname/pagination coverage |
| `quickbase` | 50% → **100%** | (earlier) |

The migration also **surfaced real source bugs**. Most are **flagged for a follow-up, not fixed here**, to keep this PR test-only: `geocode.geocode_address` silently ignores its `return_type` argument; `bigquery.copy_between_projects` had only self-mocking (circular) tests; `action_kit` mocked its own `.conn` and so never ran the real request/response code (the test was fixed to `requests_mock`, exposing the untested path).

**This PR makes no source changes** — it is test-only. One bug it surfaced, `airmeet.download_session_recordings` passing `sessionIds` as a request kwarg instead of `params=` (which raises `TypeError` whenever a `session_id` is supplied — the client-mock test masked it), is fixed in its own small PR, #1910. The migrated airmeet test here exercises the working path and documents that latent bug, deferring the `session_id` coverage to #1910, so this PR stays behavior-preserving.

## What there is left to do

- **Finish the migration:** `dbt` (P1); the remaining P2 HTTP suites (`scytl`, `mobilecommons`, `donorbox`); and the P3 tier (ETL/pandas, cloud/DB warehouses, `s3`, `azure`, `civis`, `facebook`, …). Tracked live in `test/MIGRATION.md`.
- **Contract:** once each connector has baked in CI with no drift, delete its `test_<connector>_legacy.py` (removes the temporary duplication this PR introduces).
- **Flip the gate:** turn parity from advisory to a hard CI failure (`--strict`) once the bake is complete.
- **Follow-up test debt** already flagged for real coverage: `bigquery.copy_between_projects`, and the source-bug fixes above.
- **Reviewer decision:** whether to land this as one integration PR or split it per-connector for upstream review (the branch is structured so either works).

## How to review

- Start with `docs/write_tests.rst` (the standard) and `test/tools/README.md` (the tooling), then `test/MIGRATION.md` (status + per-connector notes).
- Review **per connector** — each `test(<connector>): …` commit is self-contained: a new `test_<connector>.py`, its `conftest.py`/`data/`, and the `_legacy` sibling.
- The large line count is mostly (a) canned `data/*.json` payloads and (b) the temporary `_legacy` duplication that the *contract* phase removes.

## Breaking Changes

Breaking changes are changes to our public API which may require existing users to change their code. This PR is test/docs/tooling only, plus one disclosed inline bug fix (`airmeet.get_event_session_recordings`) that corrects a filter that never reached the API — it changes no public signature and requires no action from existing users.

<details open>

</details>


bmos:
Would you mind updating write_tests.rst to use the sphinx heading punctuation mark standards from _Python Developer’s Guide for documenting_? I just went through and updated all our rst files to match this (by hand).
https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections

bmos:
Since parity-mutation is so slow/power-hungry, might it make sense to run it conditionally based on if the pytest matrix passes (like we do with pytest-extras)?
Or would that make it not work, since it's calculating the difference between how the tests pass/fail?

matthewkrausse:
> Since parity-mutation is so slow/power-hungry, might it make sense to run it conditionally based on if the pytest matrix passes (like we do with pytest-extras)? Or would that make it not work, since it's calculating the difference between how the tests pass/fail?

Yes, this would work and am open to it. It would make the CI take longer because then it would only start after the tests run. I was on the fence about actually having this in the CI long term. Mostly was to help with the test suite transition.

matthewkrausse:
> Would you mind updating write_tests.rst to use the sphinx heading punctuation mark standards from _Python Developer’s Guide for documenting_? I just went through and updated all our rst files to match this (by hand). https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections

Sure, will do!

bmos:
> > Since parity-mutation is so slow/power-hungry, might it make sense to run it conditionally based on if the pytest matrix passes (like we do with pytest-extras)? Or would that make it not work, since it's calculating the difference between how the tests pass/fail?
> 
> Yes, this would work and am open to it. It would make the CI take longer because then it would only start after the tests run. I was on the fence about actually having this in the CI long term. Mostly was to help with the test suite transition.

Any thoughts on this, @shaunagm? As the person who merges the most PRs, you probably are also the one who spends the most time waiting on them passing!

bmos:
My chief concern with the 'test parity' concept is that to actually see the daylight between testing models (new vs old) we would need to be making lots of code changes (as this is what causes some tests to fail/pass).

We don't have that much code churn and this tells me that the parity testing would need to exist for a long time to truly provide the peace of mind that the new tests have parity.

bmos:
As is noted in the PR description, our tests are really not testing everything we would want them to.

As a simple example, let's look at [test_successful_post_request](https://github.com/move-coop/parsons/pull/1907/changes#diff-b18c49b4010d26ae2443b554d95810342a46c9fab30c227d1d606c82aafc53a8R26-R30) in [test_actblue](https://github.com/move-coop/parsons/pull/1907/changes#diff-b18c49b4010d26ae2443b554d95810342a46c9fab30c227d1d606c82aafc53a8).

The structure is great, using requests-mock is a fantastic choice and an under-utilized part of our dev dependencies. Unfortunately, the assertions are inadequate, leading to serious blind spots.

1. We make a post request three positional args via [the connector's method](https://github.com/move-coop/parsons/blob/main/parsons/actblue/actblue.py#L69-L112), mocking the response with requests-mock.
2. We compare one piece of the response (the ID) with a single string hardcoded in the test file.

What we have tested:

- That the connector's post method creates a post request
- That the connector's post method returns the response from requests
- That the connector's post method targets the "csvs" endpoint

What we have not tested:

- What happens to the args that are used to construct the json payload (are they sent? are they assigned to the right keys?)

All of this is not to say there's anything wrong with the PR, but more that it's the first step in a long process of test improvement. Getting away from unittest will be HUGE.

bmos:
I'll also just drop [pytest-recording](https://pypi.org/project/pytest-recording/) as a personal favorite for this type of stuff, although you may prefer your implementation.

PR #1908, opened 2026-07-30
Rate-Limited APIConnector

bmos:
Rate limits have come up with some of our connectors and some have one-off rate limit implementations included. Consolidating this via APIConnector functionality seems like a better solution to me.

Thanks to @matthewkrausse for making me aware of requests-ratelimiter, which has many helpful features.

### Quality of Life

- Improved type hinting in APIConnector
- When making multiple requests using a connector, the underlying TCP connection will be reused, which can result in a significant performance increase (see [HTTP persistent connection](https://en.wikipedia.org/wiki/HTTP_persistent_connection))

### API

- Allow passing requests.Session or requests-ratelimiter.Limiter into APIConnector
- Store headers and auth within session instead of within APIConnector (with backwards-compatibility layer)

- Defined rate limits in all connectors that I could find documentation of (included documentation link in commit messages)
- Some APIs have per-account limits. These connectors can now be initialized with a ratelimit to allow configuration of these limits.

## Breaking Changes

<details open>

</details>

### Details (if needed)


matthewkrausse:
@bmos I merged main to rerun CI on this because I was having testing failures that seemed like it was from dependency changes. Those same failures are happening here. After doing it, I realized I should have ask first — sorry! But this does reveal something is going wrong. 

Looks like upgrading to cryptography 50.0.0 caused this. 

According to a claude investigation: dbt-snowflake → snowflake-connector 4.0.0 → pyOpenSSL<26 → cryptography 50 
is the chain here 

cc: @shaunagm

bmos:
Totally okay!

austinweisgrau:
the solidarity tech stuff shouldn't be in scope here yeah?

bmos:
Ugh, bad merge on my part. I'll fix that later.

parsons/utilities/ratelimited_api_connector.py
austinweisgrau, line 1: I don't think I like the pattern of subclassing APIConnector for this. I do think at the connector class level is the right level to define rate-limiting behavior, but it would make more sense to me as an `__init__()` argument on APIConnector. If we subclass APIConnector for rate limiting, we can't compose rate limiting with other similar behaviors we might want to add to the APIConnector (e.g. concurrency limiting).

It would make more sense to me to modify the behavior of the `request` method on the APIConnector based on whether `rate_limit` is defined as an argument in `APIConnector.__init__()`

parsons/utilities/api_connector.py
```diff
@@ -49,6 +52,9 @@ def __init__(
         auth: _AuthType | None = None,
         pagination_key: str | None = None,
         data_key: str | None = None,
+        *,
+        ratelimiter: pyrate_limiter.Limiter | None = None,
```
austinweisgrau: I still feel like this is a little too highly coupled - if the rate limiter HAS to be this one particular kind, then we should just pass the actual configuration parameter directly here and instantiate the rate limiter during __init__. Callers shouldn't need to worry about instantiating a rate limiter correctly, they should just pass the rate limit configuration. e.g. I'm imagining the function signature here should be `rate_limit: int | None = None`. LMK if that makes sense

austinweisgrau: so then a caller could just use:
```
APIConnector(..., rate_limit=3)
```

bmos: The [Limiter object](https://pyratelimiter.readthedocs.io/en/latest/modules/pyrate_limiter.limiter.html#pyrate_limiter.limiter.Limiter) is a much more complex object allowing more configuration. Within a Limiter you can, for example, configure multiple buckets (as some APIs have different limits for different endpoints).

Within that there is a [Rate object](https://pyratelimiter.readthedocs.io/en/latest/modules/pyrate_limiter.abstracts.rate.html#pyrate_limiter.abstracts.rate.Rate) which also allows more configuration, including a count, an interval, and optional burst limits.

bmos: Obviously we could make it work if we hard coded the interval and required the integer to always specify number of requests per that interval of time, but it feels like it loses a lot of configurability.

bmos: One possibility that would help with how tightly coupled this is (but make configuration more complicated, IMO) would be to just accept a session and leave it up to the user to configure a session with a rate limiter.

austinweisgrau: Do we have known examples among the APIs in this repo where a per second configuration wouldn't work?

bmos: No, a per second rate should work everywhere.
But sometimes it is nice to burn through your entire rate limit in a shorter period of time than you would if you metered it by seconds.

Like if you can make 600 requests an hour, but you have to enter your rate limit in requests per second, you can't make 600 requests in a single minute and then stop.

bmos: What if I allow users to pass EITHER an int (assuming seconds) OR a Limiter object?

bmos: @austinweisgrau Do you have any further thoughts on where I've gone with this?
You can now pass in Limiter, Rate, or int: https://github.com/bmos/parsons/blob/7dd1ad7575e354fc3f2ea126e1dadf303a0ef611/parsons/utilities/api_connector.py#L42-L96

PR #1910, opened 2026-07-31
fix(airmeet): send session-recordings sessionIds as a query param

matthewkrausse:
## Problem

`Airmeet.download_session_recordings(airmeet_id, session_id=...)` forwarded `sessionIds` to `APIConnector.get_request` as a **keyword argument** instead of a query parameter:

```python
kwargs["sessionIds"] = session_id
self.client.get_request(url=..., **kwargs)
```

`get_request` passes unknown kwargs straight down to `requests`, so calling the method **with** a `session_id` raises `TypeError: Session.request() got an unexpected keyword argument 'sessionIds'` — the per-session filter never worked. The existing unit test masked this by mocking the entire client method, so it only checked the call shape, not real behavior.

## Fix

Pass it as a query param, matching every other method in the connector (`_get_all_pages` already uses `params=`):

```python
params["sessionIds"] = session_id
self.client.get_request(url=..., params=params)
```

Updated the existing unit test to assert the corrected call.

## Context

Surfaced while migrating the Parsons test suite (#1907). Kept as its own small PR so that migration stays test-only.

## Breaking Changes

### DIFF
--- a/parsons/airmeet/airmeet.py
+++ b/parsons/airmeet/airmeet.py
@@ -419,10 +419,12 @@ class Airmeet:
                 List of session recordings
 
         """
-        kwargs = {}
+        params = {}
         if session_id:
-            kwargs["sessionIds"] = session_id
-        response = self.client.get_request(url=f"airmeet/{airmeet_id}/session-recordings", **kwargs)
+            params["sessionIds"] = session_id
+        response = self.client.get_request(
+            url=f"airmeet/{airmeet_id}/session-recordings", params=params
+        )
         return Table(response["recordings"])
 
     def fetch_event_replay_attendance(self, airmeet_id, session_id=None) -> Table:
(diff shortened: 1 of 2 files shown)


bmos (approved):
Great catch! While the core issue here was that it was being passing `sessionIds` within kwargs, I think this became more of an issue when I started passing kwargs to `requests.request()` in May.

https://github.com/move-coop/parsons/commit/a27aae400cbf79331dde53ab139e82553e822ddb#diff-ab7b00dc9849256c403202d50e1bff6c148eb8e98ba37650684755d8e1f16350R136

bmos:
The tests on this should pass if re-run.

shaunagm (approved):

shaunagm:
@matthewkrausse, can you make this a non-draft so it can be merged?

PR #1918, opened 2026-08-01
Fix google issues after recent cryptography release

bmos:
Thanks to @matthewkrausse for pointing out the issue.

## What is this change?

- Installing the latest version of dbt-snowflake results in the installation of cryptography 50.0.0 which is causing issues with some Google connectors when using python 3.10 on all platforms and (on windows) python 3.14. I assume this is due to google-auth which relies on the cryptography package.
This PR restricts python 3.10 to `cryptography >=38, < 50`. As this incompatibility is not mentioned in any changelogs I have found, hopefully this will be resolved upstream and we can remove this constraint later.
- Improves consistency of dependency syntax with spaces around all operators

- An alternate solution is limiting dbt-snowflake to 1.11, but since cryptography is the actual culprit and it's breaking connectors that don't relate to dbt-snowflake, this seemed like a better solution.

- ci tests pass

## Breaking Changes

<details open>

</details>

### Details (if needed)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,10 +75,10 @@ dbt-postgres = [
     "rich >= 14.0"
 ]
 dbt-snowflake = [
-    "dbt-snowflake >= 1.8,<1.12;python_version < '3.11'",
-    "dbt-snowflake >= 1.8;python_version >= '3.11'",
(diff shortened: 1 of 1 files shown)


bmos:
This appears to have been resolved upstream

PR #1922, opened 2026-08-05
Auto-generate .pyi stubs file for __init__.py

bmos:
Modern IDEs have auto-completion, type inference and other features that rely on static code analysis.

Due to parsons' lazy import design, which enabled our use of extras to minimize the number of required dependencies, static analysis is not possible when importing connectors the way we recommend (`from parsons import GitHub`) and requires importing via the full path.

## What is this change?
- Created .pyi stubs file to allow static code analysis
- Added a script to generate the file programmatically
- Automatically run script via pre-commit
- Added CI workflow to check if the stubs file is up to date
- Adjusted all code imports in `parsons/` to import from specific files
- Adjusted all code imports in `test/` to import from parsons

- Additional complexity by including a type stubs file.

- See that `Type Stubs` workflow passes.
- Open a file that imports from parsons (such as `test/test_actblue/test_actblue.py`) in an IDE and hover your mouse over `ActBlue` or `Table` in the imports section (details may vary between IDEs). Compare this against the behavior on the main branch.

## Breaking Changes

<details open>

</details>

### Details (if needed)


parsons/__init__.py
```diff
@@ -1,30 +1,22 @@
...
 logger = logging.getLogger(__name__)
 _handler = logging.StreamHandler()
 _formatter = logging.Formatter("%(module)s %(levelname)s %(message)s")
 _handler.setFormatter(_formatter)
 logger.addHandler(_handler)
 
-if os.environ.get("TESTING"):
-    # Log less stuff in automated tests
-    logger.setLevel("WARNING")
-elif os.environ.get("DEBUG"):
```
bmos, line 22: Removed `TESTING` as it is not used (pytest sets `PYTEST_CURRENT_TEST`)

parsons/__init__.py
```diff
@@ -1,30 +1,22 @@
...
+if os.environ.get("DEBUG") or os.environ.get("RUNNER_DEBUG"):
```
bmos, line 15: Added `RUNNER_DEBUG` so that re-running failed tests via Github Actions with Debug Logging enabled will increase verbosity of log outputs.

parsons/__init__.py
```diff
@@ -126,3 +119,7 @@ def __getattr__(name):
             "Only core dependencies are installed by default. Learn more: "
             "https://www.parsonsproject.org/pub/improving-the-parsons-installation-experience"
         ) from e
+
+
+def __dir__() -> list[str]:
+    return sorted(set(globals().keys()) | set(_CONNECTORS.keys()))
```
bmos, line 125: Ensures that everything in `__init__` will be visible

shaunagm (approved):

PR #1927, opened 2026-08-20
Fix occasional setup-uv CI failures on windows and macos

bmos:
## What is this change?

- We keep having random intermittent test failures during the setup-uv action step. These failures are because windows and macos runners don't support eBPF. This means that during initialization our firewall is configured with IPs matching our whitelist but those IPs do not always match those that are contacted during setup-up (the github IPs rotate, so multiple DNS requests may get different results).
- As we are no longer blocking unexpected outbound traffic, we should be careful that our tests do not result in actual HTTP traffic.

- We could conditionally use block on the ubuntu runners, which could help reveal tests that are making HTTP requests, but it might be confusing to have improperly-written tests failing with timeouts on ubuntu but not macos and windows.
- I didn't remove the whitelist, as this deficiency of harden-runner may be resolved at some point in the future.

- Observe that tests pass.

## Breaking Changes

<details open>

</details>

### Details (if needed)

### DIFF
--- a/.github/workflows/python-checks.yml
+++ b/.github/workflows/python-checks.yml
@@ -20,11 +20,10 @@ env:
   ENDPOINT_WHITELIST: >-
     pypi.org:443
     github.com:443
-    releases.astral.sh
+    releases.astral.sh:443
     files.pythonhosted.org:443
     *.github.com:443
     *.githubusercontent.com:443
-    raw.githubusercontent.com/astral-sh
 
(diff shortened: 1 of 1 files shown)


bmos:
I'm reopening this. While updating setup-uv has helped work around this issue, it does still happen sometimes. The approach in this PR is the only real fix I know of.

shaunagm (approved):

PR #1936, opened 2026-08-28
Forward return_type in CensusGeocoder.geocode_address

### LINKED ISSUE #1929: [Bug] CensusGeocoder.geocode_address ignores its return_type argument
## Detailed Description

`CensusGeocoder.geocode_address` accepts a `return_type` argument and documents it
("`geographies` will return information about the Census geographies while `locations` will
information about the address"), but never forwards it to the underlying library.

The call is `self.cg.address(address_line, city=city, state=state, zipcode=zipcode)` with no
`returntype=`, so `censusgeocode` falls back to its `"geographies"` default and always requests
`https://geocoding.geo.census.gov/geocoder/geographies/address`. A caller asking for
`"locations"` silently gets geographies data back.

The sibling method `geocode_onelineaddress` does forward it, so the two methods disagree.

## To Reproduce

```python

cg = CensusGeocoder()
geo = cg.geocode_address(
    "1600 Pennsylvania Avenue", city="Washington", state="DC", return_type="locations"
)

# Expected: locations responses carry no nested "geographies" key
# Actual:   True -- the geographies endpoint was queried instead
print("geographies" in geo[0])
```

For contrast, the same request against the two endpoints directly:

| returntype | keys on each address match |
|---|---|
| `geographies` | `addressComponents`, `coordinates`, `geographies`, `matchedAddress`, `tigerLine` |
| `locations` | `addressComponents`, `coordinates`, `matchedAddress`, `tigerLine` |

## Your Environment

## Additional Context

The existing `test_geocode_address` asserts only the return value, and its mock returns whichever
fixture the test hands it, so the dropped argument was invisible. `test_geocode_onelineaddress`
already asserts call arguments via `assert_called_with`, which is why the sibling stayed correct.

This is Parsons-side only -- the Census Geocoding Services API defines returntype as a path
segment (`/geocoder/{returntype}/{searchtype}`) and `censusgeocode` forwards `returntype` through
`**kwargs` correctly.

## Priority

Low. It is a small, self-contained fix and I have a patch ready, so this is not time-sensitive.

JakeSelby:
## What is this change?

- `CensusGeocoder.geocode_address` accepted a `return_type` argument and never passed it on, so
  `censusgeocode` always fell back to its `"geographies"` default and a caller asking for
  `"locations"` silently received geographies data. The sibling `geocode_onelineaddress` already
  forwards it.
- `test_geocode_address` now asserts the call arguments as well as the return value, which is what
  the sibling test already did and why this went unnoticed.
- Fixes #1929.

- No change to the return type. Both return types produce `result.addressMatches` for the
  `address` searchtype, so `censusgeocode` builds an `AddressResult` either way -- only the
  presence of the nested `geographies` key on each match differs. `GeographyResult` is returned
  only by the `coordinates` searchtype, which this PR does not touch.
- The one behavior change is that `return_type="locations"` now returns matches without that
  nested `geographies` key, which is what the parameter has always been documented to do.

```
```

To confirm against the live API, `geocode_address(..., return_type="locations")` should now return
matches with no `"geographies"` key, matching `geocode_onelineaddress` with the same argument.

## Breaking Changes

<details open>

</details>

### Details (if needed)

- No signature or return-type change. Code passing `return_type="geographies"` (the default) is
  unaffected. Only code that passed `"locations"` and then read the nested `"geographies"` key --
  i.e. code relying on the bug -- would need to switch to `"geographies"` explicitly.

### DIFF
--- a/parsons/geocode/census_geocoder.py
+++ b/parsons/geocode/census_geocoder.py
@@ -81,7 +81,9 @@ class CensusGeocoder:
             dict
 
         """
-        geo = self.cg.address(address_line, city=city, state=state, zipcode=zipcode)
+        geo = self.cg.address(
+            address_line, city=city, state=state, zipcode=zipcode, returntype=return_type
+        )
         self._log_result(geo)
         return geo
 
--- a/test/test_geocode/test_census_geocoder.py
+++ b/test/test_geocode/test_census_geocoder.py
@@ -43,11 +43,25 @@ def test_geocode_address(cg):
     # Assert one line with geographies parameter returns expected
     cg.cg.address = mock.MagicMock(return_value=geographies_resp)
     geo = cg.geocode_address(**passed_address, return_type="geographies")
+    cg.cg.address.assert_called_with(
+        passed_address["address_line"],
+        city=passed_address["city"],
+        state=passed_address["state"],
+        zipcode=None,
+        returntype="geographies",
+    )
     assert geo == geographies_resp
 
     # Assert one line with locations parameter returns expected
     cg.cg.address = mock.MagicMock(return_value=locations_resp)
     geo = cg.geocode_address(**passed_address, return_type="locations")
+    cg.cg.address.assert_called_with(
+        passed_address["address_line"],
+        city=passed_address["city"],
+        state=passed_address["state"],
+        zipcode=None,
+        returntype="locations",
+    )
     assert geo == locations_resp
 
 


bmos (approved):
Thank you for addressing (no pun intended) this! I have noticed it before while looking at the code, but hadn't found time to test things out with the API to make a fix and ensure it all worked as intended.

shaunagm (approved):

PR #1937, opened 2026-08-28
Add a request timeout to CensusGeocoder

### LINKED ISSUE #1930: [Bug] CensusGeocoder sets no request timeout, so a call can hang indefinitely
## Detailed Description

`CensusGeocoder` sets no request timeout and gives callers no way to set one, so every request is
made with `timeout=None` and can block indefinitely if the Census endpoint stalls.

`censusgeocode` supports it -- both `_fetch` and `_post_batch` pass
`timeout=kwargs.get("timeout")` straight through to `requests` -- but Parsons never supplies a
value, and exposes no parameter through which one could be supplied. All four public methods are
affected: `geocode_onelineaddress`, `geocode_address`, `geocode_address_batch` and
`get_coordinates_data`.

A geocoding job is normally one step in a scheduled pipeline. Without a timeout a stalled endpoint
does not fail the step, it hangs it, holding a worker until something external kills it. That is
worse than a clean failure, because a failure retries on the next run while a hang may not surface
until someone notices the schedule stopped. It matters most in `geocode_address_batch`, where one
stalled chunk holds up a job that may be partway through many.

## To Reproduce

```python
from unittest import mock

seen = {}
def capture(url, **kw):
    seen["timeout"] = kw.get("timeout", "<absent>")
    raise SystemExit

cg = CensusGeocoder()
with mock.patch("requests.get", side_effect=capture):
    try:
        cg.geocode_address("1600 Pennsylvania Avenue", city="Washington", state="DC")
    except SystemExit:
        pass

print(seen)  # {'timeout': None} -- no timeout on any request
```

The same holds for `requests.post` via `geocode_address_batch`, and for the other two methods.

## Your Environment

## Additional Context

There is no workaround available to a user today: the parameter does not exist, and
`CensusGeocoder` does not expose the underlying client in a way that lets a timeout be injected
per call.

Several other connectors already take this as a constructor argument -- `Redshift`, `Postgres` and
`MySQL` use `timeout=10`, `Braintree` an effective 200 -- so there is an established pattern to
follow.

The one open question is the default. I have a patch that adds the parameter with a `None`
default, which preserves current behavior exactly since `None` is what `requests` already
receives; the change only makes a timeout possible. A real default would match those other
connectors, but a batch of up to `BATCH_SIZE` addresses can legitimately run long and I have no
data on a safe ceiling. Happy to set one if maintainers have a preference.

## Priority

Low. Small self-contained fix, patch ready, and nothing changes for existing users unless they opt
in.

JakeSelby:
## What is this change?

- `CensusGeocoder` made every request with no timeout. `censusgeocode` passes
  `timeout=kwargs.get("timeout")` through to `requests`, and Parsons never supplied one, so all
  four public methods reached `requests` with `timeout=None`. A stalled Census endpoint blocks the
  caller indefinitely with no way to opt out.
- Adds a `timeout` constructor argument and forwards it to `onelineaddress`, `address`,
  `addressbatch` and `coordinates`.
- Fixes #1930.

- **The default is `None`, which preserves today's behavior exactly** -- `timeout=None` is
  precisely what `requests` already received, so nothing changes for existing callers. The
  parameter makes it possible to set a timeout, which is currently impossible.
- Other connectors ship a real default instead (`Redshift`, `Postgres` and `MySQL` use
  `timeout=10`; `Braintree` uses 200). I did not pick one here because a batch of up to
  `BATCH_SIZE` addresses can legitimately run long and I have no data on a safe ceiling. If you
  would rather this shipped with a real default, say the number and I will change it.
- Put on the constructor rather than each method, matching those same connectors.

```
```

`test_timeout_forwarded_to_every_request` covers all four methods. To confirm at the wire, patch
`requests.get` / `requests.post` and check the `timeout` keyword: `CensusGeocoder()` sends
`None` as before, `CensusGeocoder(timeout=30)` sends `30`.

## Breaking Changes

<details open>

</details>

### Details (if needed)

- New optional keyword argument with a default that reproduces the previous call exactly. No
  signature change for existing callers, no return-type change.

### DIFF
--- a/parsons/geocode/census_geocoder.py
+++ b/parsons/geocode/census_geocoder.py
@@ -25,11 +25,16 @@ class CensusGeocoder:
         vintage: str
             The US Census vintage file to utilize. By default the current vintage is used, but
             other options can be found `here <https://geocoding.geo.census.gov/geocoder/vintages?form>`__.
+        timeout: float
+            Seconds to wait for a response from the Census API before raising
+            ``requests.exceptions.Timeout``. Applies to every request this class makes.
+            By default no timeout is set, and a request can block indefinitely.
 
     """
 
-    def __init__(self, benchmark="Public_AR_Current", vintage="Current_Current"):
+    def __init__(self, benchmark="Public_AR_Current", vintage="Current_Current", timeout=None):
         self.cg = censusgeocode.CensusGeocode(benchmark=benchmark, vintage=vintage)
+        self.timeout = timeout
 
     def geocode_onelineaddress(self, address, return_type="geographies"):
         """
@@ -48,7 +53,7 @@ class CensusGeocoder:
             dict
 
         """
-        geo = self.cg.onelineaddress(address, returntype=return_type)
+        geo = self.cg.onelineaddress(address, returntype=return_type, timeout=self.timeout)
         self._log_result(geo)
         return geo
 
@@ -81,7 +86,9 @@ class CensusGeocoder:
             dict
 
         """
-        geo = self.cg.address(address_line, city=city, state=state, zipcode=zipcode)
+        geo = self.cg.address(
+            address_line, city=city, state=state, zipcode=zipcode, timeout=self.timeout
+        )
         self._log_result(geo)
         return geo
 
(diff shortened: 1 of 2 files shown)


shaunagm (approved):

shaunagm:
I attempted to resolve a merge conflict and messed up the tests - sorry! I think the tests need to be updated to accept the new timeout parameter.

PR #1940, opened 2026-08-28
Add optional retry for transient CensusGeocoder failures

### LINKED ISSUE #1931: [Feature/Addition] Optional retry for transient CensusGeocoder request failures
## Detailed Description

`CensusGeocoder` makes exactly one attempt per request and gives callers no way to ask for
another. `censusgeocode` re-raises `requests.exceptions.RequestException` from both `_fetch` and
`_post_batch`, and `CensusGeocoder` does not catch it, so any connection reset, DNS blip or
timeout ends the call.

```python
from unittest import mock
import requests

with mock.patch("requests.get", side_effect=requests.exceptions.ConnectionError("reset")):
    CensusGeocoder().geocode_onelineaddress("1600 Pennsylvania Avenue, Washington, DC")
# ConnectionError -- one attempt, no retry, no way to configure one
```

## Context

Transient network failures against a public government endpoint are routine, and the cost of one
is disproportionate in `geocode_address_batch`: it chunks the input and loops, so a failure on
chunk N ends the job, and because the accumulator is a local variable every chunk already
geocoded is discarded with it. On a large table that is hours of work lost to a single dropped
connection.

A user cannot work around this without reimplementing the chunking loop themselves.

## Possible Implementation

A `retries` constructor argument and a small private helper that retries connection-level
failures with `2 ** attempt` backoff, wrapping the per-chunk call in `geocode_address_batch` so
retries never re-send completed chunks.

Two things worth deciding before this lands:

1. **What can actually be retried.** Only connection-level failures can be handled cleanly.
   `censusgeocode` never calls `raise_for_status()`, so a Census 5xx does not raise a
   `RequestException` -- for single-address lookups it surfaces as
   `ValueError("Unable to parse response from Census")`, indistinguishable from a genuine parse
   problem. Retrying that would also retry real bugs, so a retry here can only cover
   `RequestException`. (The status-handling gap is worth its own issue.)
2. **Whether a shared helper is wanted.** `Newmode` and `NationBuilder` each hand-roll their own
   retry loop today. If maintainers would rather have one utility than a fourth per-connector
   implementation, that is better settled first.

A default of `0` preserves current behavior exactly. `Newmode` defaults to 2 if you would prefer
retry on by default.

## Priority

Low. Patch ready, opt-in, nothing changes for existing users by default.

JakeSelby:
## What is this change?

- `censusgeocode` re-raises `RequestException` and `CensusGeocoder` never caught it, so a single
  connection reset or timeout ended the whole call. For `geocode_address_batch` that means a long
  job dies partway through its chunks.
- Adds a `retries` constructor argument and a `_request` helper that retries connection-level
  failures with `2 ** attempt` backoff. The retry wraps the **per-chunk** call in
  `geocode_address_batch`, so completed chunks are never re-sent.
- Fixes #1931.

- **The default is `0`, which preserves today's behavior exactly.** Retrying by default would
  change request timing for every existing user and could multiply load on a government endpoint,
  so it seemed wrong to opt everyone in. `Newmode` defaults to 2 retries if you would rather this
  matched that.
- **Only `RequestException` is retried, and that is a real limit worth naming.** `censusgeocode`
  never calls `raise_for_status()`, so the Census API's own error responses do not raise: a 5xx
  reaches `_fetch` as a JSON decode failure and surfaces as
  `ValueError("Unable to parse response from Census")`, which is indistinguishable from a genuine
  parse problem. Retrying `ValueError` would therefore also retry real bugs, so I left it alone.
  Making HTTP status failures visible is a separate change and I have filed it separately.
- Put on the constructor rather than each method, matching `Redshift`, `Postgres`, `MySQL` and
  `Braintree`.

```
```

Four new tests cover recovery after transient failures, re-raising once retries are exhausted,
the no-retry default, and that all four public methods are wrapped.

## Breaking Changes

<details open>

</details>

### Details (if needed)

- New optional keyword argument defaulting to `0`, which makes exactly one attempt and never
  sleeps -- identical to current behavior. No signature or return-type change.

### DIFF
--- a/parsons/geocode/census_geocoder.py
+++ b/parsons/geocode/census_geocoder.py
@@ -1,8 +1,10 @@
 import logging
+import time
 
 import censusgeocode
 import petl
 from censusgeocode.censusgeocode import GeographyResult
+from requests.exceptions import RequestException
 
 from parsons import Table
 
@@ -25,11 +27,35 @@ class CensusGeocoder:
         vintage: str
             The US Census vintage file to utilize. By default the current vintage is used, but
             other options can be found `here <https://geocoding.geo.census.gov/geocoder/vintages?form>`__.
+        retries: int
+            Number of additional attempts to make when a request fails with a transient
+            network error, waiting ``2 ** attempt`` seconds in between. Only connection-level
+            failures are retried; the Census API's own error responses do not raise. By
+            default no retry is attempted.
 
     """
 
-    def __init__(self, benchmark="Public_AR_Current", vintage="Current_Current"):
+    def __init__(self, benchmark="Public_AR_Current", vintage="Current_Current", retries=0):
+        if retries < 0:
+            msg = f"retries must be 0 or greater, got {retries}"
+            raise ValueError(msg)
         self.cg = censusgeocode.CensusGeocode(benchmark=benchmark, vintage=vintage)
+        self.retries = retries
+
+    def _request(self, func, *args, **kwargs):
+        # Retries transient network failures only; see the class docstring for the limits.
+        for attempt in range(self.retries + 1):
+            try:
+                return func(*args, **kwargs)
+            except RequestException as error:
+                if attempt == self.retries:
+                    raise
+                wait = 2**attempt
+                logger.warning(
+                    f"Census request failed with {type(error).__name__}; retrying in "
+                    f"{wait}s ({attempt + 1} of {self.retries})."
+                )
+                time.sleep(wait)
 
     def geocode_onelineaddress(self, address, return_type="geographies"):
         """
@@ -48,7 +74,7 @@ class CensusGeocoder:
             dict
 
         """
-        geo = self.cg.onelineaddress(address, returntype=return_type)
+        geo = self._request(self.cg.onelineaddress, address, returntype=return_type)
         self._log_result(geo)
         return geo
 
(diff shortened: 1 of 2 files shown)


bmos:
This is an alternate implementation of #1943, which is the preferred approach?

parsons/geocode/census_geocoder.py
```diff
@@ -25,11 +27,35 @@ class CensusGeocoder:
...
-    def __init__(self, benchmark="Public_AR_Current", vintage="Current_Current"):
+    def __init__(self, benchmark="Public_AR_Current", vintage="Current_Current", retries=0):
+        if retries < 0:
+            msg = f"retries must be 0 or greater, got {retries}"
+            raise ValueError(msg)
         self.cg = censusgeocode.CensusGeocode(benchmark=benchmark, vintage=vintage)
+        self.retries = retries
+
+    def _request(self, func, *args, **kwargs):
+        # Retries transient network failures only; see the class docstring for the limits.
```
bmos, line 46:
```suggestion
    def _request(self, func, *args, **kwargs):
        """
        Facilitate retrying requests when network failures occur.

        Failed requests resulting in retries will be logged.
        See class docstring for more information.

        Args:
            func (Callable): function to wrap with retry logic

        Returns:
            Value returned by provided callable

        """
```

Doesn't have to be exactly that, but a docstring here would be nice.

bmos (changes requested):

PR #1941, opened 2026-08-28
Make the geocode batch size configurable

### LINKED ISSUE #1933: [Feature/Addition] Make the CensusGeocoder batch size configurable
## Detailed Description

`geocode_address_batch` chunks at a hardcoded `BATCH_SIZE = 999`, with no way to change it. The
comment above the constant reads "Currently the recommendation is less than 1K records" and gives
no source.

The Census documents an upper limit of **10,000** records per batch file. `censusgeocode` quotes
that figure directly in `addressbatch`'s docstring -- "there is currently an upper limit of 10,000
records per batch file" -- warns only above 10,001, and does no chunking of its own (there is no
chunking code in the package at all).

So Parsons issues roughly ten times as many requests as the API requires.

## Context

Request count is the thing a caller most wants control over when geocoding a large table. More
requests means more wall-clock time, more exposure to transient failures, and more load on a
public endpoint -- and right now the figure is fixed at a value that appears to be an order of
magnitude below the documented ceiling.

Different users are also in genuinely different positions: someone geocoding a few thousand rows
interactively and someone running millions overnight do not want the same chunk size, and neither
can express that today.

## Possible Implementation

A `batch_size` constructor argument used in place of the constant, plus a correction to the
comment so it cites the documented limit rather than an unsourced recommendation.

Keeping the default at `BATCH_SIZE` leaves request volume unchanged for everyone. **Whether the
default itself should move toward 10,000 is a separate question and I would rather not assume
it** -- the existing comment suggests someone may have had a reason that predates the current API
documentation, and I have no data on how the endpoint behaves near the limit. Happy to change the
default in the same patch if you want it moved.

Worth noting this is safer once some retry handling exists, since larger chunks mean more lost
work per failure.

## Priority

Low.

JakeSelby:
## What is this change?

- `geocode_address_batch` chunked at a hardcoded `BATCH_SIZE = 999`, above a comment claiming
  "the recommendation is less than 1K records" with no source given.
- The Census documents an upper limit of **10,000** records per batch file. `censusgeocode` quotes
  that figure verbatim in `addressbatch`'s docstring ("there is currently an upper limit of 10,000
  records per batch file"), warns only above 10,001, and does no chunking of its own. So Parsons
  sends roughly ten times as many requests as the API requires.
- Adds a `batch_size` constructor argument and corrects the comment to cite the documented limit.
- Fixes #1933.

- **The default stays at `999`, so request volume is unchanged** unless a caller opts in. Whether
  that default should move toward 10,000 is your call, not mine -- I have no data on how the
  endpoint behaves near the limit, and the existing comment suggests someone may have had a reason
  that predates the current API docs. Happy to change it in this PR if you want it moved.
- Put on the constructor rather than on `geocode_address_batch`, matching `Redshift`, `Postgres`,
  `MySQL` and `Braintree`. A per-call argument would also be reasonable if you prefer that.
- `BATCH_SIZE` stays exported as the default, so anything importing it is unaffected.

```
```

Two new tests: the default still equals `BATCH_SIZE` and chunks as before, and a configured
`batch_size` produces the expected chunk sizes.

## Breaking Changes

<details open>

</details>

### Details (if needed)

- New optional keyword argument defaulting to the existing `BATCH_SIZE` constant. Same chunking,
  same request count, same output for every existing caller.

### DIFF
--- a/parsons/geocode/census_geocoder.py
+++ b/parsons/geocode/census_geocoder.py
@@ -9,8 +9,9 @@ from parsons import Table
 logger = logging.getLogger(__name__)
 
 
-# The size of batches to send to the batch geocode endpoint. Currently
-# the recommendation is less than 1K records.
+# Default number of records sent per batch geocode request. The Census documents an upper limit
+# of 10,000 records per batch file, and censusgeocode does no chunking of its own, so this
+# default is deliberately conservative. Override it with the ``batch_size`` argument.
 BATCH_SIZE = 999
 
 
@@ -25,11 +26,21 @@ class CensusGeocoder:
         vintage: str
             The US Census vintage file to utilize. By default the current vintage is used, but
             other options can be found `here <https://geocoding.geo.census.gov/geocoder/vintages?form>`__.
+        batch_size: int
+            Number of records sent per request by :meth:`geocode_address_batch`. The Census
+            documents an upper limit of 10,000 records per batch file. Defaults to
+            ``BATCH_SIZE``.
 
     """
 
-    def __init__(self, benchmark="Public_AR_Current", vintage="Current_Current"):
+    def __init__(
+        self, benchmark="Public_AR_Current", vintage="Current_Current", batch_size=BATCH_SIZE
+    ):
+        if batch_size < 1:
+            msg = f"batch_size must be 1 or greater, got {batch_size}"
+            raise ValueError(msg)
         self.cg = censusgeocode.CensusGeocode(benchmark=benchmark, vintage=vintage)
+        self.batch_size = batch_size
 
     def geocode_onelineaddress(self, address, return_type="geographies"):
         """
@@ -117,7 +128,7 @@ class CensusGeocoder:
             )
             raise ValueError(msg)
 
-        chunked_tables = table.chunk(BATCH_SIZE)
+        chunked_tables = table.chunk(self.batch_size)
         records_processed = 0
 
         geocoded_tbl = Table([[]])
--- a/test/test_geocode/test_census_geocoder.py
+++ b/test/test_geocode/test_census_geocoder.py
@@ -4,6 +4,7 @@ import petl
 import pytest
 
 from parsons import CensusGeocoder, Table
+from parsons.geocode.census_geocoder import BATCH_SIZE
 from test.conftest import assert_matching_tables
 
 from .test_responses import batch_resp, coord_resp, geographies_resp, locations_resp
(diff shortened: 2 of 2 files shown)


parsons/geocode/census_geocoder.py
```diff
@@ -25,11 +26,21 @@ class CensusGeocoder:
         vintage: str
             The US Census vintage file to utilize. By default the current vintage is used, but
             other options can be found `here <https://geocoding.geo.census.gov/geocoder/vintages?form>`__.
+        batch_size: int
+            Number of records sent per request by :meth:`geocode_address_batch`. The Census
+            documents an upper limit of 10,000 records per batch file. Defaults to
+            ``BATCH_SIZE``.
```
bmos, line 32: "Defaults to ``BATCH_SIZE``" seems potentially unclear, as double ticks are used in sphinx (and markdown) to indicate a string literal. I'd recommend replacing that with an explicit reference to the constant.

```suggestion
        batch_size: int
            Number of records sent per request by :meth:`geocode_address_batch`.
            The Census documents an upper limit of 10,000 records per batch file.
            Defaults to value of :const:`BATCH_SIZE`.
```

To enable link to constant, you must also add this to the end of [docs/census_geocoder.rst](https://github.com/move-coop/parsons/blob/main/docs/census_geocoder.rst?plain=1).

```rst
.. py:data:: parsons.geocode.census_geocoder.BATCH_SIZE
   :type: int
   
   Default number of records sent per batch geocode request.
```

parsons/geocode/census_geocoder.py
```diff
@@ -25,11 +26,21 @@ class CensusGeocoder:
...
 
     """
 
-    def __init__(self, benchmark="Public_AR_Current", vintage="Current_Current"):
+    def __init__(
+        self, benchmark="Public_AR_Current", vintage="Current_Current", batch_size=BATCH_SIZE
+    ):
+        if batch_size < 1:
+            msg = f"batch_size must be 1 or greater, got {batch_size}"
+            raise ValueError(msg)
```
bmos, line 41:
```suggestion
        if not isinstance(batch_size, int):
            msg = f"batch_size must be an integer, got {type(batch_size).__name__}"
            raise TypeError(msg)
        if batch_size < 1:
            msg = f"batch_size must be 1 or greater, got {batch_size}"
            raise ValueError(msg)
```
If catching that edge case, perhaps this also makes sense?
Could be cleaner to just add a type hint to the argument, although the others currently lack them as well.

bmos (changes requested):
Good improvement to the connector, thank you!

PR #1942, opened 2026-08-28
Allow geocode_address_batch to send chunks in parallel

### LINKED ISSUE #1934: [Feature/Addition] Optional concurrency for geocode_address_batch
## Detailed Description

`geocode_address_batch` sends chunks strictly sequentially, waiting for each round trip before
starting the next:

```python
for tbl in chunked_tables:
    geocoded_tbl.concat(Table(petl.fromdicts(self.cg.addressbatch(tbl))))
```

For a large table the wall-clock time is dominated by that serialisation, and there is no way for
a caller to opt into sending several chunks at once.

## Context

At `BATCH_SIZE = 999`, a million-row table is over a thousand sequential round trips. The work is
entirely I/O bound and embarrassingly parallel -- each chunk is independent -- so the current
shape leaves a lot on the table for users running large jobs.

That said, **the important part of this proposal is the default, not the feature.** Parsons runs
inside other organisations' production pipelines and the Census geocoder is a government
endpoint. A concurrency increase that arrives with a version bump is not something users should
inherit; it should be a deliberate choice by whoever operates a given pipeline.

So the ask is narrowly for the *option*, with sequential remaining the default.

## Possible Implementation

A `workers` constructor argument defaulting to 1, where the default path is the existing
sequential code rather than a one-worker pool, so request timing and log interleaving are
unchanged. Above 1, chunks go through a `ThreadPoolExecutor`.

`Executor.map` yields results in input order, so output row order is preserved regardless of
completion order. I have verified this holds for lazily-evaluated table sources
(`Table.from_csv`) as well as in-memory ones.

One behavior difference to be aware of: with `workers > 1` all chunks are submitted before any
result is consumed, so a failing chunk surfaces before any rows are concatenated.

This is best considered after retry handling exists -- more concurrent requests means more
exposure to transient failures, which currently end the whole job.

## Priority

Low. This is the least certain of the geocoder items I have raised and the largest change, so it
should be last if it is wanted at all.

JakeSelby:
## What is this change?

- `geocode_address_batch` sent every chunk sequentially, so a large table waits on one round trip
  at a time.
- Adds a `workers` constructor argument. Above 1, chunks are sent through a `ThreadPoolExecutor`.
- Fixes #1934.

- **The default is 1 and I would argue it should stay 1.** Raising it increases the load Parsons
  places on a government API. That is a decision for whoever operates a given pipeline, not
  something users should inherit by upgrading.
- **The default path is unchanged, not merely equivalent.** With `workers=1` chunks are still
  requested lazily through a generator, one immediately before its result is concatenated, so
  request timing and interleaving with the progress log are identical to today.
- **Row order is preserved.** `Executor.map` yields results in input order, so the output table
  matches the input regardless of completion order.
- One behavior difference worth naming: with `workers > 1` all chunks are submitted before any
  result is consumed, so a failing chunk surfaces before any rows are concatenated. With
  `workers=1` the existing behavior is kept.

```
```

Three new tests: the default runs entirely on the calling thread, `workers=3` genuinely runs three
chunks concurrently (enforced with a `threading.Barrier`, so it cannot pass by accident), and
output order matches input order.

## Breaking Changes

<details open>

</details>

### Details (if needed)

- New optional keyword argument defaulting to 1, which preserves the current sequential path. No
  signature or return-type change, and no change to output row order.

### DIFF
--- a/parsons/geocode/census_geocoder.py
+++ b/parsons/geocode/census_geocoder.py
@@ -1,4 +1,5 @@
 import logging
+from concurrent.futures import ThreadPoolExecutor
 
 import censusgeocode
 import petl
@@ -25,11 +26,19 @@ class CensusGeocoder:
         vintage: str
             The US Census vintage file to utilize. By default the current vintage is used, but
             other options can be found `here <https://geocoding.geo.census.gov/geocoder/vintages?form>`__.
+        workers: int
+            Number of batch chunks to geocode in parallel. Defaults to 1, which sends chunks
+            one at a time as before. Raising it increases the load placed on the Census API,
+            so leave it at 1 unless you have confirmed a higher rate is acceptable.
 
     """
 
-    def __init__(self, benchmark="Public_AR_Current", vintage="Current_Current"):
+    def __init__(self, benchmark="Public_AR_Current", vintage="Current_Current", workers=1):
+        if workers < 1:
+            msg = f"workers must be 1 or greater, got {workers}"
+            raise ValueError(msg)
         self.cg = censusgeocode.CensusGeocode(benchmark=benchmark, vintage=vintage)
+        self.workers = workers
 
     def geocode_onelineaddress(self, address, return_type="geographies"):
         """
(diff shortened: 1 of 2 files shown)


bmos:
I'm iffy on whether this should be included until after #1940 (or similar) is merged.
While there is no published ratelimit or SLA, I would assume that running parallel connections is likely to result in 429s pretty quickly.

bmos:
If you update this branch from main it'll fix the failures.

bmos:
@JakeSelby if you update this branch from main it'll fix the blocking test failures.

PR #1958, opened 2026-09-18
Fix Catalist matchback CSV parsing on unbalanced quote characters

sharinetmc:
## What is this change?

- Adds csv.QUOTE_NONE param to catalist matching output, since the csv reader is restringing everything after an non-closed quote as a giant string, which breaks the csv field limit

## Breaking Changes

<details open>

</details>

### Details (if needed)


codybraun (changes requested):
hmmmm is it possible to get some tests on this? I'd love to understand the behavior here a bit better- I would think that no quoting just introduces a different type of problem. For example, let's say we have a two-column CSV: 1, "groceries\n 2, "milk, carrots, bread" - if you don't respect quotes delineating string fields this CSV breaks because the second line appears to have four columns now

I might be misunderstanding what this config does but I'd love to see a test illustrating that if possible

bmos:
Yes, some of those issues are discussed in [the documentation for QUOTE_NONE](https://docs.python.org/3/library/csv.html#csv.QUOTE_NONE).

bmos:
Does Catalist have API documentation?

test/test_catalist/test_catalist.py
```diff
@@ -72,6 +73,30 @@ def test_load_matches_unzip(client, tmp_path):
...
+
+    A field like `"BN2` (an unescaped, unclosed quote) makes csv.reader's
+    default Excel dialect treat it as the start of a quoted field spanning
+    the rest of the file, eventually raising `_csv.Error: field larger than
+    field limit`. Matchback files are plain TSVs where quoting has no
+    meaning, so load_matches parses them with quoting=csv.QUOTE_NONE
+    (parsons/catalist/catalist.py).
+    """
+    results_csv = tmp_path / "results.csv"
+    results_csv.write_text('COL1-first_name\tCOL2-zip\tDWID\nJane\t"BN2\t123\nJohn\t90210\t456')
```
IanRFerguson, line 88: nice I think this ought to do it

IanRFerguson (approved):
bizarre label behavior notwithstanding this looks great

bmos:
The label behavior is because @sharinetmc  didn't check either of the boxes in the pull request description.

test/test_catalist/test_catalist.py
```diff
@@ -72,6 +73,30 @@ def test_load_matches_unzip(client, tmp_path):
...
+    swallowing the rest of the file.
+
+    A field like `"BN2` (an unescaped, unclosed quote) makes csv.reader's
+    default Excel dialect treat it as the start of a quoted field spanning
+    the rest of the file, eventually raising `_csv.Error: field larger than
+    field limit`. Matchback files are plain TSVs where quoting has no
+    meaning, so load_matches parses them with quoting=csv.QUOTE_NONE
+    (parsons/catalist/catalist.py).
+    """
+    results_csv = tmp_path / "results.csv"
```
codybraun, line 111: okay, I think this is a great test- the other thing that would be nice to see is what happens to this row if it looks like this: `"John, Jr", Smith` - that case is what I worry about- basically when there is a quoted field with a comma in it that needs to not be treated as a delimiter. Does that make sense?

sharinetmc: Hmmm...yes, I can def add that, but if this is tab deliminated, should I testing situations where it's like "John      Jr", Smith?

test/test_catalist/test_catalist.py
```diff
@@ -72,6 +73,76 @@ def test_load_matches_unzip(client, tmp_path):
...
+        'last_line_no_newline\t"stray'
+    )
+
+    cleaned = _strip_lone_quotes(text)
+    lines = cleaned.splitlines()
+
+    assert lines[0] == "no_quotes\tplain"
+    assert lines[1] == "one_quote\tBN2"
+    assert lines[2] == 'balanced\t"quoted value"'
+    assert lines[3] == 'three_quotes\t"a"b"'
```
sharinetmc, line 95: I made it strip quotes only if there are 1 long quote in a line. This will not catch if there are odd number of quotes greater than one, but hopefully that's rare enough that we can just. not deal with that for now.

parsons/catalist/catalist.py
```diff
@@ -21,6 +21,24 @@
...
+    untouched, since those are presumably balanced pairs. Lines with more
+    than one stray quote are also left untouched, since it's ambiguous
+    which one is the culprit.
+    """
+    lines = text.splitlines(keepends=True)
+    return "".join(
+        line.replace(quotechar, "", 1) if line.count(quotechar) == 1 else line for line in lines
+    )
+
+
```
sharinetmc, line 41: was trying to see if we could ignore bad rows for this instance because I agree, it would be the best option, but because the open ended quote makes everything after it a single row, I _don't_ think there's a good way to tell the csv reader to differentiate the giant chunk as separate rows. Idk if what I wrote makes sense?

codybraun (approved):
yeah this seems fine, seems like we already have the whole CSV in memory so splitting into lines seems okay

PR #1959, opened 2026-09-19
Full type hinting for Table, ETL, and ToFrom classes

bmos:
Picking up from #1740 

If it wasn't for documentation improvement, this would be a noticeable reduction in lines of code!

## What is this change?

- Add full type hinting for Table object and all methods
- Add support for Path objects wherever local paths are accepted
- Add/improve documentation throughout
- Move Exception messages to string variables to improve exception output
- Add sql_parameters arg to all relevant from sql methods in ToFrom to guard against SQL injection
- Fix args in ETL.convert_column
- Change ToFrom.from_csv_string arg name from str to csv_string (with deprecation)
- Update links to external documentation
- Misc code cleanup / modernization (such as using assignment expressions to avoid referencing a single key in a dict twice in a row)

## Breaking Changes

<details open>

</details>

### Details (if needed)


bmos:
Of the code not covered by tests, I think that `ToFrom.to_sftp_csv` and `ToFrom.to_s3_csv` are the only methods that are likely to need scrutiny.

